PR #478, opened 2013-12-10
Regex support for checking usernames against the USERNAME_BLACKLIST.

eezis:
A case-insensitive regex check to prevents users from creating Usernames that use *words* in the ACCOUNT_USERNAME_BLACKLIST.

if ACCOUNT_USERNAME_BLACKLIST = ['shit', 'ass',] 

then ......usernames like 'AssHat' and 'ShitMonkey' will be disallowed. This was a TODO item.

NOTE: While 'AssHat' would not be allowed; 'BassFisherman' and 'Lassie' would also get stopped! it might make sense to add a BLACKLIST_EXACT_MATCH option and then update the docs so that people can choose between an exact case-insensitve match (using re.match) or the more aggressive match on partial word approach that I have used.

### DIFF
--- a/allauth/account/adapter.py
+++ b/allauth/account/adapter.py
@@ -1,5 +1,6 @@
 import warnings
 import json
+import re
 
 from django.conf import settings
 from django.http import HttpResponse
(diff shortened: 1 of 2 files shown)


pennersr:
Closing this -- `clean_username` is the recommended way to address these kind of things.

PR #616, opened 2014-05-15
added commit=bool(not app_settings.SIGNUP_FORM_CLASS)

HemantNegi:
The problem was with custom signup form with some required fields. It simply trys to save the user without adding mandatory fields and which throws "required key error". So this was the simple workaround i figured out without adding a extra setting.

i.e. if someone has a custom signup form then let him save the form manually in   
     user.save()

method.


SteveByerly:
I'm pretty sure you can already do this in the 'save_user' method of the adapter. Just derive your own adapter and use that.

You also introduced a new setting in the FP, but didn't document what it was or add any useful logic.

pennersr:
@SteveByerly is correct, so I prefer to pass this one ...

Issue #622, opened 2014-05-22
djangoall auth not respecting the EMAIL_CONFIRMATION_AUTHENTICATED_REDIRECT_URL setting


venkatesh22:
djangoall auth not respecting the EMAIL_CONFIRMATION_AUTHENTICATED_REDIRECT_URL setting

as per the account/views.py  after confirming the email , performing user login and returning response from there , please see the below code

```python
    class ConfirmEmailView(TemplateResponseMixin, View):
    
        def post(self, *args, **kwargs):
            self.object = confirmation = self.get_object()
            confirmation.confirm(self.request)
            get_adapter().add_message(self.request,
                                  messages.SUCCESS,
                                  'account/messages/email_confirmed.txt',
                                  {'email': confirmation.email_address.email})
            if app_settings.LOGIN_ON_EMAIL_CONFIRMATION:
                resp = self.login_on_confirm(confirmation)
                #if resp:#
                    #return resp#  issue is here 
            redirect_url = self.get_redirect_url()
            if not redirect_url:
                ctx = self.get_context_data()
                return self.render_to_response(ctx)
            return redirect(redirect_url)
```

pennersr:
I assume this is the same issue as #633 ...

pennersr:
https://codeberg.org/allauth/django-allauth/issues/622

PR #624, opened 2014-05-27
commented return resp after login to respect get_email_confirmation_redi...

venkatesh22:
...rect_url     #622


pennersr:
This pull mixes things, it comments out some code and (unrelated) adds a request parameter, while the intentions are not fully clear to me.

Issue #628, opened 2014-06-03
add support to qq


jackxpeng:
I added qq support to allauth and would like to contribute it to this project.
1. It's based on the weibo/oauth2 provider.
2. QQ requires you to obtain an openid before being about to access their apis. I decided to use the openid as the uid since there is no id string in extra_data.
3. To add this I created an openid_connect provider that inherit from classes in the oauth2 app.

pennersr:
https://codeberg.org/allauth/django-allauth/issues/628

PR #629, opened 2014-06-04
added qq openid connect support

jackxpeng:
https://github.com/pennersr/django-allauth/issues/628

I added qq support to allauth and would like to contribute it to this project.

It's based on the weibo/oauth2 provider.
QQ requires you to obtain an openid before being about to access their apis. I decided to use the openid as the uid since there is no id string in extra_data.
To add this I created an openid_connect provider that inherit from classes in the oauth2 app.

### DIFF
--- a/allauth/socialaccount/providers/openid_connect/client.py
+++ b/allauth/socialaccount/providers/openid_connect/client.py
@@ -1,0 +1,36 @@
+from __future__ import absolute_import
+
+import re, requests
+from allauth.socialaccount.providers.oauth2.client import OAuth2Client
+
+class OpenidConnectClient(OAuth2Client):
+
+    def __init__(self, request, consumer_key, consumer_secret,
+                 access_token_method,
+                 access_token_url,
+                 callback_url,
+                 open_id_url,
+                 scope):
+        self.request = request
+        self.access_token_method = access_token_method
+        self.access_token_url = access_token_url
+        self.open_id_url = open_id_url
+        self.callback_url = callback_url
+        self.consumer_key = consumer_key
+        self.consumer_secret = consumer_secret
+        self.scope = ' '.join(scope)
+        self.state = None
+
+    def get_open_id(self, token):
+        params = {'access_token': token}
+        resp = requests.request('GET',
+                                url = self.open_id_url,
+                                params=params)
+        open_id = None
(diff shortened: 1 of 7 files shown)


brunosmartin:
Any news on this? One year and no merge...

peckjon:
@jackxpeng I'm looking to use this but don't see OpenID Connect available in /admin/socialaccount/socialapp/add/ after adding it as a provider... wondering if you could provide an example?  Any help would be appreciated.

@pennersr any mention on whether you plan on merging this?  If not, do you have an example config using the stock oauth provider?

pennersr:
Must admit, I need to dive into the OpenID connect spec before I can tackle this one, but so far I have not been getting around to do so...

pennersr:
This PR collects some base functionality in a `openid_connect` app. At this point in time, I am not sure the encapsulated code is a sound base for OpenID connect functionality. For example, there is more to it than what is encapsulated now (discovery, standardized fields, ...). So, I would rather prefer for `qq` to use the OAuth2 provider code as a base and not lay down an `openid_connect` base foundation that likely needs revisitting in the near future. It should be possible to fold the `get_open_id` into the `complete_login` and not having to tweak the base oauth2 provider code.

iloveyo123u1:
i need this

samuelchen:
Really need this. I vote +1.
Years past... hope it can be merged.

stfenjobs:
兄弟你好， 这里有一个 用OAuth2.0 实现的QQ https://github.com/pennersr/django-allauth/pull/2109
你方便帮忙改改吗

abamuc:
Hi,

as I believe it would really be useful to have oidc, what requirements are there for mergeable code? Does the sample code from #676 look sane and "just" needs to be put into an generic oidc provider in the django allauth namespace? (Looking at some other implementation the configuration looks like
      'issuer' => '$URL',
      'uid_field' => 'uid',
      'client_options' => {
        'identifier' => '$id',
        'secret' => '$secret',
        'redirect_uri' => '$OWNURL'
      }
for a generic oidc connector, and that doesn't look to bad to me).

For the other direction, django-oidc-provider works well, so this could be used to check the implementation against.

pennersr:
Closing in favor of #3165

PR #644, opened 2014-06-16
Delete email confirmations after address is verified

bitcity:
Model manager has a new method to delete all confirmations for a given email address. This is called inside model's `confirm()` method right after it calls `get_adapter().confirm_email`

### DIFF
--- a/allauth/account/managers.py
+++ b/allauth/account/managers.py
@@ -69,3 +69,6 @@ class EmailConfirmationManager(models.Manager):
 
     def delete_expired_confirmations(self):
         self.all_expired().delete()
+
+    def delete_confirmations(self, email_address):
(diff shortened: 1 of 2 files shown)


pennersr:
The are various reasons for keeping the logic as is, UX wise so that you can click the link twice, but also for statistics and analysis. I feel there are already plenty of ways you can hook into allauth to delete this confirmations as you see fit.

With respect to the referenced UX link: allauth does exactly this: "... At the very least, allow your users who are confirming their email address in the same session as they create the account to just be logged in, please! For all others (or for cases where the session data is unclear), requiring them to log in again is fine."

As for the issue you run into when using signals: "A user can request a verification email for an account that's already verified.". This is an issue that is handled as part of #754 

All in all, closing this..

PR #705, opened 2014-09-02
prevent users from submitting social signup form twice (eg. using back b...

mwjackson:
...utton)

we end up with multiple token errors inside SocialLogin.lookup


pennersr:
You mention you were getting "multiple token errors". Are you referring to the MultipleObjectsReturned exception? I am wondering how you could manage to run into that, considering that `app` and `account` are `unique_together` on SocialToken, and the lookup is based on just these two keys. Could you please provide more details on the problem you ran into?

mwjackson:
No idea really, it was 6 months ago I created this PR.

I'll be updating our fork from your master in the next few days, I'll have a look then

pennersr:
Closing until then, feel free to reopen if you run into something ...

PR #784, opened 2014-12-01
Update advanced.rst

wasabigeek:
Seems on a custom user model w/o username, the ACCOUNT_USERNAME_REQUIRED = False, and ACCOUNT_EMAIL_REQUIRED = True need to be specified, as the defaults do not work.


pennersr:
Closed via 2f69cd7

PR #793, opened 2014-12-18
Remove need for _base_signup_form_class() function and simplify Signup form system for subclassing

wldcordeiro:


wldcordeiro:
@pennersr I'm going to clean up the commit history for this PR but could you possibly take a look at it? Currently there's 8 tests that have errors but I'm not sure how it's happening. They seem to be changing settings that I have overridden back to the defaults.

PR #797, opened 2014-12-26
Adds initial Evernote OAuth support

pztrick:
I ended up copying/pasting the entire `OAuthCallbackView` to fix a `KeyError` because Evernote doesn't return an `oauth_token_secret`:

https://github.com/pztrick/django-allauth/compare/pennersr:master...feature/evernote_oauth#diff-6be7fa52b030f92e42ec8b081697b3cfR57

We could make the `OAuthCallbackView` use `access_token.get('oauth_token_secret', None)` to avoid the `KeyError` but wasn't sure if this is a desired approach / if the KeyError should in fact occur for other services.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -58,6 +58,7 @@ Martin Bächtold
 Mauro Stettler
 Nariman Gharib
 Niklas A Emanuelsson
+Patrick Paul
 Paulo Eduardo Neves
 Peter Bittner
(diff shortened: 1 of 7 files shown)


pennersr:
Thanks! Integrated in 8292dfdf3dec12fa03d70d1f432fb9e842459e07

PR #820, opened 2015-02-01
Strip away empty params in OAuth2 client on login

gaqzi:
I'm working on an update to the Dropbox provider, so that it'll work with OAuth2, and noticed that Dropbox doesn't work with the `scope` attribute.

I'm hoping this fix will be able to easily take care of that issue and at the same time not affect other providers.

My thinking is that if the key is empty the remote side ought to accept the value even when missing, having some kind of default.

### DIFF
--- a/allauth/socialaccount/providers/oauth2/client.py
+++ b/allauth/socialaccount/providers/oauth2/client.py
@@ -46,6 +46,7 @@ class OAuth2Client(object):
                 'scope': self.scope,
                 'code': code}
         params = None
+        self._strip_empty_keys(data)
         url = self.access_token_url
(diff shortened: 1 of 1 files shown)


gaqzi:
Looking at the [OAuth2 specs](https://tools.ietf.org/html/rfc6749#section-3.3) I found this:

> If the client omits the scope parameter when requesting
> authorization, the authorization server MUST either process the
> request using a pre-defined default value or fail the request
> indicating an invalid scope.  The authorization server SHOULD
> document its scope requirements and default value (if defined).

So I feel omitting the scope attribute when there's no defined values is valid according to the spec. :)

Issue #821, opened 2015-02-01
Upgrade Dropbox provider to use OAuth2


gaqzi:
I'm currently working on updating the Dropbox provider to use OAuth2.

But I'm a bit curious whether I should update the current one or create a new one for OAuth2?

Since the API key would change for users I guess it could break things for other users of the Dropbox provider.

I will for now prepare for it to be a dropbox2 provider. :)

pennersr:
We indeed need to support both protocols. Have a look at the LinkedIn provider -- it also offers both. So, if you could put your work in a `dropbox_oauth2` package that would be great..

JshWright:
This can be closed now, right? (per PR #824)

pennersr:
https://codeberg.org/allauth/django-allauth/issues/821

PR #824, opened 2015-02-03
Add Dropbox OAuth2 provider

gaqzi:
A basic provider which will as far as I can tell works well. 
I've gotten an access token and users to login, so I'm happy with it for now. :)

For this commit to work it relies on PR #820.
This closes issue #821.


pennersr:
Integrated in 0f744dd3bc1ba578de3bb804b45fa02f958c0e9c

PR #828, opened 2015-02-06
Update allauth to utilize url namespaces.

wldcordeiro:
Allauth was using urls like `account_signup` and this seemed to be antiquated with the existence of namespaces. I updated it to utilize namespaces. The entire test-suite is now passing.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,2 +1,7 @@
 [run]
 include = allauth*
+omit =
+    */migrations/*
+    *__init__*
+    *test*
+    *app_settings*
--- a/.coveragerc
+++ b/.coveragerc
@@ -5,3 +5,4 @@ omit =
     *__init__*
     *test*
     *app_settings*
+    allauth/socialaccount/providers/*/models.py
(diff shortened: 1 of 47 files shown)


JshWright:
Wouldn't this be a breaking change for folks who have references to those URLs in their existing code?

SteveByerly:
I would argue that any namespacing should be done by the user integrating this app into their own project, in the project level urls where they are included. Adding a namespace in the app forces you to abide by this namespace which can potentially cause conflict.

wldcordeiro:
@JshWright It would but it's a simple fix and creates a cleaner URL pattern overall.

@SteveByerly if the patterns already resemble namespaced ones e.g. `account_login`, `account_signup` it doesn't make sense to not utilize the feature and in this case the user can still namespace the top-level url patterns like

```py
url(r'^users/', include('allauth.urls', namespace='users'))
```
then they can use the urls like

```py
reverse('users:account:signup')
```

JshWright:
Personally I would be opposed to breaking changes in a point release, but I'm just a random contributor...

wldcordeiro:
@JshWright This project hardly follows any form of semver.

pennersr:
I am sorry, while probably cleaner, trading in an underscore for a colon does not bring any gains, but it would break a lot of installations. Without any clear benefit, this is really not worth it ...

PR #832, opened 2015-02-09
Small clarifications to how to use email rather than username

jessamynsmith:
Thank you for writing such a configurable package! I ended up reading the source to figure out how to set up a user with just email and no username, so I thought I'd make it easier for others.


jessamynsmith:
I can't see how me editing the docs could have caused a failed build. :(

pennersr:
The change you made is actually incorrect. If you only have an `email` column and no `username`, then set `ACCOUNT_USER_MODEL_USERNAME_FIELD` to `None`.

jessamynsmith:
Thank you for that, I've edited the PR accordingly.

PR #856, opened 2015-02-27
Added docs for missing signals

julen:
Some signals were missing in the docs. I have added them, improved sectioning and added the provided args as well. Feel free to improve on the descriptions — they're very basic!


pennersr:
Thanks!

PR #861, opened 2015-03-08
Adding two-factor support to allauth - WIP

valberg:
This PR is about adding two-factor authentication support to django-allauth.

Todo:
- [x] Setup two-factor authentication
- [x] Backup token generation
- [ ] Do two-factor authentication on login when `ACCOUNT_TWO_FACTOR_ENABLED = True`
- [ ] Docs
- [ ] Tests

Requirements:
- django-otp==0.3.0
- qrcode==5.1

All help is more than welcome :smile:


pennersr:
Adding support for 2FA will introduce a lot of new flows and make the allauth core larger then I feel comfortable with. I do not think it is necessary nor beneficial to build this into allauth itself. It could just as well reside in a separate `allauth.contrib.2fa` style package/project. If, while implementing this you run into allauth limitations that make things difficult to implement this separately then please let me know as I am eager to address any of these. Other than that, I prefer to keep this, and other functionality such as invitation handling, completely separate.

valberg:
@pennersr I totally agree, I've also begun doing it as a side project one can install along allauth - this way we can test it and maybe merge into allauth.contrib in the future :)

valberg:
Just as a small update on this:
https://github.com/percipient/django-allauth-2fa/
I don't know if there are other initiatives in this direction, but just wanted to put it on your radar @pennersr 😄

nscozzaro:
With 2FA being almost synonymous with "authentication" these days it makes sense to me to merge 2FA into AllAuth in my opinion.

valberg:
@nscozzaro It is quite easy to add 2FA by using https://github.com/percipient/django-allauth-2fa/. The nice thing about not having this in "allauth core" is that since not all authentication situations demand 2FA, it can exist without meaning more development cost for "allauth core". Just like django doesn't just adopt 3rd party packages because they are widely used (django rest framework comes to mind).

Glowsome:
2FA is a/the way in the world of federated logins someting that is forced (or not) by a/the IDP we are talking with.
if one does not 'dictate' a/the Authentication contract it will always execute the default contract (unless explicitly configured)

Simple example : 

- i have an IDP who had a default authentication contract of username/password.
- The IDP also has a username/password/Multifactor contract configured with its own urn.

IF i hook up ( in this case an OAUTH(2) client and dont send anything extra, it will on landing on the application unauthenticated redirect me to my IDP asking me to authenticate execute the default contract ( in my case user/password). and then send me(after having fulfilled authentication) back authenticated as such to the application.

HOWEVER IF i hook up an application and in the authentication request being sent i include an acr_value defining the urn of a contract (should be available on the IDP) explicitly telling it to authenticate with it it will ( if it points to a 2FA/Multifactor contract) execute this.

So IMHO the whole deal in getting this project '2FA/Multifactor'-enabled is all about being able to send an acr_value which will if set execute a/the given Authentication contract as defined on the IDP.

PR #862, opened 2015-03-10
Follow OpenID Connect (OAuth 2.0 for Login)

sspross:
Switch from `verified_email` to `email_verified`.

Deprecated early version using `verified_email`: https://developers.google.com/accounts/docs/OAuth2LoginV1
OpenID Connect: https://developers.google.com/accounts/docs/OpenIDConnect


sspross:
hm... sry I thought it's a quick fix like in another project, will dig into it deeper.

PR #863, opened 2015-03-11
Stop creation of extraneous migration files

briandant:
This commit edits the first migration to use `registry.as_choices()` for the `choices` argument. `makemigrations` will no longer create migrations files when a provider is removed or added.


sebslomski:
+1

PR #871, opened 2015-03-17
regenerated migrations

sih4sing5hong5:
The field are changed from bytestring to string.
This file is generated by `python manage.py makemigrations`


pennersr:
This conflicts now that #863 is merged ...

PR #874, opened 2015-03-24
Unset email widget's is_required status too

markhepburn:
I've only noticed this when email is optional, and also using
django-bootstrap3.

The bootstrap3 default render checks the widget's `is_required` status,
which is only initialised in the django field constructor.  So, setting
field.required to False after initialisation still leaves
`widget.is_required` in the default state of True, which then causes the
bootstrap renderer to mistakenly add a required=required attribute.

### DIFF
--- a/allauth/account/forms.py
+++ b/allauth/account/forms.py
@@ -228,6 +228,7 @@ class BaseSignupForm(_base_signup_form_class()):
         else:
             self.fields["email"].label = ugettext("E-mail (optional)")
             self.fields["email"].required = False
+            self.fields["email"].widget.is_required = False
             if self.username_required:
                 field_order = ['username', 'email']
 


pennersr:
I am wondering, Isn't this an issue in django-bootstrap3 then? I don't like adding workarounds to allauth for issues in other packages unless absolutely necessary.

markhepburn:
I did have a think about which project it belonged to.

I think it's an allauth issue though; allauth creates a non-required field with a required widget -- bootstrap3 is just rendering that widget correctly, based on what it can interrogate.  I'm open to suggestions otherwise though.

pennersr:
But, if this scenario works with a standard Django installation, then isn't BS3 to blame?

markhepburn:
Yeah, that's a fair question -- I think if the widget.required flag is part of the "API" (such as it is) then it's an allauth issue.  I did notice that it doesn't seem to used almost anywhere else, so I wondered at the time if it's just a legacy attribute that BS3 still uses, in which case it's a BS3 issue.

Do you know any Django contributors we could ask?

(I don't think Django adds the required attribute at all, which might be why it works)

markhepburn:
Many thanks.

PR #875, opened 2015-03-25
Minor cleanups

julen:
Minor cleanup I came across while reading through the source code.


pennersr:
I must admit, I actually like the "superfluous `else` branching", as it gives me more explicit context that 2 is returned if (not something) holds. If I were to tackle this, I would prefer to turn this:

    if something:
        return 1
    else:
        return 2

not in this:

    if something:
        return 1
    return 2

but into this:

    if something:
        ret = 1
    else:
        ret = 2
    return ret

In general, I don't like methods that exit  somewhere in the middle...

julen:
Fair enough, notice both ways are somehow mixed all over the codebase, the superfluous being prevalent. I've removed that commit, the other one should be fine I believe.

PR #877, opened 2015-03-25
Added tests for verified and unverified login attempts

julen:
Tests that user will be able to either
* log in and therefore be redirected to `settings.LOGIN_REDIRECT_URL` (optional verification), or
* reach the email verification view (mandatory verification)

Also removed the empty `test_login_view()` test.

### DIFF
--- a/allauth/account/tests.py
+++ b/allauth/account/tests.py
@@ -260,6 +260,39 @@ class AccountTests(TestCase):
         EmailAddress.objects.add_email(request, u, u.email, confirm=True)
         self.assertTrue(mail.outbox[0].subject[1:].startswith(site.name))
 
+    @override_settings(
+        ACCOUNT_EMAIL_VERIFICATION=app_settings.EmailVerificationMethod
+        .OPTIONAL)
+    def test_login_unverified_account_optional(self):
+        """Tests login behavior when email verification is optional."""
+        user = get_user_model().objects.create(username='john')
+        user.set_password('doe')
+        user.save()
+        EmailAddress.objects.create(user=user,
+                                    email='john@example.com',
+                                    primary=True,
+                                    verified=False)
+        resp = self.client.post(reverse('account_login'),
+                                {'login': 'john',
+                                 'password': 'doe'})
+        self.assertEqual(resp['location'],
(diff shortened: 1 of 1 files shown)


allauth/account/tests.py
```diff
@@ -301,27 +330,6 @@ def test_ajax_login_success(self):
...
-        user.set_password('doe')
-        user.save()
-        EmailAddress.objects.create(user=user,
-                                    email='john@example.com',
-                                    primary=True,
-                                    verified=False)
-        resp = self.client.post(reverse('account_login'),
-                                {'login': 'john',
-                                 'password': 'doe'},
-                                HTTP_X_REQUESTED_WITH='XMLHttpRequest')
```
pennersr: Why did you remove this test case, as it tests AJAX based logins, which is not tested by the tests you added?

allauth/account/tests.py
```diff
@@ -301,27 +330,6 @@ def test_ajax_login_success(self):
         data = json.loads(resp.content.decode('utf8'))
         self.assertEqual(data['location'], '/accounts/profile/')
 
-    @override_settings(
-        ACCOUNT_EMAIL_VERIFICATION=app_settings.EmailVerificationMethod
-        .MANDATORY,
-        ACCOUNT_AUTHENTICATION_METHOD=app_settings.AuthenticationMethod
```
pennersr: Please override the `ACCOUNT_AUTHENTICATION_METHOD` method as well, as you are logging in by username.

allauth/account/tests.py
julen: It's exactly the same test as `test_login_unverified_account_mandatory`. The fact that a view is requested with the `X-Requested-With: XMLHttpRequest` header shouldn't be tested for every single scenario IMO.

allauth/account/tests.py
julen: Unless I'm missing something, `AccountTests` already defaults to `AuthenticationMethod.USERNAME`, and [the `@override_settings` decorator](https://docs.djangoproject.com/en/1.7/topics/testing/tools/#django.test.override_settings) overrides settings only for a method. So is it really necessary?

PR #889, opened 2015-03-29
Travis: switched to container-based builds

julen:
This enables [Travis' container-based builds](http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/), which should speed-up tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 language: python
 python:
  - "2.6"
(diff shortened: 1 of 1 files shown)


pennersr:
This does not seem to work?:
 
    sudo: must be setuid root

    The command "sudo apt-get update -qq" failed and exited with 1 during .

julen:
Ah yeah, didn't realize `sudo` was being used in *.travis.yml* to install `sloccount`. I wonder what is that needed for (I guess coverage, but not sure where it's used exactly) and if there's any possible replacement without the need to use apt-get?

julen:
I've removed sloccount as I see it's used nowhere. The build passes and you can see the speed-up: builds start quicker and takes ~7 minutes less to build them all.

PR #891, opened 2015-03-29
Try application setting when looking for login redirect url after email

pbf:
This fixes an issue I've encountered when using email confirmation with ACCOUNT_EMAIL_CONFIRMATION_AUTHENTICATED_REDIRECT_URL = '/' and LOGIN_REDIRECT_URL at default. The expected behavior is for django-allauth to redirect using relevant setting, but as of now only LOGIN_REDIRECT_URL and "next" field are tested.


pennersr:
This needs to be routed via the adapter, so `self.get_redirect_url()` is the url that is to be passed.

PR #894, opened 2015-04-01
Password reset: deal with invalid tokens when in XHR

### LINKED ISSUE #890: Reset password from key endpoint response when token fails.
When you call reset password endpoint using ajax, the normal flow works fine but if you send an used token, the response is a HTML template saying the token is invalid and its status is 200. 

Response should be json dict with an error and status 403 or 400.

julen:
Not a big fan of special-casing here, but honestly I don't see any other way around without overhauling the entire architecture behind XHR views (but that's a separate topic).

Fixes #890.

### DIFF
--- a/allauth/account/views.py
+++ b/allauth/account/views.py
@@ -607,7 +607,19 @@ class PasswordResetFromKeyView(AjaxCapableProcessFormViewMixin, FormView):
         return super(PasswordResetFromKeyView, self).form_valid(form)
 
     def _response_bad_token(self, request, uidb36, key, **kwargs):
-        return self.render_to_response(self.get_context_data(token_fail=True))
+        # Special-casing ajax/non-ajax scenarios due to #890
+        if request.is_ajax():
+            data = {
+                'errors': ['token_failed'],
+            }
+            response = _ajax_response(request,
+                                      HttpResponseBadRequest(reason=data))
+        else:
+            response = self.render_to_response(
+                self.get_context_data(token_fail=True)
+            )
+
+        return response
 
 password_reset_from_key = PasswordResetFromKeyView.as_view()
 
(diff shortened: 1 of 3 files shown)


allauth/account/views.py
```diff
@@ -602,7 +607,19 @@ def form_valid(self, form):
...
 
     def _response_bad_token(self, request, uidb36, key, **kwargs):
-        return self.render_to_response(self.get_context_data(token_fail=True))
+        # Special-casing ajax/non-ajax scenarios due to #890
+        if request.is_ajax():
+            data = {
+                'errors': ['token_failed'],
+            }
+            response = _ajax_response(request,
+                                      HttpResponseBadRequest(reason=data))
```
pennersr: For HttpResponse, the reason is supposed to be a string type. This is not enforced, so the above code will work, but it may break in the future. Also, allauth so far managed to stay away from being opinionated with respect to the contents of the XHR responses. This would be the first place (adapter excluded) where choices are being made as to what content is being passed.  So I am wondering, wouldn't it be better to simpy raise a 400 without passing any specific data?

Perhaps cleaner is this approach: we could create a specific form (`UserTokenForm`) and move the token validation and user lookup in there. Then, in PasswordResetFromKeyView instantiate this form (not passing request.GET/POST but simply the uidb36/key in directly). If the validation fails, then this form could be passed to the `ajax_response`, where in the adapter people can do whathever they please, including a data error response like above.

julen: Yeah, having the form would work transparently for XHR views, as the `AjaxCapableProcessFormViewMixin` already takes care of that. Will try something down that route.

pennersr: Still, `AjaxCapableProcessFormViewMixin` does not kick in for POST, so that issue remains..

julen: It does for POST; you mean for GET?

pennersr: Oops, ofcourse.

julen:
Closing in favor of #895.

PR #895, opened 2015-04-01
Password reset: perform token validation via form

### LINKED ISSUE #890: Reset password from key endpoint response when token fails.

julen:
This allows to reuse the handlers for XHR responses, and therefore return an error message with a 400 status code when the token is invalid.

Note that this implements the alternative approach as suggested in #894.

Fixes #890.

### DIFF
--- a/allauth/account/forms.py
+++ b/allauth/account/forms.py
@@ -17,8 +17,9 @@ from ..utils import (email_address_exists,
                      get_username_max_length)
 
 from .models import EmailAddress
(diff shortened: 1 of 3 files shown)


pennersr:
Thanks!

PR #896, opened 2015-04-01
Cleanup: use `self.assert*()` functions

julen:
I've previously used ordinary `assert` calls (sorry I'm used to py.test mostly), which don't contain extra information when tests fail. This commit switches them for their `self.assert*()` counterparts.


pennersr:
Merging as this is an improvement, but for cases like these, `self.assertIn` would be the appropriate choice.

julen:
Oh thanks, didn't know that! As said I'm used to py.test...

julen:
Oh, [as I see](https://docs.python.org/2/library/unittest.html#test-cases) that wouldn't be 100% correct since they're only available starting in Python 2.7, and Travis also tests on 2.6.

pennersr:
Ah, indeed... then let's keep it as is now...

PR #908, opened 2015-04-07
Update advanced.rst

sih4sing5hong5:
Add docs for custom user model


sih4sing5hong5:
I have no idea to document how to customize these function.
By example? But it may be mass.

PR #912, opened 2015-04-09
activate_url should be set in the adapter.

mfa:
We have a strictly seperated rest-api from javascript frontend.
The link in the email should direct to an endpoint in the javascript app.
The easiest way is move get_activate_url to adapter.


lukeburden:
Hey @mfa,

I've hit a similar issue but have taken the approach of deferring full responsibility for composition and sending of the confirmation mail to the adapter layer. This let's custom adapters do just about anything they need to, and would solve your problem as well as mine.

It'd be great if you could take a look at my pull request and throw your support behind it if it meets your needs and you agree that it's a solid approach.

https://github.com/pennersr/django-allauth/pull/1021

Thanks!

Luke

PR #913, opened 2015-04-10
Add ACCOUNT_LOGOUT_ON_PASSWORD_CHANGE setting

rodxavier:
Since Django 1.7, if the SessionAuthenticationMiddleware is enabled, the session is invalidated when the user changes the account's password.

Starting Django 2.0, session verification will become mandatory whether or not the SessionAuthenticationMiddleware is enabled.

I figured this may be a good time to add this setting.

[Source: Session invalidation on password change](https://docs.djangoproject.com/en/1.8/topics/auth/default/#session-invalidation-on-password-change)

### DIFF
--- a/allauth/account/views.py
+++ b/allauth/account/views.py
@@ -28,9 +28,10 @@ from . import app_settings
 
 from .adapter import get_adapter
 
-import django
-if django.VERSION >= (1, 7):
+try:
     from django.contrib.auth import update_session_auth_hash
+except ImportError:
+    update_session_auth_hash = None
 
 
 sensitive_post_parameters_m = method_decorator(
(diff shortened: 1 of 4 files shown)


docs/configuration.rst
```diff
@@ -58,6 +58,11 @@ ACCOUNT_LOGOUT_ON_GET (=False)
   mere GET request. See documentation for the `LogoutView` for
   details.
 
+ACCOUNT_LOGOUT_ON_PASSWORD_CHANGE (=False)
+  Determines whether or not the user is automatically logged out after
+  changing the password. See documentation for `Django's session invalidation on password change`
```
julen: Nit: mind the line wrapping. It'd also help to link to [Django docs on the topic](https://docs.djangoproject.com/en/1.7/topics/auth/default/#session-invalidation-on-password-change).

allauth/account/views.py
```diff
@@ -28,6 +28,10 @@
 
 from .adapter import get_adapter
 
+import django
```
julen: Instead of checking for version numbers, alternatively you might want to:
```python
try:
    from django.contrib.auth import update_session_auth_hash
except ImportError:
    update_session_auth_hash = None
```

Then later on you can check for
```python
if (update_session_auth_hash is not None and
    not app_settings.LOGOUT_ON_PASSWORD_CHANGE):
```

julen:
This looks good cc @pennersr.

Issue #922, opened 2015-04-19
Avoid SQL joins


fmalina:
SQL joins are computationally expensive and allauth does a couple of them. This can easily hang queries on larger MySQL tables as I unfortunately discovered today. By larger I mean just 350k records was quite enough. Queries such as this should be broken down in 2.

    SELECT * FROM users
	LEFT OUTER JOIN account_emailaddress ON ( us.id = account_emailaddress.user_id )
	WHERE (us.email LIKE '...@...' OR account_emailaddress.email LIKE '...@...');

    SELECT DISTINCT (1) AS a, * FROM users
	LEFT OUTER JOIN account_emailaddress ON ( us.id = account_emailaddress.user_id )
	WHERE (us.email LIKE '...' OR account_emailaddress.email LIKE '...') LIMIT 1

fmalina:
And I am really not sure about the `LIKE` either, `=` is always faster. I am yet to discover a real world case sensitive email address. Lowercasing email addresses (on save, input) saves all the hassle.

fmalina:
The culprit is usage of Q objects in utility `filter_users_by_email` and auth backend method `_authenticate_by_email`. Using Q objects like that doesn't bode well with larger MySQL tables.

Here's an alternative implementation that produces a list useful in both scenarios (form clean_email methods need cleanup to use list produced instead of queryset):

```python
    def filter_users_by_email(email):
        User = get_user_model()
        mails = EmailAddress.objects.filter(email__iexact=email)
        users = [e.user for e in mails.prefetch_related('user')]
        if app_settings.USER_MODEL_EMAIL_FIELD:
            q_dict = {app_settings.USER_MODEL_EMAIL_FIELD + '__iexact': email}
            users += list(User.objects.filter(**q_dict))
    return list(set(users))
```

One can still make a queryset like so:
```python
User.objects.filter(pk__in=[u.pk for u in filter_users_by_email(email)])
```

Using this in `_authenticate_by_email` would also fix a bug, where it assumes the `USER_MODEL_EMAIL_FIELD = 'email'`.

I'll try to clean this up and make a pull request.

fmalina:
Pull request ready, please test/merge when you get a chance.

pennersr:
Merged 1e6062bc

pennersr:
https://codeberg.org/allauth/django-allauth/issues/922

PR #924, opened 2015-04-20
Fixes issue #922 "avoid SQL joins"

fmalina:
Use fast python code instead of slow SQL joins. This fixes draining of all CPU followed by "too many connections" on MySQL server. Also fixes hardcoded e-mail in auth_backend as it's customisable via USER_MODEL_EMAIL_FIELD.


fmalina:
No need for further testing, please merge.

PR #928, opened 2015-04-22
Fix account initial migration

comzeradd:
At Django 1.8 the default `max_length` for `EmailField` was increased from 75 to 254 in order to be compliant with RFC3696/5321.

[Reference](https://docs.djangoproject.com/en/1.8/ref/models/fields/#emailfield)


feffe:
Why would you update the initial migration file? I already have this in production so an update of the initial migration file wouldn't be applied.

The max_length change should only be applied if Django >=1.8 is used. Is there any way to check for this in a migration file so that it can be applied later if Django is upgraded to >=1.8?

pennersr:
The max_length change can be applied regardless of the Django version. This has been done in 4e81a3c

PR #930, opened 2015-04-23
use base64 encoding for user pk in password reset routine

barszczmm:
This fixes issues when users pk is new UUID field (but it should also help in other cases when user pk is not int).


pennersr:
Changing of the name (uidb36) may cause backwards compatibility issues, which is not needed per se. Also, the previous code checked the pk type and took the easy route in case of a regular int-like field. Not sure why that needs to be dropped.

pennersr:
I think this has been sufficiently addressed by 8f5c9b0 and 34a4e65

Issue #935, opened 2015-04-30
Google provider does not check status of authentication response - results in KeyError


yarbelk:
https://github.com/pennersr/django-allauth/blob/cffb971ad691a1f9a24775161d08387ca6a64c17/allauth/socialaccount/providers/google/views.py#L20

when you are authenticating with google - if you get a non 200 response (say 401), the provider still tries to parse the response for ID.

however - there is no ID in a 401 response.

Result: 500 Server Error

Expected Result 401

pennersr:
https://codeberg.org/allauth/django-allauth/issues/935

PR #936, opened 2015-04-30
Respect Google Provider's HTTP reponse

yarbelk:
- if google responds to you with  a 401 where your token is no good,
   raise an `HTTPError` (resp.raise_for_status) so that you can handle
   the exception corretly.  Otherwise you just get a 500 stauts because
   of a `KeyError`.


yarbelk:
Slightly messy fix for #935

PR #942, opened 2015-05-06
Cleanup: removed redundant conditional branching

julen:
Spotted this redundancy in the social account adapter.


pennersr:
This is probably subjective, but personally I prefer to keep branches on major app diversity (such as `account_settings.UNIQUE_EMAIL`) separate from branches on regular application logic...

PR #948, opened 2015-05-15
Fix django 1.8 RequestSite error on LoginView

tony:
Django 1.8 will break if using sites framework

- https://docs.djangoproject.com/en/1.8/ref/contrib/sites/#get-current-site-shortcut
- https://github.com/django/django/blob/1.8/django/contrib/sites/models.py#L50

When using django 1.8 on settings without a ``SITE_ID``, an exception will be raised.

Tthis is permitted in 1.8, since django introduces [get_current_site](https://docs.djangoproject.com/en/1.8/ref/contrib/sites/#get-current-site-shortcut) support in ``get_current`` method in the Manager through the newly-added ``request`` kwarg.

   > Changed in Django 1.8:
   > This function will now lookup the current site based on request.get_host() if the SITE_ID setting is not defined.

Backward compatible with < 1.8.

### DIFF
--- a/allauth/socialaccount/views.py
+++ b/allauth/socialaccount/views.py
@@ -1,7 +1,6 @@
 from django.contrib import messages
 from django.http import HttpResponseRedirect
 from django.core.urlresolvers import reverse, reverse_lazy
-from django.contrib.sites.models import Site
 from django.contrib.auth.decorators import login_required
 from django.views.generic.base import TemplateView
 from django.views.generic.edit import FormView
@@ -9,7 +8,7 @@ from django.views.generic.edit import FormView
 from ..account.views import (CloseableSignupMixin,
                              RedirectAuthenticatedUserMixin)
 from ..account.adapter import get_adapter as get_account_adapter
-from ..utils import get_form_class
+from ..utils import get_form_class, get_current_site
 
 from .adapter import get_adapter
 from .models import SocialLogin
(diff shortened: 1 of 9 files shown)


pennersr:
There are more places where the site is being looked up. Could we move your lookup into a new to be introduced `def get_current_site(request=None)` method in `allauth.utils`, and then change all lookups to use that one method?

tony:
@pennersr I will follow up tomorrow.

Best

tony:
@pennersr : How is it now?  Would you like it renamed to ``get_current_site``? I used ``get_current_site_model`` to get consistent with ``get_user_model``.

pennersr:
Thanks! If you could rename to `get_current_site` than I can merge it in. The method returns the site instance, whereas `get_user_model` actually returns the model. Hence the naming difference.

tony:
@pennersr Done

allauth/socialaccount/views.py
```diff
@@ -53,7 +52,7 @@ def form_valid(self, form):
 
     def get_context_data(self, **kwargs):
         ret = super(SignupView, self).get_context_data(**kwargs)
-        ret.update(dict(site=Site.objects.get_current(),
+        ret.update(dict(site=get_current_site(),
```
pennersr: Here you can pass `self.request` as well...

example/example/demo/models.py
```diff
@@ -13,9 +13,11 @@ def setup_dummy_social_apps(sender, **kwargs):
     `allauth` needs tokens for OAuth based providers. So let's
     setup some dummy tokens
     """
-    site = Site.objects.get_current()
+    request = None if 'request' not in kwargs else kwargs['request']
```
pennersr: How about simply: `request = kwargs.get('request')`

tony:
@pennersr now?

pennersr:
Great, thanks for tweaking ...

Issue #830, opened 2015-02-08
Django 1.8a1 issue with TEMPLATE_CONTEXT_PROCESSORS


orokusaki:
I'm receiving django.core.exceptions.ImproperlyConfigured: socialaccount context processor not found in settings.TEMPLATE_CONTEXT_PROCESSORS.See settings.py instructions here: https://github.com/pennersr/django-allauth#installation despite having added all the necessary context processors to the new-style `TEMPLATES['options']['context_processors]`. It functions when adding the old-style setting `TEMPLATE_CONTEXT_PROCESSORS`, but I'm not planning to use that setting any longer.

twcook:
Now that Django 1.8 has shipped is there any plan to fix this?

pennersr:
This is fixed in b1bce450

vaibhav-jain:
Even after `updating` I am getting this error. How to resolve this issue????

pawciobiel:
will this help https://github.com/pennersr/django-allauth/pull/950

vaibhav-jain:
Yeah.... It is fixed. Thanks

pennersr:
https://codeberg.org/allauth/django-allauth/issues/830

PR #950, opened 2015-05-17
another fix for context processor socialaccount for django >= 1.8

pawciobiel:
fix for fix https://github.com/pennersr/django-allauth/issues/830


vaibhav-jain:
Thanks

Issue #926, opened 2015-04-21
Django 1.8 Templates configuration


Koed00:
Django 1.8 introduced a new `TEMPLATES` settings:

```
TEMPLATES = [
    {
        'BACKEND': 'django.template.backends.django.DjangoTemplates',
        'DIRS': [
            'templates'
        ],
        'APP_DIRS': True,
        'OPTIONS': {
            'context_processors': [
                "django.template.context_processors.request",
                "allauth.account.context_processors.account",
                "allauth.socialaccount.context_processors.socialaccount",
                "django.template.context_processors.static",
                "django.contrib.auth.context_processors.auth",
            ],
        },
    },
```

This means that `TEMPLATE_DIR` , `TEMPLATE_CONTEXT_PROCESSORS` and `TEMPLATE_LOADERS` are being ignored when using the new configuration method.

Allauth will check if it's context processors are in `TEMPLATE_CONTEXT_PROCESSORS`, so it will either give a false positive or a false negative.

Also the Django Admin module now requires the  `django.contrib.auth.context_processors.auth` to be present in `context_processors`. I'm not sure if this affects the operation of the allauth account processors.

Koed00:
Similar to #830

luzfcb:
@Koed00 this issue was fixed on master.

@pennersr what is missing for a 0.20.0 release

pennersr:
Done

luzfcb:
@pennersr Thanks.

pennersr:
https://codeberg.org/allauth/django-allauth/issues/926

PR #951, opened 2015-05-17
Updated documentation for new TEMPLATE setting in Django 1.8

ka7eh:
See:
- https://docs.djangoproject.com/en/1.8/ref/settings/#templates
- https://github.com/pennersr/django-allauth/issues/926


pennersr:
Thanks, but this is already in (b710cf66f8e61a18f63a369ce2316f95e5337611).

Issue #963, opened 2015-05-28
TypeError: <function save at 0x7f9b4eab48c0> is not JSON serializable


owais:
I'm trying to write Slack provider plugin but just can't get to log in. I'm constantly seeing the error below. Oddly, I was able to login twice out of 20-30 attempts I made and I didn't really make any changes to the code.

This is also happening if I try to login using bundled LinkedIn oAuth2 provider. I've no clue what is going on. The same code and requirements work on my staging server just fine. I also reset my database to make sure it's is not corrupt data but that didn't fix it either.

No user, social account or social token is created. I can share the preliminary code I wrote for slack provider but I doubt that has got anything to do with it since it happens with the linkedin provider also.

I tried version 0.15, 0.20 and master. Same thing.

Any pointers?

```python
Environment:

Request Method: GET
Request URL: http://localhost:8000/accounts/slack/login/callback/?code=xxxxxxxxxx&state=xxxxxxx

Django Version: 1.7.7
Python Version: 2.7.3


Traceback:
File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py" in get_response
  111.                     response = wrapped_callback(request, *callback_args, **callback_kwargs)
File "/usr/local/lib/python2.7/dist-packages/allauth/socialaccount/providers/oauth2/views.py" in view
  55.             return self.dispatch(request, *args, **kwargs)
File "/usr/local/lib/python2.7/dist-packages/allauth/socialaccount/providers/oauth2/views.py" in dispatch
  125.             return complete_social_login(request, login)
File "/usr/local/lib/python2.7/dist-packages/allauth/socialaccount/helpers.py" in complete_social_login
  145.         return _complete_social_login(request, sociallogin)
File "/usr/local/lib/python2.7/dist-packages/allauth/socialaccount/helpers.py" in _complete_social_login
  161.         ret = _process_signup(request, sociallogin)
File "/usr/local/lib/python2.7/dist-packages/allauth/socialaccount/helpers.py" in _process_signup
  27.         request.session['socialaccount_sociallogin'] = sociallogin.serialize()
File "/usr/local/lib/python2.7/dist-packages/allauth/socialaccount/models.py" in serialize
  187.                    user=serialize_instance(self.user),
File "/usr/local/lib/python2.7/dist-packages/allauth/utils.py" in serialize_instance
  162.     return json.loads(json.dumps(ret, cls=DjangoJSONEncoder))
[...]

PR #967, opened 2015-06-02
Fixed a bug in utils.py that crashed the serializer if the user insta…

owais:
…nce returns a function as one of the attributes.


owais:
Fixed #963

pawciobiel:
:+1: nice one dude!
Adding a tiny test would help.

owais:
@pawciobiel Done. The test case `allauth.tests.BasicTests.test_serializer` check for this now. It fails on master but passes on this PR.

PR #970, opened 2015-06-05
New migration does not check ACCOUNT_UNIQUE_EMAIL setting

shaunbrady:


shaunbrady:
I realize that I'm changing a migration, but this would break for any one who had previously had the setting ACCOUNT_UNIQUE_EMAIL=False (our case).  I don't think it is possible just add a 0003 to fix this issue.  If someone has the case of ACCOUNT_UNIQUE_EMAIL=True, and applied this migration, I believe this change will be benign to them.

jezdez:
@pennersr This may be interesting to you as it breaks 0.20.0 right now

pennersr:
I agree with changing the migration in this case. It will require some explanation in the release notes though...

jezdez:
@pennersr Cool, thanks for merging. Let me know if I can help with anything to move along a bugfix release :+1:

Issue #618, opened 2014-05-16
Flickr Authorization


romdim:
Hello everyone!

I've been trying to authorize my flickr app through allauth all day unsuccessfully...
I keep getting "Oops! Flickr doesn't recognise the permission set." after I press "Ok, I'll authorize it" button...
Do you know what may be the problem?

I guess it's a permission problem, but don't know where and how it should be declared...
Any ideas?

Thanks in advance,
Romanos

pennersr:
https://codeberg.org/allauth/django-allauth/issues/618

PR #982, opened 2015-06-18
Allow user to specify extra auth parameters for Flickr (and generic OAuth) providers

pmrowla:
Flickr apparently requires an app to pass a 'perms' parameter when authenticating, even though the API documentation states that it is optional. Users can now configure this for flickr (and any other OAuth provider) with the AUTH_PARAMS setting. This works the same way as the existing functionality for OAuth2 providers.


pmrowla:
I just noticed that my commit message has the wrong issue number for the flickr perms, it should be #618.

PR #988, opened 2015-06-21
Update urls.py - delete url which will never be called

wimfeijen:
According to my knowledge, this url will never be called. I propose to delete it to avoid confusion.


pennersr:
This was necessary for backwards compatibility with < 0.14. It can indeed be removed by now, but simply remove it completely -- no need to keep it in commented.

Issue #994, opened 2015-06-26
Password Reset form doesn't work with UUID pks


yarbelk:
the block in ```accounts.utils.url_str_to_user_pk(s)``` is, as the comment states, Ugh.  with new UUID fields, it breaks on that test. to find its type.

NB - I think 1.8 moves to base 64 encoding

```
    def url_str_to_user_pk(s):
        User = get_user_model()
        # TODO: Ugh, isn't there a cleaner way to determine whether or not
        # the PK is a str-like field?
    try:
        User._meta.pk.to_python('a')
        pk = s
    except ValidationError:
        pk = base36_to_int(s)
    return pk
```

pennersr:
Fixed by #995

pennersr:
https://codeberg.org/allauth/django-allauth/issues/994

PR #995, opened 2015-06-26
Add UUID support for password reset

yarbelk:
Fix for #994 
Basically add some tests for the uuid field serialization in the urls, and force casting to string on the user_pk_to_url_str method

### DIFF
--- a/allauth/account/utils.py
+++ b/allauth/account/utils.py
@@ -8,6 +8,7 @@ except ImportError:
 import django
 from django.contrib import messages
 from django.core.urlresolvers import reverse
+from django.db import models
 from django.conf import settings
 from django.http import HttpResponseRedirect
 from django.utils import six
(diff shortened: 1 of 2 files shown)


yarbelk:
fixing python 2.6 ignoring of tests now (UUIDFields are not available in python2.6, so the one direction of test doesn't matter.)

JshWright:
Django 1.6 (and therefore Python 2.6) are no longer supported. Why not drop it entirely from the Travis config?

JshWright:
Since it seems kind of dodgy to hack around unsupported Django/Python versions in the test suite, I've opened PR1001

yarbelk:
it doubled the joy :)

going to squash them all down.

allauth/account/utils.py
```diff
@@ -362,16 +363,25 @@ def passthrough_next_redirect_url(request, url, redirect_field_name):
...
+    """
+    User = get_user_model()
+    if hasattr(models, 'UUIDField') and issubclass(type(User._meta.pk), models.UUIDField):
+        return user.pk.hex
+
     ret = user.pk
     if isinstance(ret, six.integer_types):
         ret = int_to_base36(user.pk)
-    return ret
+    return str(ret)
```
pennersr: This str() is superfluous right?

yarbelk: yes.  from a squashed merge before i realized that I had to return hex.  will fix it.

PR #1000, opened 2015-06-30
change redirect_ui to make Android cross client identity happy

jacobtopper:
This is a change that can make the code generated by Android for OAuth work. Using the standard logic here causes the incorrect redirect_uri value to be passed, which makes Google unhappy. This fix should not be pushed to production, because clearly it wipes out important functionality for other uses. This is a reference point only.


jacobtopper:
oops, that was not the fork I was looking for.

PR #1001, opened 2015-06-30
Don't test deprecated Django/Python versions

JshWright:
Perhaps this is a bit premature, but if you feel like dropping test coverage for deprecated Django versions, here's the merge button...


yarbelk:
+1

PR #1020, opened 2015-07-07
Added default values for a couple of Facebook provider settings

rapilabs:


pennersr:
Reminder before merging: the default has just been changed to v2.4...

PR #1021, opened 2015-07-07
Moving control of email confirmation into adapter layer

lukeburden:
This pull request allows full control over sending of confirmation emails through the standard approach taken in this project: adapter customization. In my specific case, this is critical such that I can defer preparation and sending of this mail to an async queue plus take more sophisticated approaches to confirmation, such as deep-linking into a mobile application.

Also, it removes any remaining reversal/knowledge of specifically named URLs out of the model layer allowing easier use of allauth in a pure API capacity without forcing dependent projects to set up dummy url configurations. For example, https://github.com/Tivix/django-rest-auth/blob/master/rest_auth/registration/urls.py#L22.

Thanks! Let me know if there's anything you'd like adjusted. :)

### DIFF
--- a/allauth/account/adapter.py
+++ b/allauth/account/adapter.py
@@ -5,10 +5,12 @@ import warnings
 import json
 
 from django.conf import settings
+from django.contrib.sites.models import Site
 from django.http import HttpResponse
 from django.template.loader import render_to_string
 from django.template import TemplateDoesNotExist
 from django.core.mail import EmailMultiAlternatives, EmailMessage
+from django.core.urlresolvers import reverse
 from django.utils.translation import ugettext_lazy as _
 from django import forms
 from django.contrib import messages
@@ -20,7 +22,8 @@ except ImportError:
 
 from ..utils import (import_attribute, get_user_model,
                      generate_unique_username,
-                     resolve_url, get_current_site)
+                     resolve_url, get_current_site,
+                     build_absolute_uri)
 
 from . import app_settings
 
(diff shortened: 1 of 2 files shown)


pennersr:
The first thing that comes to mind is that the adapter already has a `send_mail` method. Here, you can check on the type of mail being sent and take whatever action you see fit. I think it would be cleaner to expose just a `def get_activation_url(self, request, emailconfirmation)` (which is almost what #912 has implemented).

lukeburden:
There are two issues I'm trying to deal with here .. neither of which an approach such as that in #912 satisfies. Teasing them apart ..
1) Making allauth shine for both web apps and api usage is a bit tricky. Any url reversal needs to be easily avoidable or deferrable such that the API needn't fake url confs or be forced hardcode urls
2) It's nice to have the option to push work off to a queue. When you do, it's generally best practice to minimise the amount of data you need to send to your task queue for the task queue workers to be able to complete their tasks. For example, in this case, ideally you'd just send a task id saying "send confirmation email" and the pk of the EmailConfirmation object. The task handler is then free to pull it out of the DB and do as it needs, generating activation urls, doing the whole shebang.

The approach in #912 would encourage a developer in my situation to do one of two undesirable approaches for the deferral of sending email to a queue:
- within send_email, based on context detect what the intention is and queue up the appropriate remote task. If I wanted to send the EmailConfirmation pk, it'd mean using the context.user and context.key to look it up for inclusion. Kinda gross given we just had it a moment ago in the current worker.
- write a remote task that essentially replaces send_email. This means serializing potentially complex data in the context; fragile and not fun. It also means activation url reversal is still happening in my API layer and I either have to have a dummy URL conf to allow that, or not use reverse (urgh).

I think your instinct is good in preferring to expose small chunks of functionality but in this case just doing get_activation_url() probably isn't enough. It could be done as well, though, for added flexibility?

pennersr:
You would indeed lose out on having access to the confirmation pk, I overlooked that use case. So the best way forward is to do both, as for the more frequent use case `get_activation_url` suffices.

lukeburden:
Sounds good to me! And cheers for your hard work on the library, it's appreciated.

lukeburden:
Thanks for the merge @pennersr, shifting a project back to your repo now. Quick question - should I have tacked myself onto the AUTHORS list or is there a minimum contribution level before getting in there?

pennersr:
Sorry for the delay... you've been added now: https://github.com/pennersr/django-allauth/commit/7e980623a667658afaa2ce6d6d04a8f558ca8d32

PR #1040, opened 2015-07-20
Add a "login proxy" feature, useful for testing deploys without updating the OAuth provider redirect_uri whitelist; add a specific android callback to the Google Oauth2 provider

benbardin:
Hello - we've found these changes to be very helpful in our own development. Are you interested in incorporating them?

Best,
Ben


pennersr:
I am sorry, but I prefer not to pick this up, it convolutes things too much. I think most of this logic could be kept in a standalone package. As for the callback handling, if there is anything that can be done to make it easier for you to add this and avoid having to resort to forking/monkey patching, please let me know.

benbardin:
Not to worry, I get where your coming from. Will-do.

Issue #1002, opened 2015-06-30
Support Multiple providers of the same kind


yarbelk:
A very large feature request - Support multiple SocialApps of the same provider.
- in some places this will help clean up some of the code conceptually - using standard ORM FK instead of lots of helper methods and callbacks.
- Useful for those people who are running sites with a Facebook Business account and lots of apps (additional strangeness here b/c of unique UIDs).  Or are building api services that make use of the same.

If there is interest in this, I'd love to contribute.

ivanvenosdel:
I would like this feature as my site would like to offer two companion applications on the same social provider. So if a user signs in or connects to a provider using an app and then signs-in/connects a second app with the same provider it should re-use the same social account (which should be fine because social account has an FK to provider, not app). 

To me this all seems possible to implement without the need for model changes or making them swap-able. If I am right the feature would be far smaller in scope but am I missing something?

esabunor:
@ivanvenosdel Did you "put in the time" for this feature maybe i could use your fork or something. :)
@pennersr I would like this feature as well because as @yarbelk  said:

- Useful for those people who are running sites with a Facebook Business account and lots of apps (additional strangeness here b/c of unique UIDs). Or are building api services that make use of the same.

I willing to contribute a PR as well.

hishamkaram:
@pennersr  any progress on this issue?

skumarcm:
This feature may be an important one as more mobile applications are being developed based on the same webserver.
Hence there could be multiple mobile apps that allow users to login through the same provider (eg. facebook) but still want to deliver the service through the same webserver powered by allauth.

Is there a solution available for this already? Or any plans to integrate this feature with django allauth?

delton-ecometrica:
Has there been any update on this feature, to havet multiple social apps for the same provider

yarbelk:
I'm not involved in the project that gave rise to this.  I think its a great feature that someone should do, but I won't be the one to do it.  Feel free to have someone take over.

gboor:
Multiple apps for the same provider is one thing, but when talking about self-hosted providers - like Keycloak - having multiple providers isn't a bad idea either...

Or maybe just move the endpoint URL and realm for Keycloak into the config rather than setting it via env variables.

I might have a look at this in a few weeks, when I have some time.

derek-adair:
Seems valid, @pennersr seems like there are people interested in contributing and need guidance on this one if it's desirable or not.

Ramon-W:
[...]

PR #1044, opened 2015-07-23
Swappable SocialApp

yarbelk:
Make SocialApp swappable.  This should have no impact on existing installs.

This is currently at first pass with django 1.8 support (only tested on 1.8 ATM, haven't looked at other djangos yet).

Looking for feedback on how to improve this.  Note - its really rough right now and at the very least needs  a code cleanup (pep8 etc).

Also looking for feedback on older djangos (<1.8) and south migrations.


pennersr:
I am curious to learn what your use case for this is. Could you elaborate a bit?

yarbelk:
Sure. but Really Dont Merge it Yet even if its cool - this doesn't test both sides of the swapping (default and swapped).


I have busness logic that needs one-to-many (Product to SocialApp).  This is sort of related to issue #1002 By doing this, I can multiplex my providers based on information passed to me about a product.

By being able to swap out SocialApp I can side step most of the things I need to do to get multiple 'facebook' or 'g+' providers working, I only need to subclass the SocialLogin views to substitute the `current_app` logic.

Later I might put a more concerted effort into #1002 if it seems worthwhile.

pennersr:
I think supporting multiple apps of the same kind could be taken care of without having to completely swap out the `SocialApp` model. Currently, the logic for choosing the `SocialApp` to use only takes the current site (as in contrib.sites) as input.  Suppose you could completely override the selection of the app the use, wouldn't that be sufficient for your use case?

yarbelk:
Unfortunatly, They all are on the same domain, so that won't work.

pennersr:
Still, even if they are all in the same domain, suppose this method was overridable in the adapter:

yarbelk:
True - that lets you change the logic, but for this I'd need to create a join table, which since I'm migrating an existing user base of millions, I'd rather not have to manage (and join tables and normalization questions).

This being said, that is my fallback plan if I can't get the testing to work coherently.  There is no point of doing this if it doesn't benefit others, and no point of doing it if it can't be tested.

yarbelk:
Gave up trying to swap out the model during the running tests, rather, I added different settings for swapped models, and rerun all the tests.

then I combine the output of the coverage for both.

yarbelk:
Ok.  Tested, passing.

though the  only way to do this was run all the tests with the default SocialApp, and then rerun them with a swapped app (along with some special swapping tests). 

handled the coverage as well, combines the output of both.

yarbelk:
Lets table this for now - I realized this is a larger kettle of ponies to boil that intended.

For this to really make sense, I need to be able to link more things together with relations than is currently possible.  As I need this locally, i'm implementing more of the swapability, and when everything works like happyness, I'll reopen this.

PR #1046, opened 2015-07-23
Update password_reset_from_key.html

gabn88:
It does not work because of the dot. So remove dot.

### DIFF
--- a/allauth/templates/account/password_reset_from_key.html
+++ b/allauth/templates/account/password_reset_from_key.html
@@ -11,7 +11,7 @@
         <p>{% blocktrans %}The password reset link was invalid, possibly because it has already been used.  Please request a <a href="{{ passwd_reset_url }}">new password reset</a>.{% endblocktrans %}</p>
     {% else %}
         {% if form %}
-            <form method="POST" action=".">
+            <form method="POST" action="">
                 {% csrf_token %}
                 {{ form.as_p }}
                 <input type="submit" name="action" value="{% trans 'change password' %}"/>


pennersr:
An empty action is not allowed by the HTML5 standard and will cause validation errors. `action="."` will work, unless perhaps you've got `APPEND_SLASH` set to `False` (defaults to `True`) *and* you are on a URL that does not end with a slash?

For reference, see: https://stackoverflow.com/questions/1131781/is-it-a-good-practice-to-use-an-empty-url-for-a-html-forms-action-attribute-a

Perhaps this needs to be changed to `action="{% url ... %}"` to bypass this discussion...

gabn88:
I have append slash to False indeed! I will put the url there... What is it? {% url 'reset_password_from_key' key %} or what is the name of the key parameter?

gabn88:
well, for now... Let's simply put {{request.path}} there if you agree?

pennersr:
Unfortunately, the template does not have access to the uidb36, so you cannot reconstruct the url with {% url %}. I don't like the request.path approach.

Btw, even if you have append slash disabled, if you use proper URL reversal everywhere you should never end up on the password reset page without having a `/` at the end, unless you are manually typing in the url.

gabn88:
The problem is not that I don't have a / at the end, I have! But the problem is that even with the / at the end, the POST action (".") does not work and goes into the root of the page! Maybe because I have set <base href="/"> for my hidden angular application? Anyway, without the dot is works for me..

Issue #735, opened 2014-10-11
Automatically Sign In on Password Reset


jgadelange:
When a password reset link is clicked you are able to set the password for a certain account. Afterwards you still have to sign in with your credentials (which you just entered). This feels unnecessary since you clearly know the password for the account. 
Is there a reason why this is not done currently?

SteveByerly:
I don't think either way is inherently wrong, but I think the reasoning for this is the expectation of the user. 

When I change my password, I sort of expect to have to log in since I never actually logged in the first time. If you automatically log a user in after changing their password, you introduce an element of surprise to the user - they may not even realize they are now logged in.

I think that you could attain this functionality, though, by simply customizing the reset form or adapter.

julen:
I'm facing the same situation as the OP, and as I see it's not possible to implement this behavior by customizing the form or the adapter — the `PasswordResetFromKeyView` view needs to be adjusted.

Doing this in 3rd party code is possible but a mess at the same time because you have to override the URL pattern to route to your custom view. Doing it directly in allauth looks simple OTOH: conditionally implement the requested behavior behind a setting in [`form_valid()`](https://github.com/pennersr/django-allauth/blob/ef8fe4ec2d7013e517bd1c480834cf2bcb131545/allauth/account/views.py#L592-L600).

@pennersr would you be willing to accept such a PR?

SteveByerly:
You could probably log the user in by hooking into the signal that is that is sent in the form_valid method of ```PasswordResetFromKeyView```:

``signals.password_reset``

If you do this, make sure you update the template, ```password_reset_from_key_done.html``` so that you inform the user they are now logged in.

julen:
Oh, undocumented signals :) I'm creating a PR for those as a start.

Allowing this flow in Allauth directly wouldn't hurt IMO. Moreover, there's already the `ACCOUNT_LOGIN_ON_EMAIL_CONFIRMATION` setting for a similar behavior. Let's hear from @pennersr.

SteveByerly:
@julen @pennersr I guess I'm generally a fan of not adding settings if they can be avoided. Settings like these build up and create a ton of forks in the logical flow of the code. AllAuth is already very easy to extend for your own needs. I think that it's power is in the ability to take the framework built in and adapt it if and when you have to.

julen:
Yes I agree on the point of setting bloat, but in this specific case my understanding is that it aligns with what Allauth currently has.

julen:
[...]

PR #1067, opened 2015-07-29
Added option to automatically sign in after password reset

julen:
As discussed in #735, this PR adds the option to automatically sign users in after they have reset their password successfully.

Based on the opinions on the issue, the default for `LOGIN_ON_EMAIL_CONFIRMATION` has also been adjusted to `False`.


pennersr:
Thanks!

PR #1078, opened 2015-08-05
Auth: fixed `perform_login()` call

julen:
The reference to `request` was incorrect (I had a local variable at the time of creating the patch, sorry about that). Note this prevents the `LOGIN_ON_PASSWORD_RESET` from working properly, so might as well need a release.


pennersr:
This is already addressed in some other commit ..

PR #1082, opened 2015-08-11
fix for error in performing login functionality on password reset with LOGIN_ON_PASSWORD_RESET set to True

igorgai:


julen:
This is a dupe of #1078.

igorgai:
Right, sorry about that. Anyway it would be nice is if any of these two was merged.

PR #1087, opened 2015-08-17
added signal log out

patricio-astudillo:
There is now a signal that signals when a user is logging out.

### DIFF
--- a/allauth/account/views.py
+++ b/allauth/account/views.py
@@ -653,6 +653,11 @@ class LogoutView(TemplateResponseMixin, View):
         get_adapter().add_message(self.request,
                                   messages.SUCCESS,
                                   'account/messages/logged_out.txt')
+        signal_kwargs = {}
+        signals.user_logged_out.send(sender=self.request.user.__class__,
+                                    request=self.request,
+                                    user=self.request.user,
+                                    **signal_kwargs)
         auth_logout(self.request)
 
     def get_context_data(self, **kwargs):
--- a/AUTHORS
+++ b/AUTHORS
@@ -35,6 +35,7 @@ Guoyu Hao
 Hatem Nassrat
 J. Fernando Sánchez
 Jack Shedd
+James Rivett-Carnac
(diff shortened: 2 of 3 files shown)


pennersr:
Naming-wise the signal suggests that the user is logged out, but actually that has not happened yet at the time the signal is fired. So we need to move it below the auth_logout call, and add a note to signals.rst

allauth/account/views.py
```diff
@@ -654,6 +654,12 @@ def logout(self):
                                   messages.SUCCESS,
                                   'account/messages/logged_out.txt')
         auth_logout(self.request)
+        signal_kwargs = {}
+        signals.user_logged_out.send(sender=self.request.user.__class__,
+                                    request=self.request,
+                                    user=self.request.user,
```
pennersr: `auth_logout` sets `request.user` to the anonymous user... so we lose the user that logged out. We need to preserve the user before `auth_logout` and use that here.

entvia: Hi guys! Any decision on this issue?

pennersr: There is not really a decision to take, we need to make sure that the user that was logged out is available as a signal parameter.. when that is taken care of this can be merged.

entvia: Thank you for reply. Can't we send signal before 
`adapter.logout(self.request)` or put it into adapter before `django_logout(request)`?
In this case we still have user in request. The signal could be called 'pre_logout' then.

pennersr: I tend to prefer keeping signals simple, so preferably no pre/post logged out versions. If you need that kind of control, you can already override the adapter and insert pre/post logic as you wish. So, the only thing that needs to be done here is firing a `user_logged_out` signal like so (pseudo code):

```
user = self.request.user
logout()
signals.user_logged_out.send(...user=user)
```

entvia: Thank you, Raymond!

PR #1090, opened 2015-08-17
Fix LOGIN_ON_PASSWORD_RESET

julienp:
Fixes a typo that results in an error when enabling ACCOUNT_LOGIN_ON_PASSWORD_RESET


julienp:
Should have checked first, there's already a pull request to fix this: https://github.com/pennersr/django-allauth/pull/1078 the test in this request might still be useful though.

PR #1094, opened 2015-08-18
Speedup travis: Save pip cache, install requirements as wheels

moreati:
Travis allows saving cache directories between job runs. This can eliminate
the need for pip to download/build requirements every time.

Additionally recent versions of pip will build requirements as wheels, and
cache the built wheel for future installs.

I've also taken the opportunity to
 - Remove uses of the deprecated --use-mirror switch
 - Install all requirements in a single transaction

Together these should speedup each job by 10-15 seconds.

References
 - http://docs.travis-ci.com/user/caching/#pip-cache

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,15 +10,17 @@ env:
    - DJANGO="Django<1.7"
    - DJANGO="Django<1.8"
    - DJANGO="Django<1.9"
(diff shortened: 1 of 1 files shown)


pennersr:
Picked up as part of #1185

PR #1096, opened 2015-08-19
added weixin provider

harmy:
add an provider for weixin, which is popular in China.


pennersr:
Could you please elaborate a bit on the code in utils.py ? What's the goal....why are the python/django counterparts not a viable option?

pennersr:
I will move ahead and integrate this code. I'll drop the utils.py though -- let me know if that breaks things, and in what way. Furthermore, I am leaving out the unicode related changes to the username handling. Let's discuss that as part of a different PR.

harmy:
Great,I've done some testing, everything works.

Issue #369, opened 2013-08-30
Github email issue


horva:
Allauth should query https://api.github.com/user/emails endpoint when scope is set to user:email (granting readonly access to user emails).
First it can search for verified emails (will be available soon in version v3 on /user/emails endpoint) and have fallback on other (unverified addresses).
In the end if no emails were retrieved user is presented with /account/social/signup/ page.

nokome:
+1 . In lieu of a fix, can anyone suggest a workaround?

fisadev:
Adding more information to this: there is a possible security problem related to this bug.
If the developer has ``SOCIALACCOUNT_EMAIL_VERIFICATION = 'none'`` (which is probably very common because we trust the social providers to verify the users emails), but then a user tries to login with github and a private email, then allauth asks the user the email address and does not verify it. Thus, users can inform false emails and give the impression they are verified.

fabiocerqueira:
Someone fix it? Someone suggests some workaround to resolve this without overwriting the provider?

python-social-auth do this requesting 'https://api.github.com/user/emails'
https://github.com/omab/python-social-auth/blob/master/social/backends/github.py#L41

sanketsaurav:
If anyone's still looking for a solution, you need to request for the `user:email` scope while authorizing. Add this to your settings:

```
SOCIALACCOUNT_PROVIDERS = {
    ... <other providers>...
    'github' : {
        'SCOPE': ['user:email']
    }
}
```

This should fix it.

pennersr:
Done (88e9690701ffd0f213175b015bc162a8abe03951)

pennersr:
https://codeberg.org/allauth/django-allauth/issues/369

PR #1097, opened 2015-08-19
Fix issue #369 - Github private emails

fabiocerqueira:
Github do not retrieve private emails through of profile. Then we need get it using: [/user/emails](https://developer.github.com/v3/users/emails/)


pennersr:
This should be done conditionally, based on `SOCIALACCOUNT_QUERY_EMAIL`

pennersr:
Thanks!

Issue #381, opened 2013-09-14
Dynamically extend OAuth2 scope?


ghost:
This is a feature request, or an oversight on my part.

My application relies on Facebook login, and in accordance with FB guidelines, it starts with a minimum set of permissions. When a user triggers an action that requires additional permissions, the app should request them.

As far as I can see from the readme / google / bug reports, the OAuth scope is hard-wired in the settings.

So, I came up with this:

```
from allauth.socialaccount.providers.facebook.provider import FacebookProvider
from allauth.socialaccount.providers.facebook.views import FacebookOAuth2Adapter
from allauth.socialaccount.providers.oauth2.views import OAuth2LoginView
from allauth.socialaccount.models import SocialApp
from django.conf import settings
from django.http import HttpResponseRedirect
from facepy import GraphAPI
from functools import wraps
from django.contrib.auth.decorators import login_required
from django.core.exceptions import PermissionDenied


def first( l, default=None):
    'Pick the first item of a sequence'
    return l[0] if l else default


def fb_account( user):
    'Facebook account adapter for a user'

    # Do not use get here because not everyone has a FB account
    return first( user.socialaccount_set.filter( provider='facebook'))


def get_graph( user=None):
    ''' Get a GraphAPI object.
        If user is not None, authenticate as user.
    '''

    if user:
            return GraphAPI( token)


class DynamicScopeProvider( FacebookProvider):
    'Facebook provider for allauth with dynamic scope'

    def __init__( self, scope):
        self.scope = scope
        super( DynamicScopeProvider, self).__init__()

    def get_scope( self):
        return set( self.scope).union( set( self.get_default_scope()))


class DynamicScopeOAuth2Adapter( FacebookOAuth2Adapter):
    'Custom adapter for allauth to enable dynamic Facebook scopes'

    def __init__( self, scope):
        self.provider = DynamicScopeProvider( scope)

    def get_provider( self):
        return self.provider


def fb_permission_required( permissions, login_url=None):
    ''' Decorator to check whether Facebook permissions are sufficient
        to execute a view. If additional permissions are needed,
        they are requested from Facebook.

        Permissions can be either a sequence of permission names
        (e.g. ['user_email', 'user_location']), or a comma-separated
        string (e.g. 'user_likes,friends_likes').
    '''

    def get_permissions( user):
        'Get current Facebook permissions for a user'
        bogus = ('installed', ) # FB returns this but will not accept it
[...]

PR #1104, opened 2015-08-25
Avoid OAuth2 scope duplicates (closes #1065)

### LINKED ISSUE #1065: Twitch
Hy guys,

i use allauth with Facebook and it works very well.

Now i try to add twitch and i can't get it to work i get everytime only the error that something with the social-network was not working.

How can i get a more detail error?
I found no logs.

Thanks a lot

Tobit

mxscho:
This is a fix for issues discussed here: https://github.com/pennersr/django-allauth/issues/1065
I only tested it with Twitch, but as long as there is no use case/application where scope duplicates actually *do* something - it should work for all other providers, too.
This bug might have been introduced unintentionally by adding this dynamic scope feature here: https://github.com/pennersr/django-allauth/pull/761 (see also the related feature suggestion: https://github.com/pennersr/django-allauth/issues/381)

It's the shortest possible fix I could think of without breaking anything - at least I hope so. (You could also do things like changing the scope setting to a set instead of a list and stuff...)
However, I'm always open to criticism.

### DIFF
--- a/allauth/socialaccount/providers/oauth2/provider.py
+++ b/allauth/socialaccount/providers/oauth2/provider.py
@@ -1,3 +1,5 @@
+import re
+
 try:
     from urllib.parse import parse_qsl
 except ImportError:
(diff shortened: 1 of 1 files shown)


pennersr:
This does not prevent duplicates if the scope list contains entries that are already present in the default scope list. Also, this PR adds a feature to split scopes on "|". Wondering why that is essential for this PR, as this may require some more thought (e.g. shouldn't the separator be provider dependent...)

pennersr:
Duplicates are prevent now via 77310eaa

Issue #681, opened 2014-07-30
clean_<fieldname> on user SignupForm registration


humitos:
I've configured my project like this:

```
ACCOUNT_SIGNUP_FORM_CLASS = 'chap.challenge.forms.UserSignUpForm'
```

Then, my form class is something like this:

```
class UserSignUpForm(forms.ModelForm):

    [...]

    gender = forms.ChoiceField(
        choices=settings.GENDER_CHOICES,
        widget=forms.RadioSelect,
        required=False)

    class Meta:
        model = User

        fields = (
            'username', 'email', 'password1', 'password2',
            [...]
        )

      [...]

      # README: "clean_email" is not being called because BaseSignupForm
      # (from django-allauth) inherit from this one and overrides the
      # "clean_email" method so their method is called instead of our
      # allauth/account/forms.py#L208
      def clean_email(self):
          email = self.cleaned_data['email']
          if ValidateThisEmailFunction(email):
              raise forms.ValidationError('Not Valid.')
          return email
```

Here is the line with the issue:

https://github.com/pennersr/django-allauth/blob/master/allauth/account/forms.py#L257

I think this is a problem because there is no way to validate the fields in the way I want. Maybe, we should include `super(BaseSignupForm, self).clean_email()` in the `BaseSignupForm.clean_email` method as we are doing here:

https://github.com/pennersr/django-allauth/blob/master/allauth/account/forms.py#L296

k4ml:
Overriding `clean_email` on adapter won't do much as it's not able to short circuit the logic. The subsequent logic still being run despited whatever you did in adapter.clean_email().
```
def clean_email(self):
        value = self.cleaned_data["email"]
        value = get_adapter().clean_email(value)
        if app_settings.UNIQUE_EMAIL:
            if value and email_address_exists(value):
                self.raise_duplicate_email_error()
        return value
```
For example we want to allow user to re-attempt signup using the same email address, as long as the email address not yet validated but seem impossible because of `if value and email_address_exists(value)` and it can't be overriden.

k4ml:
Is it possible if all the `clean_*` methods in `BaseSignupForm` being moved out into mixin so that if user also implement clean_* in their custom form, that would take precedence ?

pennersr:
https://codeberg.org/allauth/django-allauth/issues/681

PR #1105, opened 2015-08-26
split BaseSignupForm methods into mixin so custom class can override …

k4ml:
…it - https://github.com/pennersr/django-allauth/issues/681#issuecomment-134518899

### DIFF
--- a/allauth/account/forms.py
+++ b/allauth/account/forms.py
@@ -202,8 +202,25 @@ def _base_signup_form_class():
                 'The custom signup form must implement a "signup" method')
     return fc_class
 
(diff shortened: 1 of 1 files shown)


pennersr:
The current base signup form handling encapsulates some internals for which I am not sure it is wise to turn this into a public API. In order to tackle that comment, I would be more in favor of moving the email_address_exists check into the clean_email of the adapter as well. That way, there would be one clear way of overriding things without exposing too much of internal form handling.

pennersr:
Closing.. alternative approach is needed, see comment.

PR #1111, opened 2015-08-29
Add a `related_name` on `EmailAddress`

MickaelBergem:
This allow us to access the email addresses of a given user by accessing the new `email_addresses` property of this user:

    user = User.objects.first()
    print(user.email_addresses.all())


pennersr:
You can already do that, using `user.emailaddress_set.all()`.
While the name could be improved upon, changing it causes backwards incompatibility that is just not worth it. So I prefer not to pick this up ...

PR #1134, opened 2015-09-24
Use render instead of render_to_response

### LINKED ISSUE #1133: Can't use Django 1.8's Jinja2 template backend
In Django 1.7, I was able to substitute Jinja2 templates for almost all django-allauth pages, so that authentication pages could be styled the same way as the rest of my content.  In Django 1.8, almost all can be substituted except for ``socialaccount/authentication_error.html``, and ``openid/login.html``, because of how the views render the templates.

In the case of ``socialaccount/authentication_error.html``, the call is in the file``allauth/socialaccount/helpers.py``, in ``render_authentication_error()``:

```python
        return render_to_response(
            "socialaccount/authentication_error.html",
            context, context_instance=RequestContext(request))
```

Using ``context_instance`` triggers the deprecated code path in Django's [template/loader.py](https://github.com/django/django/blob/master/django/template/loader.py#L84-L86), which includes removing template backends (like Jinja2) that don't support the legacy parameters.

jwhitlock:
render also wraps the context in a RequestContext, but works with the Jinja2 template backend in Django 1.8. Fixes #1133.

See [source of shortcuts.py](https://github.com/django/django/blob/master/django/shortcuts.py), and [deprecation note for context-instance](https://docs.djangoproject.com/en/1.8/releases/1.8/#dictionary-and-context-instance-arguments-of-rendering-functions).

### DIFF
--- a/allauth/socialaccount/helpers.py
+++ b/allauth/socialaccount/helpers.py
@@ -1,8 +1,7 @@
 from django.contrib import messages
 from django.contrib.auth import logout
-from django.shortcuts import render_to_response, render
+from django.shortcuts import render
 from django.http import HttpResponseRedirect
-from django.template import RequestContext
 from django.forms import ValidationError
 from django.core.urlresolvers import reverse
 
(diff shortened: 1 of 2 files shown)


jwhitlock:
Checks are failing because coverage 4.0 (recently released) doesn't work with Python 3.2

PR #1135, opened 2015-09-27
Fix Python3 related issues

avsd:
This is minimum changes to make it work under Python 3:

* Importing mock library for different Python versions
* Explicit relative imports.

However, I still could not make it work with Django 1.9. It raises a `KeyError` exception:

screen shot 2015-09-27 at 13 11 43

### DIFF
--- a/allauth/account/templatetags/account_tags.py
+++ b/allauth/account/templatetags/account_tags.py
@@ -3,4 +3,4 @@ import warnings
 warnings.warn("{% load account_tags %} is deprecated, use {% load account %}",
               DeprecationWarning)
 
-from account import *  # noqa
+from .account import *  # noqa
--- a/allauth/account/tests.py
+++ b/allauth/account/tests.py
@@ -27,7 +27,11 @@ from .adapter import get_adapter
 from .utils import url_str_to_user_pk, user_pk_to_url_str
 
 import uuid
-import mock
+
+try:
+    from unittest import mock
+except ImportError:
+    import mock
 
 
 @override_settings(
--- a/allauth/__init__.py
+++ b/allauth/__init__.py
@@ -8,7 +8,7 @@
 
 """
 
-VERSION = (0, 23, 0, 'final', 0)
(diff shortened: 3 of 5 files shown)


pennersr:
What issue are you tackling here? allauth already works fine with python3...:

https://travis-ci.org/pennersr/django-allauth

allauth/account/tests.py
```diff
@@ -27,7 +27,11 @@
 from .utils import url_str_to_user_pk, user_pk_to_url_str
 
 import uuid
-import mock
+
+try:
+    from unittest import mock
+except ImportError:
+    import mock
```
avsd: @pennersr this isn't an issue for Travis CI (probably because `mock` library is installed there), but if you try with a clean Python 3 environment, `mock` can't be imported. See also `setup.py`, lines 103-107.

allauth/account/templatetags/account_tags.py
```diff
@@ -3,4 +3,4 @@
 warnings.warn("{% load account_tags %} is deprecated, use {% load account %}",
               DeprecationWarning)
 
-from account import *  # noqa
+from .account import *  # noqa
```
avsd: @pennersr this (and a corresponding fragment below) is somehow not covered in unittests, but if you try this command in Python 3, it will fail, because implicit relative imports are not supported anymore:

```
from account.templatetags import account_tags
```

pennersr: True -- the `_tags` files have been deprecated since 2012 and can be simply removed by now..

avsd: Ah, in that case a PR is probably required here: https://github.com/pydanny/cookiecutter-django/blob/master/%7B%7Bcookiecutter.repo_name%7D%7D/%7B%7Bcookiecutter.repo_name%7D%7D/templates/account/email.html#L5

avsd: hmm strange... these `_tags` are not imported explicitly anywhere, but Django still tries to import them...

PR #1137, opened 2015-09-28
Added Reddit Oauth2 functionality

wayward710:
Added new support for Reddit Oauth2 functionality


pennersr:
Thanks for the pull! I am wondering if we can improve things in such a way that the RedditOauth2Client is no longer required? For example:

- Headers: support for this can be added in a generic fashion. E.g. add a property headers={} to the standard OAuth2Adapter, override it to  `headers = {"User-Agent": "django-allauth-header"}` in the RedditOAuth2Adapter, and use it accordingly.

- Similarly, we can add a `def get_auth_headers(self)` to the OAuth2Adapter (we'll need to make sure adapter.app is filled with the current SocialApp though).

wayward710:
Thanks, I'll look into it.

pennersr:
I would like to get rid of the `RedditOauth2Client`... we can add additional parameters to `RedditOauth2Adapter` such as `request_headers` that the standard `OAuth2Client` uses to make its requests.

wayward710:
I added some functionality to the Oauth2Client class -- custom header, scope delimiter (since there was an issue where someone said that a provider required a comma rather than a space), and ability to specify whether to send the client ID/secret as parameters or as an HTTPBasicAuth header (as some providers require).  It currently seems to be passing with Python 2 (which is what I'm using), but it's having a problem with Python 3.  I can look into it.

wayward710:
Fixed issue that was making code break in Python 3.  Got rid of RedditOauth2Client and added to Oauth2Client class to support custom headers and sending client ID and secret via HTTP Basic Auth (rather than as parameters).

avinassh:
Any update on this?

wayward710:
I updated my code with the requested changes from Raymond and submitted a pull request in November.  If there's anything else I need to do, I'm happy to do it.

pennersr:
Thanks for picking this up. I've made some changes to your code, making things a bit more DRY and avoiding the need to fork the OAuth2 views. This has been integrated now. I must admit, I did not test these changes in practice, so I would appreciate it if you could give a heads up if these changes are running fine on your end.

PR #1141, opened 2015-10-02
Moving account user stashing and logout logic to the DefaultAccountAdapter

lukeburden:
I am hitting issues using allauth in a pure API capacity. I do not support sessions/session auth on the API, only token-based auth. The problem is that allauth assumes the availability of `request.session` in several places.

In most instances I've been able to customize my adapters to avoid this dependency on sessions without loss of functionality. The rest are interactions with sessions in utility functions, which I cannot easily override.

This change is just shifting a little logic around to allow me to avoid hitting sessions in those remaining cases.

### DIFF
--- a/allauth/account/adapter.py
+++ b/allauth/account/adapter.py
@@ -44,10 +44,10 @@ class DefaultAccountAdapter(object):
         request.session['account_verified_email'] = None
         return ret
 
-    def stash_account_user(self, request, user):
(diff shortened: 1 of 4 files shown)


pennersr:
Looks good, I would only suggest renaming `(un)stash_account_user` to simply `(un)stash_user`

lukeburden:
Just a bump in case you weren't notified of my update.

PR #1144, opened 2015-10-06
Added missing to_str to FoursquareAccount - fixes #1125

### LINKED ISSUE #1125: Foursquare Name
Hello the foursquare connect ist connecting fine but is displaying Foursquare as the connected name. How can i change that?

LuanP:


LuanP:
@bspotted not sure why, can you spot where's the issue?

bspotted:
@LuanP 
Yes, foursquare has no name value. 
You have to either use firstName or lastName. Which means it should say:


return self.account.extra_data.get('firstName', dflt)

LuanP:
Ok. I'll change it, but where's the 'firstName'? I've used 'name' because of the FoursquareProvider :'( I thought it would use it from there.

bspotted:
It uses the return values from, foursquare. They are in the account model and saved in the extra_data instance. Didn't know myself ^^ until i had to check and fix it.

LuanP:
Thanks a lot, I'll do that tonight, if you need any help with git you can ask, I'd glad to help.

PR #1158, opened 2015-10-19
Document existing views

martinsvoboda:
When I started with django-allauth I miss documentation of prepared views. So I write it down.


pennersr:
Thanks for extending the documentation! I am marking this as "Needs work" because before this can be merged in some effort is required to improve upon the English correctness of the sentences here and there ...

PR #1166, opened 2015-10-29
Increase field sizes for Tokens

joshowen:


pennersr:
Can you let me know which provider required larger field sizes?

joshowen:
It is for a new provider I'm working on.  I will create a pull request for the provider once their API is publicly available.

PR #1167, opened 2015-10-30
add closing singup by is_open_for_signup in the document

sih4sing5hong5:


pennersr:
Thanks.. minor nitpick.. True/False need to be marked with backticks

PR #1168, opened 2015-10-30
isolate the populate_username functions from DefaultSocialAccountAdapter

sih4sing5hong5:
the original flows of `DefaultSocialAccountAdapter.save_user` are two:

1. `DefaultSocialAccountAdapter.save_user` => `DefaultAccountAdapter.populate_username`
2. `DefaultSocialAccountAdapter.save_user` => `DefaultAccountAdapter.save_user` => `DefaultAccountAdapter.populate_username`

So I merged the first flow to the second one. 

The final flow will only be

* `DefaultSocialAccountAdapter.save_user` => `DefaultAccountAdapter.save_user` => `DefaultAccountAdapter.populate_username`

The `DefaultSocialAccountAdapter.save_user` only calls `DefaultAccountAdapter.save_user` whether `form` is `None` or not.

### DIFF
--- a/allauth/account/adapter.py
+++ b/allauth/account/adapter.py
@@ -192,7 +192,7 @@ class DefaultAccountAdapter(object):
         """
         from .utils import user_username, user_email, user_field
 
-        try:
+        if form:
             data = form.cleaned_data
             first_name = data.get('first_name')
             last_name = data.get('last_name')
@@ -208,7 +208,7 @@ class DefaultAccountAdapter(object):
                 user.set_password(data["password1"])
             else:
                 user.set_unusable_password()
-        except AttributeError:
+        else:
             user.set_unusable_password()
         self.populate_username(request, user)
         if commit:
--- a/allauth/account/adapter.py
+++ b/allauth/account/adapter.py
@@ -192,20 +192,23 @@ class DefaultAccountAdapter(object):
         """
         from .utils import user_username, user_email, user_field
 
-        data = form.cleaned_data
-        first_name = data.get('first_name')
-        last_name = data.get('last_name')
-        email = data.get('email')
-        username = data.get('username')
-        user_email(user, email)
-        user_username(user, username)
-        if first_name:
-            user_field(user, 'first_name', first_name)
-        if last_name:
-            user_field(user, 'last_name', last_name)
-        if 'password1' in data:
-            user.set_password(data["password1"])
-        else:
+        try:
+            data = form.cleaned_data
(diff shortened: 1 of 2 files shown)


sih4sing5hong5:
How about making `populate_username` private, using `_populate_username` instead of?

It is only used by `DefaultAccountAdapter.save_user` now .

pennersr:
I prefer to leave `populate_username` as is... changing it breaks backwards compatibility without any good reason.

allauth/account/adapter.py
```diff
@@ -192,20 +192,23 @@ def save_user(self, request, user, form, commit=True):
...
+            user_username(user, username)
+            if first_name:
+                user_field(user, 'first_name', first_name)
+            if last_name:
+                user_field(user, 'last_name', last_name)
+            if 'password1' in data:
+                user.set_password(data["password1"])
+            else:
+                user.set_unusable_password()
+        except AttributeError:
```
pennersr: I am not very fond of using a try/except around a large block of code while catching a very generic exception. I would prefer to avoid the try altogether and keep exceptions for exceptional situations, not for normal code flow. So I would prefer to introduce `if form:` instead.

allauth/socialaccount/adapter.py
```diff
@@ -61,10 +61,7 @@ def save_user(self, request, sociallogin, form=None):
         """
         u = sociallogin.user
         u.set_unusable_password()
```
pennersr: This is no longer needed...

pennersr:
I have given this some more thought, I and think it is better not to streamline the flows as is done here. 

Rationale:

- It breaks backwards compatibility -- all of a sudden the `form` parameter can be `None` which previously was not the case.

- I prefer to keep a strict separation in responsibilities between the `account` and `socialaccount` apps. The `account` adapter should remain independent of knowledge with respect to social related scenario's. The changes made here move knowledge about the social signup flow (auto signup where form can be `None`) into the `account` module.

I am sorry this did not occur to me earlier...

sih4sing5hong5:
For backwards compatibility, it is only in the situation with `save_user(self, request, user, None)`.
In the previous version, the function call will get `Exception`. It is bug for this version.
In the new one, it will populate the username and pass.
I think it is OK for compatibility.

I modified the flows because they were too complexity. `socialaccount` app used `account`'s function too more. It is hard to custom user models and write docs. We can write the docs for keeping the knowledge if using the new flow. I think the new flow is better for maintaining in both code and docs.

PR #1169, opened 2015-11-01
Check if value not in blacklist during unique username generation

### LINKED ISSUE #1150: ACCOUNT_USERNAME_BLACKLIST and populate_username
I've noticed that populate_username function is not working as desired.

It doesn't check ACCOUNT_USERNAME_BLACKLIST while populating account username.

https://github.com/pennersr/django-allauth/blob/1b59ba0f622bc7fc151509cd9cef8f76d28d8f83/allauth/account/adapter.py#L158

sasha0:
Fixes #1150

### DIFF
--- a/allauth/account/adapter.py
+++ b/allauth/account/adapter.py
@@ -214,7 +214,7 @@ class DefaultAccountAdapter(object):
             user.save()
         return user
 
-    def clean_username(self, username):
+    def clean_username(self, username, shallow=False):
         """
         Validates the username. You can hook into this if you want to
         (dynamically) restrict what usernames can be chosen.
@@ -229,16 +229,19 @@ class DefaultAccountAdapter(object):
         if username.lower() in username_blacklist_lower:
             raise forms.ValidationError(_("Username can not be used. "
                                           "Please use other username."))
-        username_field = app_settings.USER_MODEL_USERNAME_FIELD
-        assert username_field
-        user_model = get_user_model()
-        try:
-            query = {username_field + '__iexact': username}
-            user_model.objects.get(**query)
-        except user_model.DoesNotExist:
-            return username
-        raise forms.ValidationError(_("This username is already taken. Please "
-                                      "choose another."))
+        # Skipping database lookups when shallow is True, needed for unique username generation.
+        if not shallow:
+            username_field = app_settings.USER_MODEL_USERNAME_FIELD
+            assert username_field
+            user_model = get_user_model()
+            try:
+                query = {username_field + '__iexact': username}
+                user_model.objects.get(**query)
+            except user_model.DoesNotExist:
+                return username
+            raise forms.ValidationError(_("This username is already taken. Please "
+                                          "choose another."))
+        return username
 
     def clean_email(self, email):
         """
--- a/allauth/utils.py
+++ b/allauth/utils.py
@@ -27,6 +27,8 @@ except:
 
 
 def _generate_unique_username_base(txts, regex=None):
+    from .account.adapter import get_adapter
+    adapter = get_adapter()
     username = None
     regex = regex or '[^\w\s@+.-]'
     for txt in txts:
@@ -43,8 +45,12 @@ def _generate_unique_username_base(txts, regex=None):
         username = username.split('@')[0]
         username = username.strip()
         username = re.sub('\s+', '_', username)
-        if username:
+        # Finally, validating base username without database lookups etc.
+        try:
+            username = adapter.clean_username(username, shallow=True)
             break
+        except ValidationError:
+            pass
     return username or 'user'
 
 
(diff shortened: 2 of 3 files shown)


pennersr:
Hmm, this duplicates some logic that is already contained in `DefaultAccountAdapter.clean_username`. I would rather opt for a solution where the generate-unique-username method invokes the adapter method to see if the username is valid. That way, any custom logic you plug in there is transparently handled as well.  

Another issue with the approach taken here is that if I were to place "username2" in the blacklist, that it would still be accepted as the checking is done in the `_base` method, whereas the postfix is added in the `generate_unique_username` method.

So we need to invoke `clean_username` in the `generate_unique_username` method, while at the same time making sure the base is sound (to avoid an endless loop) by checking in the `_base` method as well but *without* checking against the database. For this, we need to change `clean_username(self, username)` into `clean_username(self, username, shallow=False)` or something like that, where database lookups, or checking against other upstreams services is not allowed when `shallow` is `True`.

sasha0:
Actually, why `_base` default username is hardcoded - https://github.com/pennersr/django-allauth/blob/master/allauth/utils.py#L48? Maybe it also does make sense to make it setting? E.g. `ACCOUNT_DEFAULT_BASE_USERNAME` or etc.

sasha0:
@pennersr anyway, updated PR, please review.

allauth/account/adapter.py
```diff
@@ -229,14 +229,16 @@ def clean_username(self, username):
...
+        if not shallow:
+            username_field = app_settings.USER_MODEL_USERNAME_FIELD
+            assert username_field
+            user_model = get_user_model()
+            try:
+                query = {username_field + '__iexact': username}
+                user_model.objects.get(**query)
+            except user_model.DoesNotExist:
+                return username
         raise forms.ValidationError(_("This username is already taken. Please "
```
pennersr: If `shallow` is `True` we always raise a `ValidationError` here, that should not be the case.

allauth/utils.py
```diff
@@ -70,12 +78,12 @@ def generate_unique_username(txts, regex=None):
                 pfx = str(i + 1)
             else:
                 pfx = ''
-            ret = username[0:max_length - len(pfx)] + pfx
-            query = {USER_MODEL_USERNAME_FIELD + '__iexact': ret}
-            User.objects.get(**query)
+            ret = base_username[0:max_length - len(pfx)] + pfx
+            username = adapter.clean_username(ret)
+            if username:
```
pennersr: This check should not be necessary... if clean does not raise a `ValidationError` it must return a valid username...

pennersr:
With respect to having the `_base` username hardcoded, I would rather make this whole logic to the adapter. Probably the signature of the methods needs to be changed as well, all in all this is out of scope for this pull request I think ...

sasha0:
> With respect to having the _base username hardcoded, I would rather make this whole logic to the adapter.

You mean both unique username generation methods?

> Probably the signature of the methods needs to be changed as well, all in all this is out of scope for this pull request I think ...

I could continue with this in separate PR. What would you suggest regarding separation?

allauth/account/adapter.py
sasha0: Fixed.

allauth/utils.py
sasha0: Fixed.

pennersr:
Well, I was thinking along the lines of moving the `_generate_unique_username_base` to the adapter as well. Currently, we already have `generate_unique_username` in the adapter, but I would expect that most people just want to be able to influence the base name generation, and leave the part that makes it unique (by adding a numeric postfix) as is. 

So how about adding:

    def generate_unique_username_base(txts, regex=None, default='user')

Then, you can change easily change the `"user"` to something else by invoking `super()` with `default='somethingelse'`... avoiding a setting.

With respect to my remark about the signature... I don't really like the `txts` approach.. if you override the methods in the adapter you have no clue which entry in `txts` is the email. But, changing this breaks backwards compatibility so I would leave that as is for now.. 

If you are willing to pick this up in a separate PR... great!

allauth/utils.py
```diff
@@ -43,6 +45,11 @@ def _generate_unique_username_base(txts, regex=None):
         username = username.split('@')[0]
         username = username.strip()
         username = re.sub('\s+', '_', username)
+        # Finally, validating base username without database lookups etc.
+        try:
+            username = adapter.clean_username(username, shallow=True)
+        except ValidationError:
+            pass
         if username:
             break
```
pennersr: This is not sound... `username` is always se.. so we always break. So the break should be done above the `except` just after invoking `clean_username`

sasha0: Indeed. Fixed.

allauth/utils.py
```diff
@@ -43,8 +45,13 @@ def _generate_unique_username_base(txts, regex=None):
         username = username.split('@')[0]
         username = username.strip()
         username = re.sub('\s+', '_', username)
-        if username:
-            break
+        # Finally, validating base username without database lookups etc.
+        try:
+            username = adapter.clean_username(username, shallow=True)
+            if username:
```
pennersr: The `if` is not necessary, as `clean_username` always returns a cleaned username, and in case of error, raises an exception.

sasha0: Yep. Done.

PR #1171, opened 2015-11-02
Add Shopify as a provider

anka-sirota:
I've been using this for a while as an app on top of allauth, it would have been great to have it as part of allauth though, maybe someone else will find it useful

### DIFF
--- a/allauth/socialaccount/providers/shopify/urls.py
+++ b/allauth/socialaccount/providers/shopify/urls.py
@@ -1,0 +1,5 @@
+from allauth.socialaccount.providers.oauth2.urls import default_urlpatterns
+from .provider import ShopifyProvider
+
+urlpatterns = default_urlpatterns(ShopifyProvider)
+print(urlpatterns)
--- a/allauth/socialaccount/providers/shopify/urls.py
+++ b/allauth/socialaccount/providers/shopify/urls.py
@@ -2,4 +2,3 @@ from allauth.socialaccount.providers.oauth2.urls import default_urlpatterns
 from .provider import ShopifyProvider
 
 urlpatterns = default_urlpatterns(ShopifyProvider)
-print(urlpatterns)
--- a/allauth/socialaccount/providers/shopify/views.py
+++ b/allauth/socialaccount/providers/shopify/views.py
@@ -1,0 +1,58 @@
+from allauth.socialaccount.providers.oauth2.client import OAuth2Client
+from allauth.socialaccount.providers.oauth2.views import (OAuth2Adapter,
+                                                          OAuth2LoginView,
+                                                          OAuth2CallbackView)
+import requests
+from .provider import ShopifyProvider
+
+
+class ShopifyOAuth2Client(OAuth2Client):
+    def get_redirect_url(self, request, extra_params):
+        self.shop = extra_params['shop']
+        _url = 'https://{}.{}'.format(
+            self.shop,
+            super(ShopifyOAuth2Client, self).get_redirect_url(request, extra_params))
+        return _url
+
+
+class ShopifyOAuth2LoginView(OAuth2LoginView):
+    def get_client(self, request, app):
+        client = super(ShopifyOAuth2LoginView, self).get_client(request, app)
+        client.__class__ = ShopifyOAuth2Client
+        return client
+
+
+class ShopifyOAuth2CallbackView(OAuth2CallbackView):
+    def get_client(self, request, app):
+        client = super(ShopifyOAuth2CallbackView, self).get_client(request, app)
+        client.access_token_url = 'https://{}/{}'.format(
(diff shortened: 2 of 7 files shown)


allauth/socialaccount/providers/shopify/urls.py
```diff
@@ -0,0 +1,5 @@
+from allauth.socialaccount.providers.oauth2.urls import default_urlpatterns
+from .provider import ShopifyProvider
+
+urlpatterns = default_urlpatterns(ShopifyProvider)
+print(urlpatterns)
```
pennersr: You'll probably want to remove this print :)

allauth/socialaccount/providers/shopify/views.py
```diff
@@ -0,0 +1,58 @@
...
+    def profile_url(self):
+        return 'admin/shop.json'
+
+    def complete_login(self, request, app, token, **kwargs):
+        headers = {'X-Shopify-Access-Token': '{token}'.format(token=token.token)}
+        response = requests.get('https://{}/{}'.format(request.GET['shop'], self.profile_url),
+                                headers=headers)
+        extra_data = response.json()
+        # save access token with extra data, to be able to use it in social_account_removed
+        extra_data.update({'token': token.token})
```
pennersr: Wondering why this is necessary. The access token is available in SocialToken, so you can simply look that up in your `social_account_removed` signal. Personally, I would like to keep the extra_data clean ...

allauth/socialaccount/providers/shopify/views.py
```diff
@@ -0,0 +1,58 @@
...
+        client = super(ShopifyOAuth2LoginView, self).get_client(request, app)
+        client.__class__ = ShopifyOAuth2Client
+        return client
+
+
+class ShopifyOAuth2CallbackView(OAuth2CallbackView):
+    def get_client(self, request, app):
+        client = super(ShopifyOAuth2CallbackView, self).get_client(request, app)
+        client.access_token_url = 'https://{}/{}'.format(
+            request.GET.get('shop'), client.access_token_url)
```
pennersr: We'll probably need to use form validation to check if `shop` is a proper domain. Or, better yet, use <input type=url> and have users fill in the URL to their shop...

allauth/socialaccount/providers/shopify/views.py
anka-sirota: afair, I had problems recovering the token in social_account_removed because respective SocialToken is already deleted by the time signal is being fired

gonna recheck that, it doesn't sound plausible now

allauth/socialaccount/providers/shopify/views.py
pennersr: Wondering: I am not fully familiar with shopify, but are only subdomains allowed here? How about shop owners that have their own fqdn ?

pennersr:
Thanks for the pull... it will definititely come in useful for some other ongoing projects.

allauth/socialaccount/providers/shopify/views.py
anka-sirota: the idea was to have users fill in only the name of their shop (subdomain of .myshopify.com), because this is a "familiar" way for Shopify users, making them to fill in the whole URL might be a bit too much

connect_shopify

allauth/socialaccount/providers/shopify/urls.py
anka-sirota: oh boy. also, there's a urlparse issue there, gonna fix that too

allauth/socialaccount/providers/shopify/views.py
anka-sirota: > Wondering: I am not fully familiar with shopify, but are only subdomains allowed here? How about shop owners that have their own fqdn ?

they all retain their default myshopify subdomains: 

```
{"shop": {
  "domain": "mycooldomain.com", 
...
  "myshopify_domain": "my-cool-shop.myshopify.com"
}}
```
authentication flow via {name}.myshopify.com always works, even for shops with custom domains

agriffis:
What remains to be done here before this can be merged?

pennersr:
This PR adds a few templates that don't seem to be used (login.html, base.html). Also, the views.py has some hacks to deal with the fact that shopify requires a dynamic oauth2 domain, depending on the shop parameter. I am in the process of cleaning this up.

ivanvenosdel:
> Also, the views.py has some hacks to deal with the fact that shopify requires a dynamic oauth2 domain, depending on the shop parameter. I am in the process of cleaning this up.

@pennersr I see that the hack was removed and the docs don't mention the need to implement a replacement yet I don't see how to get this to work without it. Did you have some better way to get passed the dynamic oauth2 domain issue that just hasn't yet been added or is solving it intended to be an exercise left to the reader?

PR #1173, opened 2015-11-02
Fix django 1.9 KeyError at /accounts/signup

guialante:


pennersr:
Already handled in 3eaecb9

Issue #1049, opened 2015-07-23
Visible request in adapter


nikolaz111:
The `request` object should be visible to the `send_mail` function in the adapter. Why this would be useful? I use mandrill in all my projects to handle email sending. Request objects are the only objects that can provide access to HTTP headers. I need headers in the adapter in order to provide an organized and easy way to send i18n emails. This is because its much easier to edit mandrill templates rather than editing Django templates, plus madrill provides useful analytics.

I think this would provide greater flexibility and it is a very easy to do improvement.

The function definition would be:

```
def send_mail(self, template_prefix, email, context, request):
```

Any comments appreciated.

pennersr:
Whether or not exposing `request` is a good thing is debatable, as there are many projects out there that are sending mails not in the context of a request, but for example from within a celery worker.

In your case, if it is the language you need, can't you use `django.utils.translation.get_language()`, and derive the params from that?

HEdingfield:
Something like this might be useful in my case because I send out emails that have links to my site in them. I'm currently having to do something like this to create these links:

    'https://www.%s%s' % (Site.objects.get_current().domain, reverse('my_rad_page', args=(foo)))

...but it might be nicer to grab the actual domain by using `build_absolute_uri`. Maybe request could be provided as a `kwarg` so it still runs fine without requiring a request?

pennersr:
You can use `allauth.utils.build_absolute_uri(request=None, location='/foo')` to clean up that code.

But, given that this is debatable and I do not want to be too opinionated I am ok with adding a `request=None` parameter.

HEdingfield:
I saw you expanded that helper function today, nice work! However, in this case it still means I need to manually add the "www" in, which isn't great. I think this is just an annoying limitation with the way the Django sites framework operates (e.g. I can't make my site www.example.com, or else it'll screw up the email addresses I use to send things out).

pennersr:
I don't fully understand what issue you are running into with the `www` prefix. You should be able to make your site www.example.com without any impact on your email...

Feel free to submit a PR that adds a `request=None` to the send_mail ...

HEdingfield:
My mistake, the problem I had with the sites domain is actually a result of using subdomains and dev servers, not with email addresses. E.g. sometimes I use www.example.com and other times I use dev.example.com or 127.0.0.1:8000.

In any case, I'll put it on my list to do a PR for this, unless someone (@nikolaz111?) else beats me to it.

[...]

PR #1185, opened 2015-11-08
Cleanup and update Travis and Tox file

Ecno92:
* Removed Django 1.6 from testing. It is unsupported. See (https://www.djangoproject.com/download/)
* Added Python 3.5 to the list of Python versions to test against. It is supported since Django 1.8.6 (https://docs.djangoproject.com/en/1.8/releases/1.8.6/)
* Removed `--use-mirrors` from pip install command in the Travis config since it is deprecated. (https://github.com/pypa/pip/pull/1098) 

Current Python 3.5 build fails (https://travis-ci.org/Ecno92/django-allauth/jobs/89953488). It seems that one of the providers (openid) is not Python 3.5 compatible. Since this package has a dependency on it I would suggest to keep it failing until `python-openid` resolves the issue. 

Please let me know if it is necessary to open up a issue at the `python-openid` project. I didn't investigate this failing test that well to be sure if this is necessary.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py{27,32,33,34,35}-django{17,18},
+envlist = py{27,32,33,34,35}-django{16,17,18},
 
 [testenv]
 deps =
--- a/tox.ini
+++ b/tox.ini
@@ -1,11 +1,10 @@
 [tox]
-envlist = py{27,32,33,34}-django{16,17,18},
+envlist = py{27,32,33,34,35}-django{17,18},
 
 [testenv]
 deps =
     coverage
     mock >= 1.0.1
-    django16: Django < 1.7
     django17: Django < 1.8
     django18: Django < 1.9
 commands =
--- a/tox.ini
+++ b/tox.ini
@@ -5,6 +5,7 @@ envlist = py{27,32,33,34,35}-django{16,17,18},
 deps =
     coverage
     mock >= 1.0.1
+    django16: Django < 1.7
     django17: Django < 1.8
     django18: Django < 1.9
 commands =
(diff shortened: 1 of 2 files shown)


pennersr:
See #1178 for a discussion on dropping Django 1.6 -- TL;DR: I am not at liberty to drop support just yet. I am not very fond of intentionally having a failing build, so we either need to add 3.5 to `allow_failures`, or, detect python 3.5 in just the OpenID test suite and print a big warning or something like that ... is there an open issue for the python-openid project we can refer to?

pennersr:
Btw, see #1094 for --mirrors.... (it has merge conflicts though)

Ecno92:
@pennersr All tests seem to pass now. Python 3.5 fails are ignored. Django 1.6 is tested again and all the optimizations of @moreati are integrated. 

During development I found out that I needed to explicitly set the version of coverage back to an older version in order to make it run successfully under Python 3.2. At the time #1049 was submitted not setting an explicit version worked fine, but I found out that the new version of coverage is not compatible anymore. 

Since the dropping of certain dependency and Python versions is sometimes under discussion in this project it may be a good idea to make a "roadmap" when we are going to drop support for a certain package or dependency. This may prevent some discussions and allows us to drop support for old versions quickly at a certain point in time.

tox.ini
```diff
@@ -1,11 +1,10 @@
 [tox]
-envlist = py{27,32,33,34}-django{16,17,18},
+envlist = py{27,32,33,34,35}-django{16,17,18},
 
 [testenv]
 deps =
     coverage
     mock >= 1.0.1
-    django16: Django < 1.7
```
pennersr: Django 1.6 is still gone?

PR #1195, opened 2015-11-16
Limit number of failed login attempts

### LINKED ISSUE #1007: Limit number of failed login attempts
Not sure if it's in the scope of the project, but I'd love to see something that allows you to limit the number of failed login attempts (and, say, locks an account for x minutes, hours, etc. once this limit is reached). This would prevent someone / some script from just spamming a bunch of attempts at guessing a user's password.

sasha0:
Fixes #1007

### DIFF
--- a/allauth/account/adapter.py
+++ b/allauth/account/adapter.py
@@ -1,30 +1,33 @@
 from __future__ import unicode_literals
 
-import re
-import warnings
 import json
+import re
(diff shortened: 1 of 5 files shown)


sasha0:
Basically, it's initial implementation. @pennersr please review and let me know if I'm moving in right direction. Thanks.

PR #1196, opened 2015-11-17
Add management command to cleanup expired email confirmations

### LINKED ISSUE #252: Cleanup expired email confirmation
Is there a way to clean up expired email confirmations?

Say we have two users Alice (alice@example.com) and Bob (bob@example.com). Now Alice creates an account and uses alice@example.com for her primary addresses. But she also adds bob@example.com as an additional email without confirming it (because she does not have Bob's password for the mail account).

Now Bob wants to signup using bob@example.com which is not possible.

What about a management command which will delete non-verified emails older than `timezone.now() - timedelta(days=app_settings.EMAIL_CONFIRMATION_EXPIRE_DAYS)`?

sasha0:
Fixes #252

### DIFF
--- a/allauth/account/management/commands/account_deleteexpiredconfirmations.py
+++ b/allauth/account/management/commands/account_deleteexpiredconfirmations.py
@@ -1,0 +1,9 @@
+from django.core.management.base import BaseCommand
+from allauth.account.models import EmailConfirmation
+
+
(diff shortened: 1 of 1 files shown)


pennersr:
I am planning on dropping the EmailConfirmation model, so this command would be rendered irrelevant soon.

PR #1199, opened 2015-11-18
Add signup email verification, moved password verification to mixin to avoid code duplication

### LINKED ISSUE #788: New configuration setting ACCOUNT_SIGNUP_EMAIL_VERIFICATION
Add a configuration settings `ACCOUNT_SIGNUP_EMAIL_VERIFICATION` so when users are signing up, it will let the user type in their email twice to avoid typos.

Similar to preexisting setting `ACCOUNT_SIGNUP_PASSWORD_VERIFICATION`  - When signing up, let the user type in their password twice to avoid typos.

**Use case**:
Users mistype their emails when signing up, sometimes as much as their password. Having them type in their email twice avoids adding bad emails to your system.

sasha0:
Fixes #788

### DIFF
--- a/allauth/account/forms.py
+++ b/allauth/account/forms.py
@@ -29,6 +29,16 @@ except ImportError:
     from django.utils.importlib import import_module
 
 
+class PasswordVerificationMixin(object):
+    def clean(self):
+        cleaned_data = super(PasswordVerificationMixin, self).clean()
+        password1 = cleaned_data.get('password1', None)
+        password2 = cleaned_data.get('password2', None)
+        if (password1 and password2) and password1 != password2:
+            self.add_error('password2', _("You must type the same password each time."))
+        return cleaned_data
+
+
 class PasswordField(forms.CharField):
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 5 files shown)


sssbox:
Hmmm, a bit unfortunate situation with the naming here.

In general (and within the allauth project) "Email Verification" and "Email Confirmation" both refer to sending an email and having the user click on a confirmation link, however the existing repeat password setting is called `SIGNUP_PASSWORD_VERIFICATION`.

Does anyone like either of these alternatives for the new name: `SIGNUP_EMAIL_ASK_TWICE`, `SIGNUP_EMAIL_ENTER_TWICE`?

allauth/account/forms.py
```diff
@@ -223,17 +233,31 @@ def __init__(self, *args, **kwargs):
         super(BaseSignupForm, self).__init__(*args, **kwargs)
         # field order may contain additional fields from our base class,
         # so take proper care when reordering...
+        self.email_field_name = 'email'
         field_order = ['email', 'username']
+        if app_settings.SIGNUP_EMAIL_VERIFICATION:
+            del self.fields["email"]
```
sssbox: Is there a reason to swap out `email` fields?

If someone wanted to switch the SIGNUP_EMAIL_VERIFICATION from True to False or False to True, keeping the first field name consistent would result in a better user experience for anyone who had the form open during the setting swap.

(True to False: the second field would just be ignored, False to True: the form would redisplay with the email as entered, but would now also ask for the email to be reentered).

sasha0: > Is there a reason to swap out email fields?

> If someone wanted to switch the SIGNUP_EMAIL_VERIFICATION from True to False or False to True, keeping the first field name consistent would result in a better user experience for anyone who had the form open during the setting swap.

It was made for consistency with password fields - instead of having `email` and `email_confirmation` fields it was implemented fields `email1` and `email2`. Having different approaches in email and password verification might cause confusion. 

Original `email` was kept as is by default - for backward compatibility, since email verification is not widely used and would be optional.

allauth/account/forms.py
```diff
@@ -30,6 +30,16 @@
     from django.utils.importlib import import_module
 
 
+class PasswordVerificationMixin(object):
+    def clean(self):
+        cleaned_data = super(PasswordVerificationMixin, self).clean()
+        password1 = cleaned_data.get('password1', None)
+        password2 = cleaned_data.get('password2', None)
+        if (password1 and password2) and password1 != password2:
+            raise forms.ValidationError(_("You must type the same password each time."))
```
pennersr: 👍  for making the code more DRY and putting this in a mixin. These changes do change the location of where the error is displayed (from password2 field level to form level). We should use `add_error` here and add this error to the password2 field.

allauth/account/forms.py
```diff
@@ -253,13 +277,22 @@ def clean_username(self):
...
                 self.raise_duplicate_email_error()
         return value
 
+    def clean(self):
+        cleaned_data = super(BaseSignupForm, self).clean()
+        if app_settings.SIGNUP_EMAIL_VERIFICATION:
+            email1 = cleaned_data.get('email1', None)
+            email2 = cleaned_data.get('email2', None)
+            if (email1 and email2) and email1 != email2:
+                raise forms.ValidationError(_("You must type the same email each time."))
```
pennersr: Same here, let's use `add_error` and add the error to `email2`

docs/configuration.rst
```diff
@@ -76,6 +76,9 @@ ACCOUNT_SIGNUP_FORM_CLASS (=None)
 ACCOUNT_SIGNUP_PASSWORD_VERIFICATION (=True)
   When signing up, let the user type in their password twice to avoid typo's.
 
+ACCOUNT_SIGNUP_EMAIL_VERIFICATION (=False)
+  Same as above, but for an email.
```
pennersr: I need to sort these settings alphabetically some time, to avoid issues: "When signing up, let the user type in their email twice to avoid typo's."

sasha0: Done, updated description. Sorted settings in the separate commit.

allauth/account/forms.py
sasha0: Done.

allauth/account/forms.py
sasha0: Done.

sasha0:
@pennersr I updated PR based on the feedback and recently rebased it against master branch. Please take a look.

Thanks.

pennersr:
Thanks

pennersr:
@sssbox Your ENTER_TWICE suggestion is a good one, see: e35c032

PR #1202, opened 2015-11-20
Override provider display name based on settings.

### LINKED ISSUE #466: Providers should allow their display name to be overridden from the settings.
IE: 

```
SOCIALACCOUNT_PROVIDERS = {
    "persona" : {
       "display-name": "Mozilla Persona",
       "alt-text": "Using only your email address, login with Mozilla Persona"
    }
}
```

This would override the name of a provider that is used in template where the list is built in a loop.
https://github.com/pennersr/django-allauth/blob/master/allauth/templates/socialaccount/snippets/provider_list.html#L16

sasha0:
Fixes #466

### DIFF
--- a/allauth/socialaccount/providers/base.py
+++ b/allauth/socialaccount/providers/base.py
@@ -24,6 +24,13 @@ class AuthError(object):
 
 
 class Provider(object):
+    name = None
+
+    def __init__(self):
+        settings = self.get_settings()
+        self.alt_text = settings.get('ALT_TEXT', self.name)
+        self.display_name = settings.get('DISPLAY_NAME', self.name)
+
     def get_login_url(self, request, next=None, **kwargs):
         """
         Builds the URL to redirect to when initiating a login for this
--- a/docs/providers.rst
+++ b/docs/providers.rst
@@ -25,6 +25,18 @@ For local development, use the following::
 
 
 
+Each provider could customize `display_name` and `alt_text` properties by respective settings::
+
+    SOCIALACCOUNT_PROVIDERS = {
+        "github": {
+            "DISPLAY_NAME": "Login via Github",
+            "ALT_TEXT": "Connect your github account."
+        }
+    }
+
+
+For instance, this could be used to customize providers list, returned from registry in template `socialaccount/snippets/provider_list.html`.
+
 Amazon
 ------
 
--- a/allauth/templates/socialaccount/snippets/provider_list.html
+++ b/allauth/templates/socialaccount/snippets/provider_list.html
@@ -6,16 +6,16 @@
 {% if provider.id == "openid" %}
 {% for brand in provider.get_brands %}
 <li>
-  <a title="{{brand.name}}" 
-     class="socialaccount_provider {{provider.id}} {{brand.id}}" 
+  <a title="{{ brand.name }}"
+     class="socialaccount_provider {{ provider.id }} {{ brand.id }}"
      href="{% provider_login_url provider.id openid=brand.openid_url process=process %}"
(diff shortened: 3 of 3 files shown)


docs/providers.rst
```diff
@@ -25,6 +25,18 @@ For local development, use the following::
 
 
 
+Each provider could customize `display_name` and `alt_text` properties by respective settings::
+
+    SOCIALACCOUNT_PROVIDERS = {
+        "github": {
+            "display_name": "Login via Github",
+            "alt_text": "Connect your github account."
```
sssbox: For consistency with the other per provider settings, you should probably capitalize `DISPLAY_NAME` and `ALT_TEXT`.

allauth/socialaccount/providers/base.py
```diff
@@ -24,6 +24,17 @@ class AuthError(object):
...
 
 class Provider(object):
+    alt_text = None
+    display_name = None
+    name = None
+
+    def __init__(self):
+        super(Provider, self).__init__()
+        settings = self.get_settings()
+        print settings
```
sssbox: Extra print statement here.

allauth/socialaccount/providers/base.py
```diff
@@ -24,6 +24,17 @@ class AuthError(object):
 
 
 class Provider(object):
+    alt_text = None
+    display_name = None
+    name = None
```
sssbox: I don't really think the above 3 attributes should be defined here.

The first two will always be set with `__init__`.

Not having `name` will give a helpful exception if someone adds a new provider without the `name` attribute specified. An improvement on both might be leaving `name = None` but then raising an exception if it isn't specified by the new Provider.

allauth/socialaccount/providers/base.py
```diff
@@ -24,6 +24,17 @@ class AuthError(object):
...
+
+    def __init__(self):
+        super(Provider, self).__init__()
```
sssbox: Since `Provider` is the lowest level Provider class, this `super().__init__()` call is not necessary.

allauth/socialaccount/providers/base.py
sasha0: @sssbox You're suggesting to validate `Provider.name` this way?

```
def __init__(self):
    if not self.name:
        raise ValueError('Provider class must have name attribute.')
    settings = self.get_settings()
    self.alt_text = settings.get('ALT_TEXT', self.name
    self.display_name = settings.get('DISPLAY_NAME', self.name)
```

sssbox: I'm not sure if there is a standard way to validate required attributes, but this seems reasonable.

`AttributeError` may be slightly more appropriate, but I'm not sure.

Part of me isn't a fan of extra code to make this kind of thing a little bit more explicit (since an exception would happen even without the extra check and would only be encountered by people either doing very custom things, or helping to add a provider to django-allauth itself), but it's quite possible that I'd be in the minority of Python/Django programmers in that regard. @pennersr may have an opinion for how things should work in django-allauth.

My main concern with the original pull request is that it would result in visual oddities but no explicit exceptions if someone left off the `name` attribute in a new provider (though actually, looking at the existing code, I'm not sure anything actually called `self.name` from Python, so the template would have probably seen a `None` anyways).

sasha0:
@sssbox thanks for the feedback - updated PR already.

pennersr:
I prefer not to pick this up. The templates are merely meant as a starting point, I recommend you override them yourself and tweak its contents via that way. I like to stay away from moving markup related information into the settings.

PR #1204, opened 2015-11-21
Rename context variable for consistency in email templates

### LINKED ISSUE #1180: Inconsistency: "site" vs "current_site" in email template context
Writing an email template is error prone because of this.

sasha0:
Fixes #1180

### DIFF
--- a/allauth/account/forms.py
+++ b/allauth/account/forms.py
@@ -417,10 +417,15 @@ class ResetPasswordForm(forms.Form):
         return self.cleaned_data["email"]
 
     def save(self, request, **kwargs):
-
+        current_site = get_current_site(request)
         email = self.cleaned_data["email"]
         token_generator = kwargs.get("token_generator",
                                      default_token_generator)
+
+        def deprecated_site():
+            warnings.warn("Context variable `site` deprecated, use"
+                          "`current_site` instead", DeprecationWarning)
+            return current_site
 
         for user in self.users:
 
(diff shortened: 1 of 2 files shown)


sssbox:
I'm not sure how much allauth promises to be backwards compatible, but it may make sense to add site to the context under both names.

If you make a new function that returns the site and pass the (unexecuted) function rather than the value to the old key you can add a deprecation warning to the function that'll trigger if someone's custom template uses the old variable.

sasha0:
@sssbox yeah agree, will update PR. Thanks for the tip.

sssbox:
Hey @sasha0 I was thinking something more like:

```python
+            def deprecated_site():
+                warnings.warn("Context variable `site` deprecated, use"
+                              "`current_site` instead", DeprecationWarning)
+                return current_site
-            context = {"site": current_site,
+            context = {"site": deprecated_site,
+                       "current_site": current_site,
                        "user": user,
                        "password_reset_url": url,
                        "request": request}
+
+            warnings.warn("Context variable `site` deprecated, use"
+                          "`current_site` instead", DeprecationWarning)
```

Though I'm still not sure if all of this is worth it. It may be allauth's policy to simply make it clear in the upgrade notes that the user needs to check any custom templates they've made.

sasha0:
@sssbox I think it's a good idea. I'll use this approach, thanks.

sasha0:
Also moved `get_current_site` outside of the cycle to avoid unneeded query duplication.

PR #1209, opened 2015-11-26
Avoid overwrite email on signup

caruccio:
When using a custom User model with `email `only (no `username` field) and `ACCOUNT_USER_MODEL_USERNAME_FIELD` is set to 'email', default signup adapter will overwrite `User.email` ith an auto generated `username`.

### DIFF
--- a/allauth/account/adapter.py
+++ b/allauth/account/adapter.py
@@ -198,7 +198,8 @@ class DefaultAccountAdapter(object):
         email = data.get('email')
         username = data.get('username')
         user_email(user, email)
-        user_username(user, username)
+        if username:
+            user_username(user, username)
         if first_name:
             user_field(user, 'first_name', first_name)
         if last_name:


caruccio:
..and I have no idea how to setup tests

pennersr:
Please check the "Custom user models" documentation. If you have no username field, set `ACCOUNT_USER_MODEL_USERNAME_FIELD` to `None`.

caruccio:
In fact, my username field is the email. If I set it to None, how could one login?

pennersr:
Conceptually/semantically, you do not have a username field. Only an email field. Therefore, you must set it tot None. You login by email right?

caruccio:
That makes complete sense now! I was confusing username/email roles.

Thanks for your amazing support. If you ever come to Brazil, please feel free to drop me a line so we can have a beer ;)

PR #1214, opened 2015-12-04
Replace create_oauth2_tests with OAuth2TestsMixin

agriffis:
I tried to use `create_oauth2_tests` but hit the error:

```
  File "allauth/socialaccount/tests.py", line 226, in create_oauth2_tests
      Class = type(class_name, (TestCase,), impl)
TypeError: type() argument 1 must be string, not unicode
```

This happened because I'm using `unicode_literals` in my custom `provider.py`

I started to fix this by casting to `str` in `create_oauth2_tests` but then
I wondered why it's doing dynamic class creation in the first place, rather
than simply being a mixin. So here's `OAuthTestsMixin` and `OAuth2TestsMixin`
to replace `create_oauth_tests` and `create_oauth2_tests` respectively.

The function forms are still available for the sake of third-party providers
that call them.

### DIFF
--- a/allauth/socialaccount/providers/amazon/tests.py
+++ b/allauth/socialaccount/providers/amazon/tests.py
@@ -1,11 +1,12 @@
-from allauth.socialaccount.tests import create_oauth2_tests
-from allauth.tests import MockedResponse
-from allauth.socialaccount.providers import registry
+from allauth.socialaccount.tests import OAuth2TestsMixin
+from allauth.tests import MockedResponse, TestCase
 
 from .provider import AmazonProvider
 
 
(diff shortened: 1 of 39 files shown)


agriffis:
Hmm, I'll look at the Django1.9 problem this evening

agriffis:
Failure has nothing to do with this PR, but I filed another PR to fix it, see #1216

agriffis:
@pennersr Do you have any concerns about this, or do you think you'll merge it eventually? No rush, I'm just asking because I have some improvements to `create_oauth2_tests` or `OAuth2TestsMixin` and I'd like to know if it makes sense to build the new PR on top of this or not.

pennersr:
No concerns... changes look good, thanks..

Issue #1131, opened 2015-09-22
Add support for BitBucket Oauth 2


ericholscher:
It seems Bitbucket has added support for OAuth2:

* https://blog.bitbucket.org/2015/07/30/bitbucket-now-supports-oauth2-and-fine-grained-scopes-for-resources/
* https://developer.atlassian.com/static/bitbucket/concepts/oauth2.html

I'm happy to work on adding this support if there is a good model to base it off. Should I just implement it as a named OAuth2 provider, copying one like GitHub?

pennersr:
Indeed, I suggest you to have a look at one of the existing OAuth2 providers as a skeleton/starting point. Don't forget to provide a unit test as well (which is relatively trivial, as you only need to provide a response -- again see one of the existing providers).

I am closing this for now, feel free to submit a pull request when you are done.

pennersr:
https://codeberg.org/allauth/django-allauth/issues/1131

PR #1215, opened 2015-12-04
Add Bitbucket OAuth2 provider

agjohnson:
Basing most of this on another OAuth2 provider, this adds Bitbucket OAuth2 support. The only hang up I hit while implementing this was fetching the user email from a secondary API endpoint, so some focus there in this review would be a good thing.

Refs #1131


pennersr:
Merged, thanks

PR #1249, opened 2016-01-12
Update advanced.rst

percy3d:
Found the documentation on default email behavior rather confusing. This change clarifies that the HTML email is sent in addition to txt but that there are no default HTML templates in account/email (as per site-packages).


pennersr:
I agree that the documentation is seriously lacking here, though I am not sure that this line helps. E.g. I am not sure if people understand what you mean by shadows. Also could use some consistency in wording (HTML / html, text/txt)

pennersr:
Closed via 59a61f8

PR #1251, opened 2016-01-13
Support overriding two templates with non-Django templates.

aaugustin:
These two modules were using a version of `render_to_string` that is
deprecated since Django 1.8.

### DIFF
--- a/allauth/compat.py
+++ b/allauth/compat.py
@@ -5,6 +5,17 @@ if django.VERSION > (1, 8,):
 else:
     from django.utils.datastructures import SortedDict as OrderedDict  # noqa
 
+if django.VERSION > (1, 8,):
+    from django.template.loader import render_to_string
+else:
+    from django.template.loader import render_to_string as _render_to_string
+    from django.template import RequestContext
+
+    # Wire the Django >= 1.8 API to the Django < 1.7 implementation.
+    def render_to_string(template_name, context=None, request=None, using=None):
+        assert using is None, "Multiple template engines required Django >= 1.8"
+        return _render_to_string(template_name, context, RequestContext(request))
+
 try:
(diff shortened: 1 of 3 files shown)


aaugustin:
In addition to making it possible to provide Jinja2 versions of these two templates, this prepares support for Django 1.10.

Hopefully you no longer support Django 1.7 and earlier — which no longer receive security support since December 1st, 2015. In that case, you can simply make this change. Otherwise, you need to test the version of Django and use the right signature.

aaugustin:
I see that tests are still running -- and obviously failing -- on Django 1.6 and 1.7.

darklow:
I would really love to see jinja2 support.
What about check Django version first to know which syntax to use?
Or use try/except approach?
The changes are so simple that I doubt because of these you need to drop support for 1.6 and 1.7.
Maybe then PR would be accepted faster.

aaugustin:
Yes, there are various trivial strategies for staying compatible. I'm just waiting to hear whether the maintainer has an opinion on which one to use.

pennersr:
For now I am not willing to drop support for Django 1.6 just yet. I am involved in several projects where upgrading is not an option yet (don't ask :cry: )  ... but until now keeping backwards compatibility was trivial, so that should not stop us from moving forward feature wise. How about adding a `render_to_string` method to `allauth.compat` that just works ?

aaugustin:
Yes, it's easy to write a compatibility wrapper, I'll do that.

aaugustin:
Here's a new version of this PR with a compatibility method in `allauth.compat`.

It can simply be removed, with no other changes, when you stop caring about Django < 1.8.

aaugustin:
Thanks!

(Hopefully CI was green? You merged before I could check!)

PR #1256, opened 2016-01-19
Add Arabic translation

Flimm:
Most of the strings were translated. Not the best quality translation but better than nothing.


Flimm:
The translation .po file was initially copied from the Hebrew file.

While testing this translation, I realised that the Hebrew .po file was out-of-date from the code, and that it did not have the newer msgid strings. So I filed this issue: https://github.com/pennersr/django-allauth/issues/1259

This means this pull request is already out-of-date. I'm waiting on the other issue to be resolved before fixing the translations.

PR #1260, opened 2016-01-19
AJAX-enable ConnectionsView.

### LINKED ISSUE #1253: Should ConnectionsView be AJAX-enabled?
It's the only `FormView` that doesn't include `AjaxCapableProcessFormViewMixin` in its base classes.

Is this an oversight? Or is there a reason why the mixin wouldn't work there?

aaugustin:
This brings it on parity with all other form views in the project.

Fix #1253.


aaugustin:
I tested it in one of my project. It works just fine.

PR #1265, opened 2016-01-25
Don't log user out when setting the password.

### LINKED ISSUE #1262: Setting a user's password always logs them out
I believe "set password" should honor `ACCOUNT_LOGOUT_ON_PASSWORD_CHANGE`, that is:

```diff
diff --git a/allauth/account/views.py b/allauth/account/views.py
index f4cc684..f06b8a4 100644
--- a/allauth/account/views.py
+++ b/allauth/account/views.py
@@ -525,6 +525,9 @@ class PasswordSetView(AjaxCapableProcessFormViewMixin, FormView):
 
     def form_valid(self, form):
         form.save()
+        if (update_session_auth_hash is not None and
+                not app_settings.LOGOUT_ON_PASSWORD_CHANGE):
+            update_session_auth_hash(self.request, form.user)
         get_adapter().add_message(self.request,
                                   messages.SUCCESS,
                                   'account/messages/password_set.txt')
```

Alternatively you could provide a `ACCOUNT_LOGOUT_ON_PASSWORD_SET` setting that would default to the value of `ACCOUNT_LOGOUT_ON_PASSWORD_CHANGE`. But I fail to imagine a reason why one would want them to have different values.

aaugustin:
This makes "setting the password" and "changing the password" behave
identically with regards to settings.ACCOUNT_LOGOUT_ON_PASSWORD_CHANGE.

Fix #1262.

### DIFF
--- a/allauth/account/utils.py
+++ b/allauth/account/utils.py
@@ -18,6 +18,11 @@ from django.core.exceptions import ValidationError
 from allauth.compat import OrderedDict
 
 try:
+    from django.contrib.auth import update_session_auth_hash
+except ImportError:
+    update_session_auth_hash = None
+
+try:
     from django.utils.encoding import force_text
 except ImportError:
     from django.utils.encoding import force_unicode as force_text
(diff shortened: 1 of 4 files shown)


pennersr:
Thanks!

PR #1268, opened 2016-01-27
Add ACCOUNT_TEMPLATE_EXTENSION option and view mixin for allowing other template engine extensions

erm:
I've recently begun using the Jade template language with my Django projects using pyjade. I also use django-allauth in all my projects. The problem I was facing was that the allauth account views use the `.html` extension, but I need to use the `.jade` extension. My first workaround was to import the allauth views, set the `template_name` attribute to my desired template, then call the as_view() method and add my views to my urls.py. However, this solution is proving unwieldy when moving it across many different projects.

I have since edited the views.py of django-allauth to include a `DynamicTemplateExtensionMixin` class and an `ACCOUNT_TEMPLATE_EXTENSION` configuration setting. This way I can easily override the allauth account templates as I did to prior to using Jade and merely have to change the template extensions and set a single configuration option.

### DIFF
--- a/allauth/account/app_settings.py
+++ b/allauth/account/app_settings.py
@@ -238,6 +238,13 @@ class AppSettings(object):
         return self._setting('SESSION_REMEMBER', None)
 
     @property
+    def TEMPLATE_EXTENSION(self):
+        """
+        A string defining the template extension to use, defaults to `html`.
+        """
+        return self._setting('TEMPLATE_EXTENSION', 'html')
+
+    @property
     def FORMS(self):
         return self._setting('FORMS', {})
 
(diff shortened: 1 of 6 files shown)


pennersr:
Thanks, but this would need some more work:
- [ ] Feature would need documentation
- [ ] allauth test suite would break when ran within your project, due to tests checking for the template being used
- [ ] There are other places where templates are used: mail templates, but also ad hoc render() calls. These do not respect this setting yet..
- [ ] socialaccount views are to be changed as well.

erm:
I've gone ahead and removed the `DynamicTemplateExtensionMixin` and instead just check the settings when defining the template_name in each class (this way I don't have to do anything extra for the ad hoc render calls). I've also edited the tests for both account and socialaccount and everything seems to work fine now.

PR #1287, opened 2016-02-10
In my work environments the server has no internet access, so is necessary to access external urls with proxy.

marinampires:
In my work environments the server has no internet access, so is necessary to access external urls with proxy.

To use proxy in those requests, just included the property SOCIALACCOUNT_PROXY_URL on settings with the proxy url and port. 
Ex: SOCIALACCOUNT_PROXY_URL = "proxy.test.com:3128"

### DIFF
--- a/allauth/socialaccount/app_settings.py
+++ b/allauth/socialaccount/app_settings.py
@@ -71,6 +71,10 @@ class AppSettings(object):
     def UID_MAX_LENGTH(self):
         return 191
 
+    @property
+    def PROXY_URL(self):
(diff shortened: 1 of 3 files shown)


pennersr:
I have not looked into this, but I would expect that there was a more general approach available to this problem. Isn't python requests able to pick up environment variables like these?:

   export http_proxy=http://10.203.0.1:5187/
   export https_proxy=$http_proxy

pennersr:
Closing ... see above comment

PR #1299, opened 2016-02-18
Expand Firefox Accounts provider configuration

jwhitlock:
Add ``SOCIALACCOUNT_PROVIDERS`` configuration items so that [non-production Firefox Accounts deployments](https://developer.mozilla.org/en-US/docs/Mozilla/Tech/Firefox_Accounts/Introduction#Firefox_Accounts_deployments) can easily be used. This formalizes the extended ``fxa`` provider used on the Mozilla projects [browsercompat](https://github.com/mdn/browsercompat/tree/master/bcauth/socialaccount/providers/fxa) and [push-dev-dashboard](https://github.com/mozilla-services/push-dev-dashboard/tree/master/dashboard/socialaccount/providers/fxa).

### DIFF
--- a/allauth/socialaccount/providers/fxa/provider.py
+++ b/allauth/socialaccount/providers/fxa/provider.py
@@ -1,6 +1,17 @@
+from django.conf import settings
+
 from allauth.socialaccount import providers
 from allauth.socialaccount.providers.base import ProviderAccount
 from allauth.socialaccount.providers.oauth2.provider import OAuth2Provider
+
+_FXA_SETTINGS = getattr(
+    settings, 'SOCIALACCOUNT_PROVIDERS', {}).get('fxa', {})
+FXA_OAUTH_ENDPOINT = _FXA_SETTINGS.get(
+    'OAUTH_ENDPOINT',
+    'https://oauth.accounts.firefox.com/v1')
+FXA_PROFILE_ENDPOINT = _FXA_SETTINGS.get(
+    'PROFILE_ENDPOINT',
+    'https://profile.accounts.firefox.com/v1')
 
 
 class FirefoxAccountsAccount(ProviderAccount):
--- a/docs/providers.rst
+++ b/docs/providers.rst
@@ -235,6 +235,29 @@ The provider is OAuth2 based. More info:
 
 Note: This is not the same as the Mozilla Persona provider below.
 
+The following Firefox Accounts settings are available::
+
+    SOCIALACCOUNT_PROVIDERS = \
+        {'fxa':
+            'SCOPE': ['profile'],
+            'OAUTH_ENDPOINT': 'https://oauth.accounts.firefox.com/v1',
+            'PROFILE_ENDPOINT': 'https://profile.accounts.firefox.com/v1'}}
+
+
+SCOPE:
+    Requested OAuth2 scope. Default is ['profile'], which will work for
+    applications on the Mozilla trusted whitelist. If your application is not
+    on the whitelist, then define SCOPE to be ['profile:email', 'profile:uid'].
+
+OAUTH_ENDPOINT:
+    Explicitly set the OAuth2 endpoint. Default is the production endpoint
+    "https://oauth.accounts.firefox.com/v1".
+
+OAUTH_ENDPOINT:
+    Explicitly set the profile endpoint. Default is the production endpoint
+    and is "https://profile.accounts.firefox.com/v1".
+
+
 Flickr
 ------
 
--- a/docs/providers.rst
+++ b/docs/providers.rst
@@ -479,6 +479,10 @@ Development callback URL
 Persona
 -------
 
+Note: Mozilla Persona will be shut down on November 30th 2016. See
+`the announcement <https://wiki.mozilla.org/Identity/Persona_Shutdown_Guidelines_for_Reliers>`_
+for details.
+
 Mozilla Persona requires one setting, the "AUDIENCE" which needs to be the
 hardcoded hostname and port of your website. See https://developer.mozilla.org/en-US/Persona/Security_Considerations#Explicitly_specify_the_audience_parameter for more
 information why this needs to be set explicitly and can't be derived from
--- a/AUTHORS
+++ b/AUTHORS
@@ -41,6 +41,7 @@ James Thompson
 Jannis Leidel
 Jeff Triplett
 John Bazik
+John Whitlock
 Jonas Aule
 Josh Owen
 Josh Wright
--- a/allauth/socialaccount/providers/fxa/views.py
+++ b/allauth/socialaccount/providers/fxa/views.py
@@ -3,14 +3,16 @@ import requests
 from allauth.socialaccount.providers.oauth2.views import (OAuth2Adapter,
                                                           OAuth2LoginView,
                                                           OAuth2CallbackView)
-from .provider import FirefoxAccountsProvider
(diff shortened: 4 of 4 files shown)


docs/providers.rst
```diff
@@ -235,6 +235,39 @@ The provider is OAuth2 based. More info:
...
+    to authenticate against. Possible values are "production" (default),
+    "stage", "stable", and "latest".
+
+ON_WHITELIST:
+    True (default) if your application is on the Mozilla trusted whitelist,
+    False if it is not, and has limited access to profile data.
+
+SCOPE:
+    Requested OAuth2 scope. Default is ['profile'] if ON_WHITELIST, and
+    ['profile:email', 'profile:uid'] if not.
```
groovecoder: This is in the docs, but I don't see `_FXA_SETTINGS` `SCOPE` used anywhere.

allauth/socialaccount/providers/fxa/provider.py
```diff
@@ -1,7 +1,42 @@
+from django.conf import settings
+
 from allauth.socialaccount import providers
 from allauth.socialaccount.providers.base import ProviderAccount
 from allauth.socialaccount.providers.oauth2.provider import OAuth2Provider
 
+# https://developer.mozilla.org/en-US/docs/Mozilla/Tech/Firefox_Accounts/Introduction#Firefox_Accounts_deployments
+_FXA_DEPLOYMENTS = {
```
groovecoder: We should drop the pre-set, hard-coded `FXA_DEPLOYMENTS` values here and let the end user of `django-allauth` configure their `FXA_OAUTH_ENDPOINT` and `FXA_PROFILE_ENDPOINT` directly. For [push-dev-dashboard](https://github.com/mozilla-services/push-dev-dashboard), it made the [settings](https://github.com/mozilla-services/push-dev-dashboard/blob/4f1768527cd62191289c8c049cce511e553c3f46/dashboard/settings.py#L105-L112) and [view](https://github.com/mozilla-services/push-dev-dashboard/blob/4f1768527cd62191289c8c049cce511e553c3f46/dashboard/socialaccount/providers/fxa/views.py#L12-L28) code a lot smaller and cleaner...
```
SOCIALACCOUNT_PROVIDERS = {
    'fxa': {
        'OAUTH_ENDPOINT': config('FXA_OAUTH_ENDPOINT',
                                 'https://oauth-stable.dev.lcip.org/v1/'),
        'PROFILE_ENDPOINT': config('FXA_PROFILE_ENDPOINT',
                                   'https://stable.dev.lcip.org/profile/v1/profile')
   }
}
```
```
FXA_CONFIG = getattr(settings, 'SOCIALACCOUNT_PROVIDERS', {}).get('fxa', {})


class FirefoxAccountsOAuth2Adapter(BaseAdapter):
    provider_id = FirefoxAccountsProvider.id

    @property
    def access_token_url(self):
        return FXA_CONFIG.get('OAUTH_ENDPOINT', '') + '/token'

    @property
    def authorize_url(self):
        return FXA_CONFIG.get('OAUTH_ENDPOINT', '') + '/authorization'

    @property
    def profile_url(self):
        return FXA_CONFIG.get('PROFILE_ENDPOINT', '')
```

allauth/socialaccount/providers/fxa/provider.py
```diff
@@ -16,7 +51,10 @@ class FirefoxAccountsProvider(OAuth2Provider):
     account_class = FirefoxAccountsAccount
 
     def get_default_scope(self):
-        return ['profile']
+        if FXA_ON_WHITELIST:
```
groovecoder: :+1: for handling the `FXA_ON_WHITELIST` setting. That tripped me up for a while when I made my first FxA oauth client.

groovecoder:
Overall it looks good to me. I would drop the `DEPLOYMENT` and `SCOPE` settings to simplify it though.

docs/providers.rst
jwhitlock: ``SCOPE`` is part of the [base ``oauth2``](https://github.com/pennersr/django-allauth/blob/master/allauth/socialaccount/providers/oauth2/provider.py#L28-L36) provider. If the scope is provided, then it is used, otherwise the return from ``get_default_scope()`` is used. I'm following the example of the [facebook provider docs](https://github.com/pennersr/django-allauth/blob/master/docs/providers.rst#facebook), which mention ``SCOPE`` even without special handling in the provider.

allauth/socialaccount/providers/fxa/provider.py
jwhitlock: My thinking was that this configuration:

```python
SOCIALACCOUNT_PROVIDERS = {
    'fxa': {
        'DEPLOYMENT': 'stage',
    }
}
```
was simpler and less error prone than:

```python
SOCIALACCOUNT_PROVIDERS = {
    'fxa': {
        'OAUTH_ENDPOINT': 'https://oauth-stable.dev.lcip.org/v1/',
        'PROFILE_ENDPOINT': 'https://stable.dev.lcip.org/profile/v1/',
    }
}
```
Setting a ``DEPLOYMENT`` automatically picks the base URLs in the style expected by the rest of the code. I left the ``OAUTH_ENDPOINT`` and ``PROFILE_ENDPOINT`` overrides for the the case of running a local FxA server, or if they ever decide to bump that ``v1`` before we can update django-allauth (which could be extended with a ``VERSION`` parameter like the Facebook provider).

allauth/socialaccount/providers/fxa/provider.py
jwhitlock: [It tripped me up as well](https://github.com/mozilla/fxa-oauth-server/issues/143#issuecomment-184819931), prompting me to make this PR. I think this will become better documented later, when (if?) the FxA team prioritizes opening up to 3rd party reliers.

jwhitlock:
Thanks. I've added my preferences about the settings, but the design could go either way.. Maybe @jezdez can cast a tie-breaking vote?

docs/providers.rst
groovecoder: Ah that makes sense now. Nice.

jezdez:
I concur with @groovecoder. I don't think using `DEPLOYMENT` is less error prone as it introduces a chance to accidentally missing changes to `_FXA_DEPLOYMENT` as part of an future allauth update in case something changes in the way Fxa is deployed. The auth backend URL is important enough that a mistake will be caught right away so the added wrapper is unnecessary tech debt.

jwhitlock:
Force push to:
* Drop ``DEPLOYMENT`` as method for picking endpoints
* Add a note about Mozilla Persona shutdown

jwhitlock:
Next question to @groovecoder, @jezdez. The discussion was [hidden by the force push](https://github.com/pennersr/django-allauth/pull/1299#discussion_r53320817), but you need to use a difference scope depending on if you are a trusted or untrusted application. Should we have a ``ON_WHITELIST`` parameter (default True) that switches the scopes, or remove it and rely on the documentation to override the ``SCOPE`` parameter?

groovecoder:
:1: remove it and use `FXA_SCOPE` as you did with browsercompat.

jwhitlock:
Force push to remove ``ON_WHITELIST``. It makes sense - a developer needs to read the docs anyway, to find out about the setting, and can just use the existing ``SCOPE`` declaration instead. Also, removes a code branch, so coverage won't change :grin:

I'm done making changes, I think this PR is ready.

Issue #1259, opened 2016-01-19
.po files are not up-to-date with code


Flimm:
For example, the the file `allauth/locale/uk/LC_MESSAGES/django.po` contains these lines:

```po
#: templates/account/email/email_confirmation_message.txt:7
#, python-format
msgid ""
"Thank you from %(site_name)s!\n"
"%(site_domain)s"
msgstr ""
"Спасибі Вам від %(site_name)s!\n"
"%(site_domain)s"
```

This particular language seems to be up-to-date with `allauth/templates/account/email/email_confirmation_message.txt`.

However, many of the .po files for the other languages do not contain this "msgid". This is an issue for translators like me who want to fill in the empty "msgstr" without having to learn how to update the .po files.

There also does not seem to be an easy way to update the .po files. Running this command gives an error:

```
$ python manage.py makemessages
CommandError: Unable to find a locale path to store translations for file docs/conf.py
```

Flimm:
Note that this is related to https://github.com/pennersr/django-allauth/issues/1032, but it's not the same thing.

pennersr:
.po files have been updated (7966f9b54c7b08977ba5883eecf115a88a89cf3e)

pennersr:
https://codeberg.org/allauth/django-allauth/issues/1259

PR #1306, opened 2016-02-22
Update .po .mo files

Flimm:
The .po .mo files are out-of-date with the code-base, this updates them.

In the pull request, a new script has been created `update_po_mo_files.py`, which can be run to do update both .po files and .mo files in one step. Any adjustments like the `--no-wrap --no-location --no-obsolete` flags could also be put in there.


Flimm:
This fixes issue #1259.

pennersr:
.po have been updated, for now I've added a Makefile to manage po/mo

Flimm:
Thanks!

PR #1309, opened 2016-02-23
Allow SignupView subclasses access to newly created user.

frewsxcv:


Jeewes:
You could always rewrite the whole `form_valid` method, but I guess your point is to make the method itself more extendable?

frewsxcv:
> but I guess your point is to make the method itself more extendable?

Yes.

Jeewes:
In that case looks pretty solid to me.

PR #1312, opened 2016-02-25
Keep inactive user response logic in adapter

lukeburden:
I'm using allauth to power a traditional webui as well as an API. I've had to subclass the hell out of the adapters to get the exact behaviour I want, but this is what they're intended to allow. Allauth has been a fantastic help. Particularly on the API front however, I have consistently run into inflexibility in allauth due to the utils.py and helpers.py style of logic containment.

This PR is patching yet another single instance of this inflexibility, but I'd also like to underscore how problematic logic not being in the adapters has been for me when trying to subclass and sensibly customize allauth for usage in an API. That's all :)

### DIFF
--- a/allauth/account/adapter.py
+++ b/allauth/account/adapter.py
@@ -4,17 +4,18 @@ import re
 import warnings
 import json
 
-from django.core.urlresolvers import reverse
-from django.conf import settings
-from django.http import HttpResponse
-from django.template.loader import render_to_string
-from django.template import TemplateDoesNotExist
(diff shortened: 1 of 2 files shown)


lukeburden:
If you think some sensible arguments should be made available in the added adapter methods, let me know.

pennersr:
Thanks. I've added request and user params as they are likely of use when overriding things. Also, I feel the number of adapter methods will grow too large when we offer all variants (url, response)... just one should be sufficient.

PR #1318, opened 2016-03-07
Swappable Social Models and Scale changes

yarbelk:
This is a pull request that includes hooks and changes that were needed
to accomplish scaling out to millions of users while having multiples of the
same type of provider.  For example, if you need to support two apps that use
Facebook auth, you are required to have two Facebook pages, with separate 
`SocialApp`s for each.

This pull request is much larger than I would have liked, but I needed to
figure out and get the change sets into production, and they are built on
each-other.
## Changes
- Make `SocialApp` and `SocialAccount` swappable (just like `auth.User`).
- Added settings for the swappable models (and docs for them)
- App Config has been changed to reflect the new settings.
- Add indexs to support millions of users (specifically, on `SocialAccount.uid`)
  which was causing full table scans.  This can DoS your database if
  you have enough social accounts.
- Added a test app which swaps in new models, and test settings that use this.
  This let me over ride and extend the fields in the SocialAccount/SocialApp
  to support multiple SocialApps per provider.  While we discussed earlier (#1002)
  using the Sites framework as possible way of doing this, this wasn't an option
  for the project.  The benefit of making the app swappable, is that this
  becomes transparent to the other users of allauth - they can take advantage
  of it if they want to, but existing deploys _should_ have no issues.
  - In travis, I'm running both, but I haven't figured out how to run these in tox.
    I can probably do this if I switch the test runner to `pytest` and parameterize
    the settings fixture for the pytest django plugin.
- Added a hook for determining the email `from` address, this still defaults to
  the configuration like before, but lets you determine what the email address
  on a per request basis.
  - For example, if you are using multiple providers
    based on sites or other, you don't have to duplicate all the code in the
    `send_email` method to change just this field.  We needed to deteremine
    just the from address based on some buisness logic, this was a cleaner
    interface for that.
- Propagate the request object deeper into the adapters and helpers; generally
  it defaults to None; so it doesn't have to be used.  This is really helpful
  for breaking up the larger methods in the adapters and helpers and customizing
  their functionality, and was implemented to support the above point.
### Issues to be discussed
- As I've changed the `unique_together` relationship on `SocialAccount`
  to be `uid`, `app`, rather than `uid`, `provider`, this is a point
  which needs more eyes to investigate the implications.
  - Possibly change this back and make it the responsibility of the
    person who swaps in a new model to change this?  (Re: mozilla Persona)
- Should I change the type hierarchy of `SocialApp`/`SocialAccount` to
  be
  
     SocialAppABC > BaseSocialApp > SocialApp
  
  where SocialAppABC has its interface defined with `NotImplemented`,
  and `BaseSocialApp` is an implementation of this interface?  This would
  make it cleaner for someone to swap in a new model and not be surprised
  by inherited implementations
- Test more functionality of the swappable account, like multiplexing
  providers based on a header or something.
- Lack of south migrations. I will do this once the other problems
  are fixed.


pennersr:
I did not get around to do an in-depth check yet, but I did have a quick look. Notes:

- You have added an FK from SocialAccount to SocialApp, but not all providers require a `SocialApp`. E.g. OpenID, Persona.  Given that, I am not sure adding fake SocialApps for this purpose is the right thing to do. Perhaps making it NULLable would be better.

- On Facebook, the same person has a different SocialAccount uid if a different SocialApp is used. However, this is not the case for all providers. For example, for other providers the same person has the same uid, even if the app is different. Again, this makes me wonder if it is okay to add the app FK to the SocialAccount.

- I've already added your `get_from_email` to the adapter. The request parameter is not needed, as you can inspect `self.request` in adapter methods.

- Currently there is a unique together on (provider, uid) -- doesn't that sufficiently address your uid index concern if you make sure you query on both provider and uid ?

- There are developers who dislike having the app tokens in the database, and prefer not to use `django.contrib.sites`. If we would make models swappable, it would be nice if it would become possible to swap in a SocialApp model without the sites dependency, or, perhaps don't swap in a SocialApp model at all and leave the keys in the settings. I am aware that this may be out of scope from your point of view.

- South is no longer a concern -- Django 1.6 support was dropped.

yarbelk:
Thanks,  I will addresses these;
> You have added an FK from SocialAccount to SocialApp, but not all providers require a SocialApp. E.g. OpenID, Persona. Given that, I am not sure adding fake SocialApps for this purpose is the right thing to do. Perhaps making it NULLable would be better.

I will drop this; It doesn't make sense for a default,, but can be added if some swaps in a new model


> On Facebook, the same person has a different SocialAccount uid if a different SocialApp is used. However, this is not the case for all providers. For example, for other providers the same person has the same uid, even if the app is different. Again, this makes me wonder if it is okay to add the app FK to the SocialAccount.

Good point (though persona is dieing at the end of the year)

> I've already added your get_from_email to the adapter. The request parameter is not needed, as you can inspect self.request in adapter methods.

> Currently there is a unique together on (provider, uid) -- doesn't that sufficiently address your uid index concern if you make sure you query on both provider and uid ?

Unique Together doesn't work for having 2 google apps: accounts; the UID is the same between them for the same user.  I will review this to see about what I'm specifically trying to solve.

> There are developers who dislike having the app tokens in the database, and prefer not to use django.contrib.sites. If we would make models swappable, it would be nice if it would become possible to swap in a SocialApp model without the sites dependency, or, perhaps don't swap in a SocialApp model at all and leave the keys in the settings. I am aware that this may be out of scope from your point of view.

Totally cool.  I like this; will look at it.

> South is no longer a concern -- Django 1.6 support was dropped.

Yay!

yarbelk:
@pennersr OK, improved coverage, limited scope to *just* the swappable stuff, fixed the flake8 issues.

I took a quick stab at the other parts; but believe that its a larger kettle of fish than a simple addtion to this pull request.

allauth/socialaccount/models.py
```diff
@@ -97,17 +157,20 @@ class SocialAccount(models.Model):
     # [2] http://openid.net/specs/openid-authentication-1_1.html#limits
 
     uid = models.CharField(verbose_name=_('uid'),
-                           max_length=app_settings.UID_MAX_LENGTH)
+                           max_length=app_settings.UID_MAX_LENGTH,
+                           db_index=True)
```
pennersr: Adding an index is beyond the scope of this PR. Why is it needed, as there is a unique together on provider, uid ?

allauth/socialaccount/providers/openid/tests.py
```diff
@@ -2,14 +2,24 @@
...
 
 from . import views
 from .utils import AXAttribute
 
+from allauth.utils import get_user_model, get_current_site
+
 
 class OpenIDTests(TestCase):
+    def setUp(self):
+        SocialApp = get_social_app_model()
```
pennersr: OpenID should not need an app.. why make this change?

allauth/socialaccount/providers/persona/tests.py
```diff
@@ -1,14 +1,23 @@
...
+from allauth.socialaccount.models import get_social_app_model
+from allauth.utils import get_user_model, get_current_site
 
 SOCIALACCOUNT_PROVIDERS = {'persona':
                            {'AUDIENCE': 'https://www.example.com:433'}}
 
 
 class PersonaTests(TestCase):
+    def setUp(self):
+        SocialApp = get_social_app_model()
```
pennersr: Same issue as with OpenID -- no app is needed.

allauth/socialaccount/tests.py
```diff
@@ -166,9 +175,10 @@ def test_login(self):
         self.assertRedirects(resp, reverse('socialaccount_signup'))
 
     def test_account_tokens(self, multiple_login=False):
-        email = 'some@mail.com'
+        username = str(random.randrange(1000, 10000000))
```
pennersr: Could you explain why this change was made?

bittner:
This PR looks live a valuable contribution. Having also the `SocialToken` model swappable would be helpful.

@yarbelk Do you care to address the concerns @pennersr has raised, and rebase your changes?

yarbelk:
Sure.  I was caught up in non-django development, but I'm back in python land now.

dshinzie:
This looks like a fantastic feature and exactly what my colleagues and I need right now. Any idea when this could be implemented? @yarbelk

varunsaral:
Can i do this @pennersr ? Although i will require some time to fully grasp the working of the project, this seems a big change and i am still new in Django devlopement.

pennersr:
@varunsaral This is not something that you can just dive in and start doing. It really requires a more thorough understanding of what people want to accomplish, and, a proper proposal on how that could at all be implemented. Thoughts:

- For things like adding indices you don't need to swap th emodels -- we need to have properly indexes models out of the box.

- Is it really needed to swap out the whole model, or, would it be sufficient to be able to configurate what kind of PK field woul dbe used (e.g. int vs a UUID or hashid).

- Django does offer some functionality for swapping models, but currently it's limited to the user model. There is also projects like https://github.com/openwisp/django-swappable-models. Do we need/want to invent a framework of our own, or use something prebuilt, or wait until Django offers a more officially supported mechanism to influence the models of reusable apps .

All in all, this is a big topic with low ROI, hence a low priority.

varunsaral:
@pennersr understood, this require more through understanding and will require much more efforts. I will still make progress in understading it. Thanks

Akay7:
If it's okay always have an external decencies than it could simply implemented with https://github.com/openwisp/django-swappable-models . Or we not sure yet what is the right way to do that?

pennersr:
The last release of `django-swappable-models` is from 2021. Also, before discussing implementation, we need to talk use cases first.

Akay7:
> The last release of

To support the latest Django, there is not much code and no changes required.

And there are packages that use it as well, for example [django-cities] (https://github.com/coderholic/django-cities)

As a use case, I would like to have a uuid as pk for models that are required for django-allauth.

pennersr:
For changing the PK fields, we wouldn't need to completely swap the models. A simple setting `ALLAUTH_PK_FIELD = "django.db.models.fields.UUIDField"` could very well cover this.

pennersr:
I don't think this PR is salvageable, it is far too dated. Also, the discussion over at #2997 needs to be completed first. Closing.

Issue #1321, opened 2016-03-09
username could be utf-8?


tjwudi:
I am using allauth for registration and it turns out that one could register with a utf-8 username.

<img width="328" alt="screen shot 2016-03-09 at 10 35 54 pm" src="https://cloud.githubusercontent.com/assets/5257816/13638450/513bb440-e647-11e5-8297-6661dc2d03f4.png">

I also used django's generic `UpdateView` to update user's profile, which will actually complain about utf-8 username. This leads to a bad inconsistency. According to [django docs](https://docs.djangoproject.com/en/1.9/ref/contrib/auth/):

> Usernames may contain alphanumeric, _, @, +, . and - characters.

tjwudi:
For the record, I didn't use any of the custom form but just the allauth one. All I did was changing the template to extend a different layout template.

tjwudi:
Had to use pdb to trace, and found that `username` was encoded to UTF-8. Then it could pass `USERNAME_REGEX` check.

tjwudi:
Here is a temporary solution. Set the following Adapter class as `ACCOUNT_ADAPTER` in your **settings.py**.

```
import allauth.account.adapter
import re

from django.utils.translation import ugettext_lazy as _
from django import forms


# https://github.com/pennersr/django-allauth/issues/1321
USERNAME_REGEX = re.compile(r'^[\w.@+-]+$')

class Adapter(allauth.account.adapter.DefaultAccountAdapter):
    def clean_username(self, username, shallow=False):
        username = super(Adapter, self).clean_username(username,
                shallow)
        if not USERNAME_REGEX.match(username):
            raise forms.ValidationError(_("Usernames can only contain "
                                          "letters, digits and @/./+/-/_."))
        return username
```

pennersr:
https://codeberg.org/allauth/django-allauth/issues/1321

PR #1322, opened 2016-03-09
Remove re.UNICODE

tjwudi:
To keep consistent with django's regex rule. See  [#1321](https://github.com/pennersr/django-allauth/issues/1321)


tjwudi:
I understand we could override `Adapter.clean_username` but I think remove it by default is better.

PR #1340, opened 2016-03-29
fix account signals

Hsinfu:
Here are some details.

1. account/signals.py:    revise email_confirmed and email_confirmation_sent
2. account/models.py:   add request and signup parameters to email_confirmation_sent
3. account/views.py:      add the ImmediateHttpResponse exception for form_valid function in SignupView

Hi @pennersr, see if there is something missing or misunderstanding. Thanks.

### DIFF
--- a/allauth/account/signals.py
+++ b/allauth/account/signals.py
@@ -9,8 +9,9 @@ password_set = Signal(providing_args=["request", "user"])
 password_changed = Signal(providing_args=["request", "user"])
 password_reset = Signal(providing_args=["request", "user"])
 
-email_confirmed = Signal(providing_args=["email_address"])
-email_confirmation_sent = Signal(providing_args=["confirmation"])
+email_confirmed = Signal(providing_args=["request", "email_address"])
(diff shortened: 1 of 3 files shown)


allauth/account/signals.py
```diff
@@ -9,8 +9,9 @@
 password_changed = Signal(providing_args=["request", "user"])
 password_reset = Signal(providing_args=["request", "user"])
 
-email_confirmed = Signal(providing_args=["email_address"])
-email_confirmation_sent = Signal(providing_args=["confirmation"])
+email_confirmed = Signal(providing_args=["request", "email_address"])
+email_confirmation_sent = Signal(providing_args=["request", "signup",
+                                                 "confirmation"])
```
pennersr: Sorry for nitpicking on details, but I would prefer to have them in order request-confirmation-signup, as this signal is about the confirmation, the signup is only adding extra detail.

pennersr: As we are breaking the existing signal signature, this needs a note in the `ChangeLog.rst` (backwards incompatible changes).

PR #1347, opened 2016-04-09
Make Facebook login more flexible

kangfend:
This PR make facebook login more flexible. If `Require App Secret` in our Facebook app is set to `Yes`, we need to set `REQUIRE_APP_SECRET` to `True` like this:
```python
SOCIALACCOUNT_PROVIDERS = {
    'facebook': {
        'REQUIRE_APP_SECRET': True,
        'VERSION': 'v2.4'
    }
}
```

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -86,6 +86,7 @@ Sam Solomon
 Seizan Shimazaki
 Serafeim Papastefanos
 Stuart Ross
+Sutrisno Efendi
 Terry Jones
 Tomas Babej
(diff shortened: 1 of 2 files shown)


pennersr:
Wondering, why is it needed to make this configurable? FB seems to be working just fine when we always pass along the proof ...

kangfend:
Ah sorry, I think I was wrong. Few days ago I always failed to login if appsecret_pool is included in params, so I make that variable configurable. But now everything is fine.

PR #1348, opened 2016-04-11
Added Salesforce provider

mgargenta:


pennersr:
This one breaks the build. Could also use some notes in `providers.rst`. Btw, don't forget to add yourself to the AUTHORS file. Thanks.

PR #1354, opened 2016-04-22
Adding default scope of user:email to requests with QUERY_EMAIL as true

jeffnb:
If the github user fetched does not have a public email and ACCOUNT_EMAIL_REQUIRED is true the view calls [get_email](https://github.com/pennersr/django-allauth/blob/2f433118ca344b837772601ac743768b2e974c66/allauth/socialaccount/providers/github/views.py#L26).  The method then tries to retrieve the email via the email endpoint which will 404 unless either user or user:email is added to the scope.  The code then throws a `KeyError` when attempting to [access the expected data](https://github.com/pennersr/django-allauth/blob/2f433118ca344b837772601ac743768b2e974c66/allauth/socialaccount/providers/github/views.py#L32).

This change simply adds the "user:email" scope to the OAUTH2 request to ensure the email is returned.

### DIFF
--- a/allauth/socialaccount/providers/github/provider.py
+++ b/allauth/socialaccount/providers/github/provider.py
@@ -1,6 +1,7 @@
 from allauth.socialaccount import providers
 from allauth.socialaccount.providers.base import ProviderAccount
 from allauth.socialaccount.providers.oauth2.provider import OAuth2Provider
+from allauth.socialaccount.app_settings import QUERY_EMAIL
 
 
 class GitHubAccount(ProviderAccount):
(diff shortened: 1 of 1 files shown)


jeffnb:
@pennersr This could use to go in.  When writing the test for this it turns out that `QUERY_EMAIL` is either not called or does not change when `@override_settings(ACCOUNT_EMAIL_REQUIRE)`.  I am more than willing to add a quick test if you have an idea how to make the `QUERY_EMAIL` property change in the test.

pennersr:
You are importing the QUERY_EMAIL property directly, and as a result it won't change anymore. Instead, simply import `app_settings` and invoke `app_settings.QUERY_EMAIL`.

pennersr:
See d629de7b1c463d7fe1338a3e0af8cea3456a7ce1 -- if you still have the test lying around, please file a pull request.

PR #1358, opened 2016-04-27
Add Eve Online provider

paxswill:
Not even sure if this is the sort of provider you'd want included, but I added an [Eve Online](http://www.eveonline.com) provider using their [SSO API](https://eveonline-third-party-documentation.readthedocs.org/en/latest/sso/index.html). This API is a little different from most of the other providers in that a single person can authenticate with multiple identities (any of their characters in-game)[0]. The biggest difference that most users of django-allauth might encounter is that there is not a way to get an email address for a user.

The biggest area I'm looking for review is how I'm mapping a character's information to a user in `providers.py`. Is there other information I should be extracting? The only information I can easily get is a character's name, an ID number for that character (which will not change if a character is transferred to another account) and a "[CharacterOwnerHash](https://eveonline-third-party-documentation.readthedocs.org/en/latest/sso/obtaincharacterid.html)" (which *will* change if the character is transferred to a new account). Right now I'm mapping the UID to this hash, and a user's name (the `name` attribute, not `username`) to the character's name. Should I map the `username` to the character's ID?

0: The most recent release expanded the OAuth API so that it's technically possible to log in as an account, but there isn't an identifier associated with this, only a set of the characters on an account. Players commonly have multiple game accounts with multiple characters, and can add or remove characters on those accounts at will.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -94,5 +94,6 @@ Udi Oron
 Vuong Nguyen
 Volodymyr Yatsyk
 Wendy Edwards
+Will Ross
 Yuri Kriachko
 Yaroslav Muravsky
--- a/allauth/socialaccount/providers/eveonline/models.py
+++ b/allauth/socialaccount/providers/eveonline/models.py
@@ -1,0 +1,1 @@
+# Create your models here.
--- a/allauth/socialaccount/providers/eveonline/provider.py
+++ b/allauth/socialaccount/providers/eveonline/provider.py
@@ -38,4 +38,4 @@ class EveOnlineProvider(OAuth2Provider):
                     username=data.get('CharacterID'))
 
 
-providers.registry.register(GitHubProvider)
+providers.registry.register(EveOnlineProvider)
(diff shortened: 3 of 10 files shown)


pennersr:
Thanks. I have to find some time to read into their API a bit, but a quick scan shows up this:

"The original web site will only ever get your character, they never see your account name or password. The original web site will not know what account that character is on or have any way, from us at least, of linking that character to any other character on the same account."

So, apparently, it is intentional that websites building our their API only get access to character data, not the account behind it. Given that, I think your choice of using CharacterOwnerHash is sound. If you were to transfer your character to somebody else, that does not need to imply that you wish to transfer all your accounts created on other sites using that account, as there may be personal content attached. 

So all in all, things are looking good. A few points:

- [ ] The build is failing on a few pep8 errors
- [ ] Can you add this provider to test_settings.py ?
- [ ] Should be added to installation.rst and overview.rst as well.
- [ ] Don't forget to add yourself to the AUHORS file

Thanks again...

PR #1368, opened 2016-05-12
Fix Oauth2 breakage with django-rest-auth

mdentremont:
- django-rest-auth's SocialLoginSerializer instantiates login adapters
  without passing parameters
- OAuth2Adapter's `__init__` takes a request, but OAuth2Adapter doesn't
  actually use self.request anywhere
- Make request optional to match what is done in other adapters
- shopify and weixin were the only subclasses which used `self.request`,
  but in both cases there was a request instance availble elsewhere

### DIFF
--- a/allauth/socialaccount/providers/oauth2/views.py
+++ b/allauth/socialaccount/providers/oauth2/views.py
@@ -29,7 +29,7 @@ class OAuth2Adapter(object):
     basic_auth = False
     headers = None
 
-    def __init__(self, request):
+    def __init__(self, request=None):
         self.request = request
 
     def get_provider(self):
--- a/allauth/socialaccount/providers/shopify/views.py
+++ b/allauth/socialaccount/providers/shopify/views.py
@@ -44,6 +44,7 @@ class ShopifyOAuth2Adapter(OAuth2Adapter):
         return self._shop_url('/admin/shop.json')
 
     def complete_login(self, request, app, token, **kwargs):
+        self.request = request
         headers = {
             'X-Shopify-Access-Token': '{token}'.format(token=token.token)}
         response = requests.get(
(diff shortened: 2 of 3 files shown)


mdentremont:
@pennersr Thoughts?

mdentremont:
@pennersr Thoughts?

mdentremont:
@pennersr Thoughts?

mdentremont:
@pennersr Thoughts?

pennersr:
I must admit, I don't like the direction being taken. Points:

- "Make request optional to match what is done in other adapters" -- while it is true that you'll find `request=None` here and there, mostly that is due to historical reasons. In earlier versions, the constructor did not take a request at all, and not all places have been converted to passing it, or, cannot be converted yet due to breaking of backwards compatibility.

- This change (assigning to self.request) feels a bit like a hack:

        def complete_login(self, request, app, token, **kwargs):
                self.request = request
     
I am not completely overseeing the problem that is being addressed here. Probably this is it:

"django-rest-auth's SocialLoginSerializer instantiates login adapters without passing parameters"

But, if that is the root cause, shouldn't that be directly addressed?

mdentremont:
@pennersr I guess a better summary would be that since the constructor did not take a request, when the request parameter was added, it broke backwards compatibility which caused a breakage in the OAuth2 lib.

One way to potentially get rid of the `self.request = request` hack would be to convert the properties inside of that class to regular methods and simply pass along the request object.

When I made this PR (several months ago) the hack seemed more feasible as it undid the backwards compatibility breakage -- but at this point at least one of the other libs has adapted so this change might not be useful anymore.

pennersr:
"but at this point at least one of the other libs has adapted so this change might not be useful anymore." -- do you happen to have a link handy to pointing to an issue in these upstream libraries?

mdentremont:
https://github.com/Tivix/django-rest-auth/commit/50087549e82c8f2b42ea8a6748cc66cbe3f3e472

pennersr:
Thanks. Then I think we can safely close the issue... you're correct that backwards compatibility was broken, but at least allauth internals are now a bit more clean, and the issue is being addressed (or can be addressed) upstream.

PR #1376, opened 2016-05-18
Force reauthentication by Google, not merely presentation of offline-access consent screen

peckjon:
Setting prompt=consent during google authentication does not cause a reauthentication prompt to appear; instead, it presents an "allow offline access" consent form.  Furthermore, setting access_type=offline interferes with the provider configuration value google.AUTH_PARAMS.access_type

Setting prompt=select_account, however, will display a screen in which the user can actually select the user identity with which to log in (even if the user is not using multiple Google accounts, they are still presented with a selection).

While not as good as clearing the google auth cookie entirely, this at least is closer to the spirit of reauthentication, and more intuitive to the user.

### DIFF
--- a/allauth/socialaccount/providers/google/provider.py
+++ b/allauth/socialaccount/providers/google/provider.py
@@ -38,8 +38,7 @@ class GoogleProvider(OAuth2Provider):
         ret = super(GoogleProvider, self).get_auth_params(request,
                                                           action)
         if action == AuthAction.REAUTHENTICATE:
-            ret['prompt'] = 'consent'
-            ret['access_type'] = 'offline'
+            ret['prompt'] = 'select_account'
         return ret
 
     def extract_uid(self, data):


peckjon:
@pennersr : re-auth can be truly forced by using a login link of the form https://www.google.com/accounts/Logout?continue=https://appengine.google.com/_ah/logout?continue={{ request.is_secure|yesno:"https://,http://" }}{{ request.get_host }}/accounts/google/login

...but boy is that ugly, and probably brittle (www.google.com/accounts/Logout doesn't allow external URLs in the continue param, but appengine.google.com/_ah/logout does... for now...)

pennersr:
Hmm, I have yet to test this but I recall it used to work. Did you see this?: https://developers.google.com/identity/protocols/OpenIDConnect#re-consent

peckjon:
prompt=consent redisplays the consent form; it doesn't force Google re-login

pztrick:
@peckjon I think we need some way to reconcile this use-case and re-authenticating in order to retrieve a new refresh token. Google only sends a refresh token one time (#308) and subsequent logins (including this P/R introducing `ret['prompt'] = 'select_account'`) do not receive a new refresh token -- only an access token. (In fact, it looks like the access token I am receiving is only good for one hour!)

I had originally authored P/R #1303 to ensure re-authentication requests would generate an OAuth2 request which would result in a new refresh token (rather than blank refresh token from Google).

I will try some other combinations of auth params that would at least let me be able to get a new refresh token without altering library, but so far I am getting complaints from Google like `Conflict params: approval_prompt and prompt`.

I personally think `AuthAction.REAUTHENTICATE` should fetch a new refresh token, but am not certain about other use cases here. I do understand this P/R was intended to allow users to toggle between multiple Google Accounts (?) and that behavior should be retained?

PR #1380, opened 2016-05-23
Add Battle.net provider

### LINKED ISSUE #1371: Battle.net support
We have been using django-allauth on our sites to log in with Battle.net OAuth2.
Here's the code:

* https://github.com/HearthSim/allauth-battlenet
* https://pypi.python.org/pypi/allauth_battlenet

If there are no objections, I'd like to upstream it here.

jleclanche:
Closes #1371

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -46,6 +46,7 @@ James Rivett-Carnac
 James Thompson
 Jannis Leidel
 Jeff Triplett
+Jerome Leclanche
 John Bazik
 John Whitlock
 Jonas Aule
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -6,6 +6,8 @@ Note worthy changes
 
 - Added Weixin provider.
 
+- Added Battle.net provider.
+
 - Django's password validation mechanism (see `AUTH_PASSWORD_VALIDATORS`) is now
   used to validate passwords.
 
--- a/allauth/socialaccount/providers/battlenet/provider.py
+++ b/allauth/socialaccount/providers/battlenet/provider.py
@@ -1,0 +1,41 @@
+from allauth.socialaccount import providers
+from allauth.socialaccount.providers.base import ProviderAccount
+from allauth.socialaccount.providers.oauth2.provider import OAuth2Provider
+from allauth.socialaccount.adapter import DefaultSocialAccountAdapter
+
+
+class BattleNetSocialAccountAdapter(DefaultSocialAccountAdapter):
+    def populate_user(self, request, sociallogin, data):
+        user = super().populate_user(request, sociallogin, data)
+        if "battletag" in data:
+            user.username = data["battletag"]
+        return user
+
+    def save_user(self, request, sociallogin, form=None):
+        user = super().save_user(request, sociallogin, form)
+        user.username = sociallogin.account.extra_data["battletag"]
+        user.save()
+
+        return user
+
+
+class BattleNetAccount(ProviderAccount):
+    pass
+
+
+class BattleNetProvider(OAuth2Provider):
+    id = "battlenet"
+    name = "Battle.net"
+    account_class = BattleNetAccount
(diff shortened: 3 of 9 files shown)


jleclanche:
@pennersr I don't understand why travis is failing. Any idea?

```pytb

Traceback (most recent call last):
  File "/home/travis/build/pennersr/django-allauth/allauth/socialaccount/tests.py", line 213, in test_account_refresh_token_saved_next_login
    self.test_account_tokens(multiple_login=True)
  File "/home/travis/build/pennersr/django-allauth/allauth/socialaccount/tests.py", line 182, in test_account_tokens
    self.login(self.get_mocked_response(), process='connect')
  File "/home/travis/build/pennersr/django-allauth/allauth/socialaccount/tests.py", line 217, in login
    resp = self.client.get(reverse(self.provider.id + '_login'),
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/django/core/urlresolvers.py", line 551, in reverse
    return iri_to_uri(resolver._reverse_with_prefix(view, prefix, *args, **kwargs))
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/django/core/urlresolvers.py", line 468, in _reverse_with_prefix
    (lookup_view_s, args, kwargs, len(patterns), patterns))
NoReverseMatch: Reverse for 'battlenet_login' with arguments '()' and keyword arguments '{}' not found. 0 pattern(s) tried: []
```

pennersr:
Where did you see that stack trace? In Travis, I see the following:

    allauth/socialaccount/providers/battlenet/views.py:59:80: E501 line too long (87 > 79 characters)
    allauth/socialaccount/providers/battlenet/views.py:61:80: E501 line too long (81 > 79 characters)

So it is failing on pep8 formatting errors.

jleclanche:
@pennersr hm I'm sure I saw it on travis somewhere...

Fixed the pep8 issues.

jleclanche:
@pennersr I found out during mass testing that the `username` (aka battletag) may be null if the account was *just* created, in which case in the current code it's passed down to allauth as `None`, and crashes.

I'm not sure what to do about this. We don't get the user's email with Battle.net oauth afaik. Suggestions? Is there a "Pick your username" flow?

jleclanche:
@pennersr ping, see above - I really don't know what to do about this.

Akay7:
@pennersr I see same about VK or FB, when django-auth don't get username from server, they just set first_name as your username, if it don't get first_name, then it generate user1, user2 and etc. Maybe you can try to do same?

jleclanche:
So I spent the afternoon tracking this down. Here are my findings.

Turns out, allauth has a builtin mechanism to generate unique usernames if one is not present at the time of signup. It wasn't kicking in because of the way I was overriding the adapter.

I found out we are supposed to use `extract_common_fields()` to get the username, not the battletag (so this PR is technically incorrect). However, when I switched to using that method, I was no longer getting the usernames saved at all (and the default unique username generation was kicking in).

It took me a long time to track this down to [here](https://github.com/pennersr/django-allauth/blob/6d65e3d15557e3f6af0f0d65a95b4bcfba18d5f5/allauth/socialaccount/helpers.py#L34-L38), the username regex validation. Battle.net battletags look like `SomeUsername#1234`. The hash in the middle fails the [default regex](https://github.com/pennersr/django-allauth/blob/6d65e3d15557e3f6af0f0d65a95b4bcfba18d5f5/allauth/account/adapter.py#L44) `^[\w.@+-]+$`.

I toyed around the idea of overriding that regex to `^[\w]+#[\d]+$`, but requiring clients to override the default account adapter is too much hassle. So we're back where we started: Overriding `save_user()`.

I have a PR which removes the `populate_user` overrides, adds some comments and fixes the aforementioned issue. I'm testing it in prod right now, will file it once it's good to go.

PR #1381, opened 2016-05-24
Store bitbucket refresh_token upon login

caruccio:


pennersr:
The refresh token does not belong in the `extra_data` of the `SocialAccount`. It should be stored in `SocialToken.token_secret`

PR #1387, opened 2016-05-30
adjust accout email

Alchemist85K:
I'm using django-allauth 0.24.1 and django 1.7.0.
Lately I changed my database to utf8mb4 and migration was failed because of the unique property in email inside accout model. 
I needed some options whose default value is 191 to set accout email. 
I have to use unique option of utf8mb4 and accout email. 
Please apply this soon. thanks.


Alchemist85K:
Since 0.25.0 version is not compatible with 0.24.1 version. please apply this patch as 0.24.2 version.

pennersr:
Related: https://code.djangoproject.com/ticket/18392

pennersr:
You've changed the default to 191. By default it should be 254 to be compliant with RFCs 3696 and 5321. MySQL is at fault here -- we should not force its quirks onto other databases by default.

Additionally, the migration should not be changed as changing the setting would change history, whereas you would expect a new migration to be created.

- [ ] Change default to 254

- [ ] Don't alter migration

Alchemist85K:
@pennersr 
Thanks for your detailed explaination. But I need the way to set max_length of account email.  Can you tell me how to do this?

pennersr:
I am okay with making this configurable as you have done. The only thing I would like to see different is to keep the default to 254 so that people running PostgreSQL by default are not hindered by quirks of MySQL.

Alchemist85K:
@pennersr 

I changed my prior commit as following. 
It keeps the default value to 254 and has configurable setting. 
I would like to see this PR being merged. :)

Since 0.25.0 version is not compatible with 0.24.1 version. Please apply this patch as 0.24.2 version.

Thank you so much for your advice.

allauth/account/migrations/0003_email_max_length.py
```diff
@@ -0,0 +1,32 @@
...
+    operations = [
+        migrations.AlterField(
+            model_name='emailaddress',
+            name='email',
+            field=models.EmailField(unique=UNIQUE_EMAIL, max_length=EMAIL_MAX_LENGTH, verbose_name='e-mail address'),
+
+        ),
+    ]
+
+    if not UNIQUE_EMAIL:
```
pennersr: Instead of creating a new migration, migration `0002_email_max_length` should be altered to simply use the setting over at:

```
            field=models.EmailField(unique=UNIQUE_EMAIL, max_length=254, verbose_name='e-mail address'),
```

This way, most deployments are not impacted or bothered by additional migrations which don't change the database at all. Also, if you were to keep the above migration and upgrade allauth and alter the setting thereafter, things still wouldn't work as the above migration did already run with a default of 254.

Alchemist85K:
change file.
remove 0003 migrations file and change max_length setting in 0002 migration file.

PR #1398, opened 2016-06-07
Provider for23and me

Akay7:
Added provider for 23andMe


pennersr:
Thanks. Can we rename the module to avoid this?:

    >>> from allauth.socialaccount.providers.23andme.views import *
      File "<stdin>", line 1
       from allauth.socialaccount.providers.23andme.views import *
                                              ^  
    SyntaxError: invalid syntax

`twenty3andme` or `twentythreeandme` ?

pennersr:
Can you also add an entry for the provider over at `overview.rst`? Btw, don't forget to add yourself to the authors file.

Akay7:
Okay I will do all necessary work, as soon as possible.

Akay7:
I made necessary changes.

pennersr:
Thanks. I made a small change so that the URLs look a bit cleaner (/23andme/ instead of /twentythreeandme/).

Akay7:
Very good!!! Thank you for help

PR #1407, opened 2016-06-20
Allow more dynamic email cleaning

kronok:
I've been forking django-allauth for a couple years mainly for these changes, and I think this makes the clean_email function a lot more dynamic and fits its purpose better.

Moving the app_settings.UNIQUE_EMAIL check to the adapter seems to make more sense, while removing the hook from the password reset and add email forms, because the password reset form shouldn't share a validating hook with the signup form, and the add email form already does all the necessary checks it needs to do before allowing it.

My use case for this is I allow "newsletter only" user accounts, which is essentially a User without a username and password set, and have it subscribe to a mailing list. When that same person decides to create a normal user account (normal signup), it need to check if the email exists, and if it does, it checks if it's a "newsletter only" user account to see if it needs to raise an "email taken" validation error or convert a newsletter only account into a normal one. Moving the UNIQUE_EMAIL validation check to a more customizable function allows this beautifully.

### DIFF
--- a/allauth/account/adapter.py
+++ b/allauth/account/adapter.py
@@ -299,7 +299,7 @@ class DefaultAccountAdapter(object):
         return password
 
     def validate_unique_email(self, email):
-        if email and email_address_exists(email):
+        if email_address_exists(email):
             raise forms.ValidationError(self.error_messages['email_taken'])
         return email
 
--- a/allauth/account/adapter.py
+++ b/allauth/account/adapter.py
@@ -31,7 +31,7 @@ from ..compat import validate_password
 from ..utils import (build_absolute_uri, get_current_site,
                      generate_unique_username,
                      get_user_model, import_attribute,
-                     resolve_url)
+                     resolve_url, email_address_exists)
 
 from . import app_settings
 
@@ -50,7 +50,9 @@ class DefaultAccountAdapter(object):
         'username_taken':
         AbstractUser._meta.get_field('username').error_messages['unique'],
         'too_many_login_attempts':
-        _('Too many failed login attempts. Try again later.')
+        _('Too many failed login attempts. Try again later.'),
+        'email_taken':
+        AbstractUser._meta.get_field('email').error_messages['unique'],
     }
 
     def __init__(self, request=None):
@@ -282,6 +284,9 @@ class DefaultAccountAdapter(object):
         Validates an email value. You can hook into this if you want to
         (dynamically) restrict what email addresses can be chosen.
         """
+        if app_settings.UNIQUE_EMAIL:
+            if email and email_address_exists(email):
+                raise forms.ValidationError(self.error_messages['email_taken'])
         return email
 
     def clean_password(self, password, user=None):
--- a/allauth/account/adapter.py
+++ b/allauth/account/adapter.py
@@ -284,9 +284,6 @@ class DefaultAccountAdapter(object):
         Validates an email value. You can hook into this if you want to
         (dynamically) restrict what email addresses can be chosen.
         """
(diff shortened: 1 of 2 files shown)


allauth/account/forms.py
```diff
@@ -363,7 +355,6 @@ class AddEmailForm(UserForm):
 
     def clean_email(self):
         value = self.cleaned_data["email"]
-        value = get_adapter().clean_email(value)
```
pennersr: We cannot remove this one. There are use cases where people override the `clean_email` such that you can only add @theirowncompany.com based email addresses. With this change, any email address can be added.

allauth/account/forms.py
```diff
@@ -452,7 +443,6 @@ class ResetPasswordForm(forms.Form):
 
     def clean_email(self):
         email = self.cleaned_data["email"]
-        email = get_adapter().clean_email(email)
```
pennersr: I would like to keep project specific email cleaning in tact here as welll..

allauth/account/forms.py
pennersr: I think it will become messy if the email unique handling is made part of the `clean_email`, as the `clean_email` method is unaware of the use case (signup, add-email, password reset). You would need to pass along an additional parameter to indicate whether or not the email is going to be added, or just used for input cleaning. Therefore, I think it is best to keep the original `clean_email` as is, and add a dedicated method for the uniqueness checking: `validate_email_uniqueness`?

kronok:
Yeah, you're right. I think a dedicated method is the best way to approach this so nothing breaks with anyone's use cases. I was trying too hard to keep the amount of changes to a minimum.

I'll check the uniqueness in the same place, but calls something like `get_adapter().validate_email_uniqueness(email)`. I think that would allow the customization I need for my use case and not break anything for anything else for anyone.

allauth/account/forms.py
```diff
@@ -270,6 +270,8 @@ def clean_username(self):
     def clean_email(self):
         value = self.cleaned_data["email"]
         value = get_adapter().clean_email(value)
+        if app_settings.UNIQUE_EMAIL:
+            get_adapter().validate_unique_email(value)
```
kronok: May look a little confusing since I changed things and then changed them back, but this is the main difference between master and my pull.

This is where the following code used to be. It's been moved to it's own method on the adapter. Validation check is done in the adapter just like a lot of the other fields, and follows a similar pattern as the others.
```
        if email and email_address_exists(email):
            raise forms.ValidationError(self.error_messages['email_taken'])
```

allauth/account/adapter.py
```diff
@@ -296,6 +298,11 @@ def clean_password(self, password, user=None):
         validate_password(password, user)
         return password
 
+    def validate_unique_email(self, email):
+        if email and email_address_exists(email):
```
pennersr: Can we move the `if email` part back into the original place (forms) -- so that people overriding adapters don't have to be bothered with empty values?

allauth/account/forms.py
```diff
@@ -271,7 +271,7 @@ def clean_email(self):
         value = self.cleaned_data["email"]
         value = get_adapter().clean_email(value)
         if value and app_settings.UNIQUE_EMAIL:
-            get_adapter().validate_unique_email(value)
+            value = get_adapter().validate_unique_email(value)
```
kronok: Realized someone may want to do something nuts with validate_unique_email and return something different than the original value for maybe a strange use case, so setting the value to what returns (just returns the original email if valid by default) just to be safe.

pennersr:
Thanks

PR #1420, opened 2016-06-28
battlenet: Only set the battletag as username if available

jleclanche:
References #1380

### DIFF
--- a/allauth/socialaccount/providers/battlenet/provider.py
+++ b/allauth/socialaccount/providers/battlenet/provider.py
@@ -16,7 +16,9 @@ class BattleNetSocialAccountAdapter(DefaultSocialAccountAdapter):
 
 
 class BattleNetAccount(ProviderAccount):
-    pass
+    def to_str(self):
+        battletag = self.account.extra_data.get("battletag")
+        return battletag or super(BattleNetAccount, self).to_str()
 
 
 class BattleNetProvider(OAuth2Provider):
--- a/allauth/socialaccount/providers/battlenet/provider.py
+++ b/allauth/socialaccount/providers/battlenet/provider.py
@@ -5,17 +5,13 @@ from allauth.socialaccount.adapter import DefaultSocialAccountAdapter
 
 
 class BattleNetSocialAccountAdapter(DefaultSocialAccountAdapter):
-    def populate_user(self, request, sociallogin, data):
-        user = super().populate_user(request, sociallogin, data)
-        if "battletag" in data:
-            user.username = data["battletag"]
-        return user
-
     def save_user(self, request, sociallogin, form=None):
         user = super().save_user(request, sociallogin, form)
-        user.username = sociallogin.account.extra_data["battletag"]
-        user.save()
 
+        battletag = sociallogin.account.extra_data.get("battletag")
+        if battletag:
+            user.username = battletag
+            user.save()
         return user
 
 


allauth/socialaccount/providers/battlenet/provider.py
```diff
@@ -5,17 +5,13 @@
 
 
 class BattleNetSocialAccountAdapter(DefaultSocialAccountAdapter):
```
pennersr: I must admit, I completely missed out on this class when reviewing and merging the provider. It is uncommon to provide a custom adapter with a provider. Adapters are meant to be generic, mostly provider independent. If you need to influence the way the username is chosen, then make sure that `extract_common_fields` of the `BattleNetProvider` returns a `'username'` key. So let's drop this class completely, and move the username logic over there...

jleclanche: @pennersr Did you read my comment [here](https://github.com/pennersr/django-allauth/pull/1380#issuecomment-226943798)? It explains why I am still using the adapter.

jleclanche: We can drop the adapter if we do one of two things:

* Either we change the default username regex to accept `#` as a valid character
* Or we build a mechanism for socialaccount providers to somehow accept a custom username regex.

The former would be easy, but affect all other providers. The latter would be fairly hard, and I'm not up to the task.

Regardless of the outcome, this PR should be merged IMO as it's more correct than what's on master right now and prevents a crash.

jleclanche:
@pennersr ping, see my above comments.

jleclanche:
@pennersr ping, I need to know how we're going to proceed :)

jleclanche:
@pennersr Could you comment on this please?

pennersr:
The issue in your case is that there is one provider, the battlenet one, which apparently needs full control over the username. This is a project specific choice, and not something that is generally useful/desired. For example, it is just as well possible for users to hookup Twitter or Google accounts, next to the battlenet ones. In such case, the battlenet usernames can already be in use.

I don't like for allauth to include project specific code on board. But, your use case is a perfectly valid one, so I would like to understand what can be done to make things work on your end without having to resort to forking the project.

The adapter can simply be included in your own project. As for the regular expression, it can be overriden as well:

```
class DefaultAccountAdapter(object):

    # Don't bother turning this into a setting, as changing this also
    # requires changing the accompanying form error message. So if you
    # need to change any of this, simply override clean_username().
    username_regex = re.compile(r'^[\w.@+-]+$')
```

Does that help?

jleclanche:
@pennersr Two additional things to keep in mind:

* "Battletags" of the format "username#number" are starting to get common. For example, Discord uses them and they also have [an oauth2 API](https://discordapp.com/developers/docs/topics/oauth2). So when discord support is implemented, it will face those same issues.
* The *only* information we get from the battlenet API is the Battletag. Email or any kind of name is not available. So in the case the battletag is null, the fallback right now will be `user`, `user2`, `user3` etc. Very unfriendly, *but* the case where the battletag is null should be exceedingly rare.

My recommendation at this point would be to allow the `#` character in the default username regex. I've been using [allauth-battlenet](https://github.com/hearthsim/allauth-battlenet) in [production](https://hsreplay.net/) for the past few weeks and the current version of the code, which is what I PR'd here, works wonderfully well and there's no issue I could find associated with `#` in usernames with allauth other than the regex.

Overriding `clean_username()` as you suggested would require shipping a custom adapter. I can certainly do it in my own project, but the issue we're facing here is one faced by *all* battlenet allauth users. If you want to switch to `extract_common_fields`, it's not an option. 

I'm not sure what other mechanism you could put in place to fit this use case, I'm not familiar enough with the allauth codebase.

jleclanche:
18773e363dd9411f33e07cd47892bdb815a06734 is an example of what it would look like if you were to accept a change to the username regex.

Like I said, though, I recommend merging the PR as-is, it's more correct than what is currently on master. We can keep investigating other approaches separately as this seems to be a more general issue.

jleclanche:
@pennersr I'm a little disappointed 0.26 went through without this PR getting merged. The new Battle.net provider has a fairly important bug in it right now.

Can I once again urge you to merge this as-is for the next minor release? The PR is a net improvement over what's on master right now.

pennersr:
I am sorry... I am under quite a time squeeze lately and I have to be little bit selective in what to pick up. As indicated, I don't like the idea of having a custom adapter on board. I would really like to get rid of that. Additionally, I do not want to allow for '#' in the username by default, that does not sound like a sensible default, and if a project needs this, it should be easy to change this via the adapter.

Having said that, given that the BattleNet adapter is already on board this PR does not make things worse, so if it helps on the short term I will go ahead and merge the PR.  At a later time I would like to revisit this ... possibly breaking backwards compatibility.

 I'll have to come back and discuss the issues you raised above at a later time...

jleclanche:
@pennersr Thanks! Please ping me if/when you decide to change this, or decide to allow for social backends to have their own username regexes.

PR #1422, opened 2016-07-01
Adds a setting to control internal messaes

cpinto:
This setting allows users of this app to have control over the sending of messages through django.contrib.messages.


pennersr:
There are already sufficient ways to stop allauth from sending out messages: remove the messages app from INSTALLED_APPS, or, put an empty template for the messages you want to kill in your templates, or, override the account adapter `add_message` and make sure it does nothing.

PR #1433, opened 2016-07-15
Add get_extra_context_data in ResetPasswordForm

bcanyelles:
This method is needed to avoid to override the whole save method
when you want extra context data into the email template.


pennersr:
You can already accomplish this without having to override the `save` method. Simply drop in your own `send_mail` adapter method, e.g.:

    def send_mail(self, template_prefix, email, context):
        context.update({...extra...})
        return super().send_mail(...)

PR #1452, opened 2016-08-03
possibility to ignore username validation and to take hash instead

bevedoni:
added possibility to ignore username validation and to use random hash instead

in one of our projects, the username validation was the bottleneck in high traffic, I patched the generate_unique_username() to return os.urandom(28).encode('hex') back in time. In this pr, it is now possible to deactivate username validation via USERNAME_AS_HASH=True.


pennersr:
Hmm, my first though is that adding `validate_unique_username` to the account adapter would be more flexible and avoid a setting. We already have `validate_unique_email` in there, so this would be more in line...

pennersr:
See `DefaultAccountAdapter.generate_unique_username`. Simply use your own adapter and override that method with your hash generator .. no setting is required.

PR #1454, opened 2016-08-07
Django 1.10 compat fix for socialaccount.py

powellc:
Fixes an issue where request context is no longer a dictionary in Django 1.10, but still checks for the dict key first, so it's backwards compatible.


allauth/socialaccount/templatetags/socialaccount.py
```diff
@@ -52,7 +55,10 @@ def provider_login_url(parser, token):
 
 class ProvidersMediaJSNode(template.Node):
     def render(self, context):
-        request = context['request']
```
pennersr: This introduces some duplication.. how about adding a `get_request_from_context` to `allauth/compat.py` ?

pennersr:
Handled as part of 765d0693891d7d0e58492d3c2ab4e1ddb8582195

PR #1458, opened 2016-08-14
Html email confirmation

SUguy:
Correct files to change for html email confirmation

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -134,9 +134,14 @@ templates. Templates are named as follows::
     account/email/email_confirmation_message.txt
 
 In case you want to include an HTML representation, add an HTML
-template as follows::
+templates as follows::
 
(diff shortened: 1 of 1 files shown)


docs/advanced.rst
```diff
@@ -134,9 +134,14 @@ templates. Templates are named as follows::
...
 
 In case you want to include an HTML representation, add an HTML
-template as follows::
+templates as follows::
 
     account/email/email_confirmation_message.html
+    account/email/email_confirmation_signup_message.html
+    
+    confirmation_signup_message.html should contain :
+    {% include "account/email/email_confirmation_message.html" %}
```
pennersr: It is not recommended/obligatory to include that.... this is just a simple (read: lazy) way of making the signup message identical to the non-signup variant. Typically, you would want to differentiate between the two..

PR #1492, opened 2016-09-15
Auth dynamic auth_params documentation

jezenielzapanta:
This is a feature that I needed but needed to read the source code to find out.


jezenielzapanta:
Closing this, it seems like it will break things.

PR #1506, opened 2016-10-05
surveyMonkey provider

Bruledamien:
The goal of this fork was to add SurveyMonkey to the list of providers. If you think this can be a useful addition, or have any changes to suggest, feel free to come back to me.


allauth/socialaccount/providers/oauth2/client.py
```diff
@@ -30,13 +31,15 @@ def __init__(self, request, consumer_key, consumer_secret,
...
+        self.api_key = api_key
 
     def get_redirect_url(self, authorization_url, extra_params):
         params = {
             'client_id': self.consumer_key,
             'redirect_uri': self.callback_url,
             'scope': self.scope,
-            'response_type': 'code'
+            'response_type': 'code',
+            'api_key': self.api_key
```
pennersr: It is already possible to pass along additional parameters via the `extra_params`, I would prefer that above adding survey monkey specific keys in here.

allauth/socialaccount/migrations/0004_auto_20160919_1636.py
```diff
@@ -0,0 +1,20 @@
...
+
+    dependencies = [
+        ('socialaccount', '0003_extra_data_default_dict'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='socialapp',
+            name='key',
+            field=models.CharField(blank=True, help_text='Key, optional api_key', max_length=191, verbose_name='key'),
```
pennersr: Can't we use `SocialApp.key` for this?

allauth/socialaccount/providers/oauth2/client.py
```diff
@@ -73,6 +79,9 @@ def get_access_token(self, code):
             headers=self.headers,
             auth=auth)
 
+        print 'status_code', resp.status_code
+        print 'body', resp.text
```
pennersr: Print leftovers

PR #1507, opened 2016-10-06
battlenet: Use correct Battle.net API url for China

jleclanche:
Refs HearthSim/HSReplay.net#82


jleclanche:
I suspect `sea` is broken as well, but I have no way of testing that (and since the `us` provider encompasses all regions, there's no particular reason to set the region to `sea`).

PR #1512, opened 2016-10-07
Added Doximity Oauth2 provider.

rantonov:
Doximity is a professional network for Healthcare professionals in the US.  In many ways similar to LinkedIn but with a healthcare focus.


pennersr:
Thanks... could you also add an entry to the list of supported providers, over at `docs/overview.rst` ?

rantonov:
Added Doximity to overview.rst.

Issue #1008, opened 2015-06-30
Limit number of email verifications sent


HEdingfield:
A user can potentially sign up for an account at a site that uses allauth with someone else's e-mail address, and then attempt to login repeatedly, spamming that address with verification requests. Is it possible to build in an option that lets you limit the number of email verifications sent (e.g. 1 / hour. after the 2nd one)?

pennersr:
For email verification there is already an internal configuration (`COOLDOWN_PERIOD = timedelta(minutes=3)`) that limits the # of emails sent. I would be okay with making this a public setting in the form of `ACCOUNT_EMAIL_COOLDOWN_PERIOD`).

However, a similar situation holds for the password forgotten scenario as well. And in this case, allauth simply does not have the required state to determine if the user requested a password reset before.

You could keep track of this via the adapter method `send_mail`. E.g. keep track of what type of mail per user was sent last, e.g. using a cache backend. I do feel that this goes a bit beyond allauth scope, throttling requests et al should be handled at a level higher than allauth.

So I am closing this, feel free to open a PR to move the `COOLDOWN_PERIOD` public and add a note in the docs...

9mido:
PR #2631 should hopefully solve this if merged.

pennersr:
https://codeberg.org/allauth/django-allauth/issues/1008

PR #1517, opened 2016-10-12
Made COOLDOWN_PERIOD a public setting

asdfjackal:
As suggested in: #1008 I have made COOLDOWN_PERIOD public through ACCOUNT_EMAIL_COOLDOWN_PERIOD and added documentation for the new setting.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,6 @@ virtualenv
 *.prefs
 *.mo
 /.stfolder
+
+# venv installation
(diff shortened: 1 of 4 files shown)


pennersr:
The cooldown period is currently only used when `EMAIL_CONFIRMATION_HMAC` is explicitly set to `False`. I would like to phase out the non HMAC approach completely, rendering this setting useless.

PR #1524, opened 2016-10-27
Django 1.10 deprecation warning cleanups

dwaynebailey:
* is_anonymous and is_authorized and now properties - added a compatibility function
* urlresolver has moved - import used functions into the compatibility module
* on_delete - is explicit in Django 1.10, so use models.CASCADE, the current implicit method in all models and migrations.


unho:
@pennersr It would be great to have a release with this fix.

pennersr:
Okay: 70cb30ef51555c3798e768d0aae77a5bfa506502

PR #1533, opened 2016-11-16
Update installation.rst

leonardoRC:
More detailed description about site admin changes


pennersr:
Thanks, but using personal references like "Since I am..." and "I've" is not a good fit for documentation.

PR #1538, opened 2016-11-24
assignment_tag is deprecated

dwaynebailey:
See https://docs.djangoproject.com/en/1.9/releases/1.9/#assignment-tag

Introduces a compatibility layer for template.Library().assignment_tag()
to use simple_tag() in Django >= 1.9


dwaynebailey:
Likely this could be more elegantly placed into `allauth/compat`

Issue #1442, opened 2016-07-25
Revisit BattleNet adapter


pennersr:
BattleNet contains a custom adapter. See discussion over at #1380  ... TBD on how to proceed

pennersr:
https://codeberg.org/allauth/django-allauth/issues/1442

PR #1556, opened 2016-12-09
Assorted improvements to the Battle.net backend

jleclanche:
Alrighty, now that 8dc2f2d5cc3ce0e5e1b999129ceaa57ed4e75390 has landed I can go ahead with this :)

The first commit starts addressing #1442. It doesn't outright *remove* the adapter but it makes it way less nasty. The current adapter *always* causes the "user", "user1"... mechanism because the battletag is never saved. Additionally, the current adapter always forces at least two sql writes (first the user, then re-saves the user's battletag).

So we first allow overriding the `username_regex` attribute. Rather than make it a setting, we allow specifying it in the adapter, which I believe to be a reasonable solution.
With that done, we can change the Battle.net adapter to use that and get the battletag as the username data.

The next commits address the china issue:

- `valid_region` is simply a list of the valid region inputs for `battlenet_region`
- Appending `-cn` to the UIDs matching `cn` region is required because the UIDs can clash between US and China; if we don't do that, an account in china can take over an account in the US if running both on the same server.
- Finally, the last commit allows overriding the region in `/account/battlenet/login/` by specifying, for example, `?region=cn`. I'm not sure if there is existing functionality for this but I think it's acceptable.

### DIFF
--- a/allauth/socialaccount/providers/battlenet/provider.py
+++ b/allauth/socialaccount/providers/battlenet/provider.py
@@ -27,7 +27,11 @@ class BattleNetProvider(OAuth2Provider):
     account_class = BattleNetAccount
 
     def extract_uid(self, data):
-        return str(data["id"])
+        uid = str(data["id"])
+        if data.get("region") == "cn":
+            # China is on a different account system. UIDs can clash with US.
+            return uid + "-cn"
+        return uid
 
     def extract_common_fields(self, data):
         return {"battletag": data.get("battletag")}
--- a/allauth/socialaccount/providers/battlenet/provider.py
+++ b/allauth/socialaccount/providers/battlenet/provider.py
@@ -1,18 +1,6 @@
 from allauth.socialaccount import providers
 from allauth.socialaccount.providers.base import ProviderAccount
 from allauth.socialaccount.providers.oauth2.provider import OAuth2Provider
-from allauth.socialaccount.adapter import DefaultSocialAccountAdapter
-
-
-class BattleNetSocialAccountAdapter(DefaultSocialAccountAdapter):
-    def save_user(self, request, sociallogin, form=None):
-        user = super().save_user(request, sociallogin, form)
-
-        battletag = sociallogin.account.extra_data.get("battletag")
-        if battletag:
-            user.username = battletag
-            user.save()
-        return user
 
 
 class BattleNetAccount(ProviderAccount):
--- a/allauth/socialaccount/providers/battlenet/provider.py
+++ b/allauth/socialaccount/providers/battlenet/provider.py
@@ -34,7 +34,7 @@ class BattleNetProvider(OAuth2Provider):
         return uid
 
     def extract_common_fields(self, data):
-        return {"battletag": data.get("battletag")}
+        return {"username": data.get("battletag")}
 
     def get_default_scope(self):
         # Optional scopes: "sc2.profile", "wow.profile"
--- a/allauth/socialaccount/providers/oauth2/views.py
+++ b/allauth/socialaccount/providers/oauth2/views.py
@@ -43,6 +43,11 @@ class OAuth2Adapter(object):
         """
         raise NotImplementedError
 
+    def get_callback_url(self, request, app):
+        callback_url = reverse(self.provider_id + "_callback")
+        protocol = self.redirect_uri_protocol
+        return build_absolute_uri(request, callback_url, protocol)
+
     def parse_token(self, data):
         token = SocialToken(token=data['access_token'])
         token.token_secret = data.get('refresh_token', '')
@@ -67,10 +72,7 @@ class OAuth2View(object):
         return view
 
     def get_client(self, request, app):
-        callback_url = reverse(self.adapter.provider_id + "_callback")
-        callback_url = build_absolute_uri(
-            request, callback_url,
-            protocol=self.adapter.redirect_uri_protocol)
+        callback_url = self.adapter.get_callback_url(request, app)
         provider = self.adapter.get_provider()
         scope = provider.get_scope(request)
         client = OAuth2Client(self.request, app.client_id, app.secret,
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -23,6 +23,8 @@ Note worthy changes
 - The OAuth2Adapter class has gained a ``get_callback_url`` method for when
   customizing the callback URL is desired.
 
+- The Battle.net login backend now accepts the ``region`` GET parameter.
+
 Backwards incompatible changes
 ------------------------------
 
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -20,6 +20,9 @@ Note worthy changes
   username. For now, the default is set to ``True`` to maintain backwards
   compatibility.
 
+- The OAuth2Adapter class has gained a ``get_callback_url`` method for when
+  customizing the callback URL is desired.
+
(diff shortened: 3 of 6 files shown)


jleclanche:
@pennersr please review :)

jleclanche:
Do not merge

jleclanche:
@pennersr So the obvious problem I overlooked when specifying `/account/battlenet/login/?region=cn` we lose it after we return from upstream, ending up in `/account/battlenet/login/callback/?code=x&state=y` but missing `&region=cn`.

So I think I need to be able to specify the callback url. I see that the only way to do that right now is to override `OAuth2View.get_client()`, which sounds overzealous although a couple of providers do do that.

What do you think of adding a method `get_callback_url(self, request)` to OAuth2View?

jleclanche:
Disregard the last message. I added that method anyway because it's useful but it's not necessary for the rest of the commits.

Everything works right now. I have to wait for tomorrow for somebody to test in china, though.

I'm not a *huge* fan of how much code is involved in passing the region around, but I really do think it's the correct solution.

allauth/socialaccount/providers/battlenet/provider.py
```diff
@@ -1,18 +1,12 @@
...
 class BattleNetSocialAccountAdapter(DefaultSocialAccountAdapter):
-    def save_user(self, request, sociallogin, form=None):
-        user = super().save_user(request, sociallogin, form)
-
-        battletag = sociallogin.account.extra_data.get("battletag")
-        if battletag:
-            user.username = battletag
-            user.save()
-        return user
+    username_regex = re.compile(r"^[\w.]+#\d+$")
```
pennersr: If BattleNet requires this particular regexp, why not simply add it to the documentation. You could even go as far doing this:

```
from allauth.account.adapter import get_adapter

if get_adapter().uername_regex != 'this-one':
  raise ImproperlyConfigured()

```

Then, we the issue remains contained within this specific adapter/provider, and developers won't be able to accidentally wrongly configure things.

allauth/socialaccount/providers/oauth2/views.py
```diff
@@ -64,12 +64,17 @@ def view(request, *args, **kwargs):
                 return e.response
         return view
 
-    def get_client(self, request, app):
+    def get_callback_url(self, request, app):
```
pennersr: Wouldn't the OAuth2Adapter be a better place to move this into? You already have a BattleNetOAuth2Adapter ...

jleclanche: This particular change is not related to the BattleNetOAuth2Adapter - see my comments above. I left it in because I think it improves the code.

jleclanche:
@pennersr Replying to the aggregate comments regarding `username_regex`:

I agree that it doesn't ideally belong in the social adapter. However, we then need to figure out what to do for Battle.net by default. I think it's very user unfriendly to require writing a bunch of extra code in order to use the adapter - by removing the adapter, that would be what's needed.

There is one alternative: We drop the `#`. So by default, the username would be, instead of `adys#1234`, just `adys`. If there's already an `adys` in the system, I would *like* to pass the `1234` down to the unique username generator as a hint, so that we end up with `adys_1234` or `adys1234`, something like that. Or, we can always replace it with a dot, as the dot is not used upstream at all.

I'm not a fan of any of this - I have gotten to really like having the full Battle.net username available, battletag included. But maybe that's just me having gotten used to it. I don't mind, I can always override the adapter for my own project.

Either way, I need you to take a decision on this: Do we...
- Continue trying to find a way to have the hash by default?
- Replace the hash with a dot/dash/underscore?
- Strip it and fall back to the generator (and if so how?)?
- Accept `username_regex` as a valid feature and land this as is?

Again, keep in mind that whatever the outcome of this is, Discord is using a similar system for usernames so we have to find something that works for both (I intend to commit a Discord backend at some point).

allauth/socialaccount/helpers.py
```diff
@@ -20,40 +20,39 @@
...
-            username = user_username(sociallogin.user)
-            try:
-                get_account_adapter().clean_username(username)
-            except ValidationError:
-                # This username is no good ...
-                user_username(sociallogin.user, '')
-        # FIXME: This part contains a lot of duplication of logic
-        # ("closed" rendering, create user, send email, in active
-        # etc..)
+        return HttpResponseRedirect(url)
```
jleclanche: Can I introduce you to the wonderful, wonderful `q` library? :)

https://pypi.python.org/pypi/q

jleclanche: I find this style a ton cleaner and more readable. I can change it back if you insist, but I'm putting up a little fight ;)

pennersr:
How about the following approach.

- I'll prep a change where the username regex becomes pluggable, based on this: https://github.com/django/django/blob/1.10/django/contrib/auth/validators.py

- Then, you can provide a BattleNetUsernameValidator in your provider. Additionally, you can do an assert inside the constructor of your adapter that a battlnet username actually matches the currently configured username regex.

People have been asking for a configurable regex anyway, so this tackles 2 issues in one shot..

jleclanche:
Awesome! Yes, that sounds great.

pennersr:
The pluggable regex is in, see: 4e534e0e1f39753bc3001c1d3427757d084db90f

pennersr:
I am still pondering about the regex now that it is in. Basically, it does not make much sense to have a setting for that, as under normal circumstances people that have special requirements should have setup a custom user model anyway. There, one is free to setup whatever username validator.

The problem I see is that if you ignore the above and use a configurable validator that is solely used by allauth, then you will run into trouble saving your user instance from the Django admin later on.

jleclanche:
@pennersr Just let me know how you'd like to see it laid out and I'll push a commit out today/tomorrow.

pennersr:
Are you using a custom user model? If so, then I recommend you to put the proper username validator in there.

pennersr:
As for this PR, I would:

- Undo the username_regex changes from the social adapter

- Add a BattlenetUsernameValidator

- Add a note in the docs that this validator should be used by setting up a custom user model using: `username = models.CharField(validators=[BattlenetUsernameValidator(),..])` 

- Undo the `ret` related changes (sorry for being pedantic here)

- See the `get_callback_url` related discussion above.

jleclanche:
@pennersr PR is ready for a new review.

By the way, if you are unhappy with the last commit, the first 6 can land without it. In fact, all the commits are independent from one another except for "Allow overriding the region in GET parameters" which requires `get_callback_url` and `valid_regions`.

pennersr:
Thanks, LGTM

pennersr:
... except for the pep8 error:

allauth/socialaccount/providers/battlenet/views.py:99:80: E501 line too long (80 > 79 characters)

jleclanche:
@pennersr Fixed, but you really should change that - 80 characters is an unnecessarily low limit and it's making the code look bad in some places.

Also if flake8 issues are blocking, it should be set up on travis. I can submit a PR for that if you like.

pennersr:
Thanks for the fix... (please don't file a PR on coding style... like pep8 as it is).

jleclanche:
@pennersr I meant a PR to get flake8 set up on travis. If you're blocking PRs on it, it should be done automatically not at review time.

pennersr:
That is already the case... the travis build did fail:

```
script:
  - flake8 --exclude=migrations allauth
  - coverage run manage.py test allauth
```

Issue #1375, opened 2016-05-17
KeyError at /accounts/weibo/login/callback/


Tanych:
It works well when signing up with the Weibo app management username and password. However, when I try to sign up using others Weibo accounts and get the authorization from the apps. It will raise KeyError due to the absence of key "idstr" in data in WeiboProvider.extract_uid().

``` python
Request Method:     GET
Request URL:    http://127.0.0.1:8080/accounts/weibo/login/callback/?state=vZO8wmTjF3DP&code=c6a9d9ba4a4b3bdfd4e3393a92f5f039
Django Version:     1.8.8
Exception Type:     KeyError
Exception Value:    

'idstr'

Exception Location:     /usr/local/lib/python2.7/dist-packages/allauth/socialaccount/providers/weibo/provider.py in extract_uid, line 25
Python Executable:  /usr/bin/python
Python Version:     2.7.6
```

nujabes8:
Following view class only catches PermissionDenied, OAuth2Error
* OAuthCallbackView
* OAuth2CallbackView
However we get following error on production level when there are swarming request or undistinguishable api return from providers. 

```
HTTPError at /accounts/facebook/login/callback/

500 Server Error: Internal Server Error
```

```
HTTPError at /accounts/google/login/callback/

500 Server Error: Internal Server Error
```

```
ConnectionError at /accounts/twitter/login/callback/

HTTPSConnectionPool(host='api.twitter.com', port=443): Max retries exceeded with url: /1.1/account/verify_credentials.json?include_email=true (Caused by <class 'socket.error'>: [Errno 110] Connection timed out)

```

```
TypeError at /accounts/facebook/login/callback/

'NoneType' object has no attribute '__getitem__'
````

pennersr:
The Weibo case is closed via 2082160e915371f1ff03f91c3bb55855811398b3

Not sure what needs to be done for VK, please file a pull request for that using the approach taken for Weibo.

pennersr:
https://codeberg.org/allauth/django-allauth/issues/1375

PR #1558, opened 2016-12-10
Add exception handling for oauth error case for ceratin cases

nujabes8:
I know it is not perfect solution, 
can you at least refer to this suggestion and make exception handling fix?

Thank you.

refer to issue #1375 
----

Following view class only catches PermissionDenied, OAuth2Error
* OAuthCallbackView
* OAuth2CallbackView
However we get following error on production level when there are swarming request or undistinguishable api return from providers. 

```
HTTPError at /accounts/facebook/login/callback/

500 Server Error: Internal Server Error
```

```
HTTPError at /accounts/google/login/callback/

500 Server Error: Internal Server Error
```

```
ConnectionError at /accounts/twitter/login/callback/

HTTPSConnectionPool(host='api.twitter.com', port=443): Max retries exceeded with url: /1.1/account/verify_credentials.json?include_email=true (Caused by <class 'socket.error'>: [Errno 110] Connection timed out)

```

```
TypeError at /accounts/facebook/login/callback/

'NoneType' object has no attribute '__getitem__'
```

```
KeyError at /accounts/vk/login/callback/

'response'
Request Method:	GET
Request URL:	***/accounts/vk/login/callback/?code=f3e4cd7478618ee313&state=DIsBbTeTQwuQ
Django Version:	1.7.1
Exception Type:	KeyError
Exception Value:	
```

### DIFF
--- a/allauth/socialaccount/providers/oauth2/views.py
+++ b/allauth/socialaccount/providers/oauth2/views.py
@@ -12,7 +12,8 @@ from allauth.utils import build_absolute_uri
 from allauth.socialaccount.helpers import render_authentication_error
 from allauth.socialaccount import providers
 from allauth.socialaccount.providers.oauth2.client import (OAuth2Client,
-                                                           OAuth2Error)
+                                                           OAuth2Error,
+                                                           OAuthHTTPError)
 from allauth.socialaccount.helpers import complete_social_login
 from allauth.socialaccount.models import SocialToken, SocialLogin
 from allauth.utils import get_request_param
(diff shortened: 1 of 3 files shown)


pennersr:
See 2082160e915371f1ff03f91c3bb55855811398b3

allauth/socialaccount/providers/oauth2/views.py
```diff
@@ -133,7 +134,7 @@ def dispatch(self, request):
             else:
                 login.state = SocialLogin.unstash_state(request)
             return complete_social_login(request, login)
-        except (PermissionDenied, OAuth2Error) as e:
+        except (PermissionDenied, OAuth2Error, OAuthHTTPError, KeyError) as e:
```
nujabes8: greate thanx!

PR #1569, opened 2016-12-21
Cast uid to string.

shulcsm:
Endpoint used to return string so everything was fine. Their API is not versioned and i can't seem to find any changelog so there is that. Ping me if you want some kind of test for this.


Ameriks:
@shulcsm Thanks for fixing! I started receiving errors on 14th of December, but didn't have time to look into this.

PR #1586, opened 2017-01-10
Added GGather.com

DominikSerafin:
Added my project GGather.com which uses allauth for all the accounts needs


pennersr:
Closing -- I removed the showcase altogether

PR #1593, opened 2017-01-20
Shopify embedded auth

ivanvenosdel:
This adds auth support for Shopify apps that are configured as "embedded". 

Shopify has been recommending embedded apps over the traditional version heavily, at least within the last year or two. Embedded apps get placed within an iFrame on Shopify's end which requires that the first redirect in the auth process [originate from the client](https://help.shopify.com/api/sdks/embedded-app-sdk/getting-started#oauth). The result ends up looking like a Heroku App on the authenticator's end if you are familiar.

Anyway, this makes it possible for an allauth user's embedded-shopify-app to work as intended if they set IS_EMBEDDED to True (off by default) within the provider settings.

Thanks goes to the [django-shopify-auth](https://github.com/discolabs/django-shopify-auth) project for some inspiration (not code).

### Out of scope 
There is more development that a Shopify app creator has to do in order to have their page work as an iFrame within shopify (building the x_frame_exempt landing page and handing session expiration for instance). I didn't document those details because they seemed outside the scope of Django Allauth's purpose.


allauth/socialaccount/providers/shopify/views.py
```diff
@@ -54,5 +58,31 @@ def complete_login(self, request, app, token, **kwargs):
             request, extra_data)
 
 
-oauth2_login = OAuth2LoginView.adapter_view(ShopifyOAuth2Adapter)
+class ShopifyOauth2LoginView(OAuth2LoginView):
```
pennersr: Nitpick: Oauth->OAuth

allauth/socialaccount/providers/shopify/views.py
```diff
@@ -54,5 +58,31 @@ def complete_login(self, request, app, token, **kwargs):
...
+            Shopify embedded apps (that run within an iFrame) require a JS
+            (not server) redirect for starting the oauth2 process.
+
+            See Also:
+            https://help.shopify.com/api/sdks/embedded-app-sdk/getting-started#oauth
+            """
+            js = ''.join((
+                '<script type="text/javascript">',
+                'window.top.location.href = "{url}";'.format(url=response.url),
+                '</script>'
```
pennersr: Shouldn't this become proper HTML, with doctype, HTML/HEAD tags et al ?

ivanvenosdel: No reason not to. I will add that.

allauth/socialaccount/providers/shopify/views.py
ivanvenosdel: Np, consistency matters.

ivanvenosdel:
@pennersr Should be all set now.

PR #1594, opened 2017-01-28
[Docs] Add note to Bitbucket provider (Key vs. client_id)

bittner:
- Fix indentation issue with bulleted list in Advanced chapter
- Add note to Bitbucket about Key vs. client_id in Providers chapter
- Use `code-block` directive for Python code samples
- Reformat code samples (PEP8) and 80-character limit
- Fix markup violations in subtitles, use uniform vertical spacing


pennersr:
Thanks for the cleanup...

PR #1598, opened 2017-02-03
Make Slack provider work with only the 'identify' scope.

morgante:
This change reduces the number of scopes which the Slack authentication provider depends on. It's important to do this because otherwise the authentication flow will not work if developers want to use any non-identity scopes, due to Slack's restrictions.


pennersr:
Thanks.. only one nitpick, as this is causing the build to fail:

allauth/socialaccount/providers/slack/views.py:10:1: E302 expected 2 blank lines, found 1

pennersr:
Closed via 2e41b11

PR #1606, opened 2017-02-08
Added support for requesting extra_data from OpenId servers

sedlar:
This commit adds possibility to specify data you want to request from openid provider. The returned data will be stored in extra_data attribute of socialaccount.


allauth/socialaccount/providers/openid/provider.py
```diff
@@ -55,9 +55,22 @@ def get_brands(self):
...
+            if endpoint == server.get('openid_url'):
+                return server
         return {}
 
+    def extract_extra_data(self, response):
+        extra_data = {}
+        server_settings = \
+            self.get_server_settings(response.endpoint.server_url)
+        extra_attributes = server_settings.get('extra_attributes', [])
+        for id, name, _ in extra_attributes:
```
mpavlase: "id" variable will shadow with built in function

PR #1609, opened 2017-02-10
[Docs] Add scope configuration details for GitHub

bittner:
Configuring the GitHub OAuth correctly in order to be allowed to do actual things via GitHub v3 API  [is obviously a challenge](http://stackoverflow.com/questions/6802516/github-api-create-repo/39727575). This addition to the docs should remediate this problem for django-allauth users.

### DIFF
--- a/docs/providers.rst
+++ b/docs/providers.rst
@@ -419,6 +419,21 @@ App registration (get your key and secret here)
 Development callback URL
     http://127.0.0.1:8000/accounts/github/login/callback/
 
+If you want more than just read-only access to public data specify the scope
+as follows. See https://developer.github.com/v3/oauth/#scopes for details.
+
+.. code-block:: python
+
+    SOCIALACCOUNT_PROVIDERS = {
+        'github': {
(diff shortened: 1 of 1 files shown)


bittner:
Is there anything I can do in addition to have this PR merged?

bittner:
Thank for merging! :+1: 

I've noticed that the [latest docs](https://django-allauth.readthedocs.io/en/latest/providers.html#github) are not being built automatically from `master`. Should a trigger for that towards RTD maybe be added to the Travis configuration? Something like this:

``` YAML
notifications:
  webhooks:
    urls:
      - https://readthedocs.org/build/django-allauth
    on_start: never
    on_failure: never
    on_success: always
```

bittner:
@pennersr Is [my comment on generating the docs](https://github.com/pennersr/django-allauth/pull/1609#issuecomment-279645264) relevant? Should I create a PR?

pennersr:
Yes, would be nice...

PR #1610, opened 2017-02-11
Order of fields in SignupForm

polarise:
When a developer needs to have email and email (confirm) the second email appears after passwords. This is because email (confirm) is declared in BaseSignupForm's ``__init__`` method while passwords in SignupForm are class methods. The solution is to place password declarations in SignupForm's ``__init__`` method.

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -59,8 +59,8 @@ instances are created, and populated with data
 - `allauth.account.adapter.DefaultAccountAdapter`:
 
   - `is_open_for_signup(self, request)`: The default function
-  returns `True`. You can override this method by returning `False`
-  if you want to disable account signup.
+    returns `True`. You can override this method by returning `False`
+    if you want to disable account signup.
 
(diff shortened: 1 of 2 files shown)


allauth/account/forms.py
```diff
@@ -327,12 +327,10 @@ def custom_signup(self, request, user):
...
-
-    password1 = PasswordField(label=_("Password"))
-    password2 = PasswordField(label=_("Password (again)"))
-
     def __init__(self, *args, **kwargs):
         super(SignupForm, self).__init__(*args, **kwargs)
+        self.fields['password1'] = PasswordField(label=_("Password"))
+        self.fields['password2'] = PasswordField(label=_("Password (again)"))
         if not app_settings.SIGNUP_PASSWORD_ENTER_TWICE:
             del self.fields["password2"]
```
pennersr: It makes little sense to first add the field only to remove it... I would turn this `if not` into an `if` and add the field in there.

pennersr:
Closed via 882030ced0584457448bdcf548a6c482fd66a827

PR #1615, opened 2017-02-16
Drop Django 1.4/Python 2.6 compatibility hacks

jleclanche:
Also, @pennersr any thoughts on dropping Django 1.7 compatibility? It's no longer supported upstream, and Django 1.9 support ends in April.


pennersr:
I am 👍 with dropping 1.7...

jleclanche:
Sounds good.

PR #1621, opened 2017-02-19
Fix local development without https.

btimby:
https is forced even for local development. A change was made about a year ago that seems to have introduced this behavior.

https://github.com/pennersr/django-allauth/commit/4032ad85e517c3e38b1ad809375554d61f78bee6

### DIFF
--- a/allauth/socialaccount/providers/amazon/views.py
+++ b/allauth/socialaccount/providers/amazon/views.py
@@ -15,7 +15,7 @@ class AmazonOAuth2Adapter(OAuth2Adapter):
     authorize_url = 'http://www.amazon.com/ap/oa'
     profile_url = 'https://www.amazon.com/ap/user/profile'
     supports_state = False
-    redirect_uri_protocol = 'https'
+    redirect_uri_protocol = None
 
     def complete_login(self, request, app, token, **kwargs):
         response = requests.get(
--- a/allauth/socialaccount/providers/dropbox_oauth2/views.py
+++ b/allauth/socialaccount/providers/dropbox_oauth2/views.py
@@ -14,7 +14,7 @@ class DropboxOAuth2Adapter(OAuth2Adapter):
     access_token_url = 'https://api.dropbox.com/1/oauth2/token'
     authorize_url = 'https://www.dropbox.com/1/oauth2/authorize'
     profile_url = 'https://api.dropbox.com/1/account/info'
-    redirect_uri_protocol = 'https'
+    redirect_uri_protocol = None
 
     def complete_login(self, request, app, token, **kwargs):
         extra_data = requests.get(self.profile_url, params={


btimby:
Even though the documentation states that Amazon requires https for callback URLs, this is apparently not true. I was able to register successfully with the second commit using http://localhost:8000/.

http://django-allauth.readthedocs.io/en/latest/providers.html#amazon

btimby:
If you are happy with the above changes, I will gladly modify the documentation, but since this is my first time using django-allauth, I will await your response before proceeding (in case I am mistaken or missing something).

btimby:
You know better than I. Two things I noticed:

1. The history shows this logic has flip-flopped a few times, the code might be "too clever" trying to figure it out for the user.
2. It seems to me that if the hostname is "localhost" you **ALWAYS** want http.

FWIW, I had the following in settings.py:

```python
if DEBUG:
    ACCOUNT_DEFAULT_HTTP_PROTOCOL = 'http'
```

And I was very confused why this did not do what I expected (force http) before I went digging into the code.

jleclanche:
> It seems to me that if the hostname is "localhost" you ALWAYS want http.

This is untrue for several providers (eg. Battle.net always wants https even for localhost).

Why not just set it to `getattr(settings, "ACCOUNT_DEFAULT_HTTP_PROTOCOL", "https")`?

spool:
I used [django-extensions](http://django-extensions.readthedocs.io/en/latest/runserver_plus.html?highlight=https#ssl) to [deal with this for MailChimp](http://django-allauth.readthedocs.io/en/latest/providers.html#testing-locally), though that would imply a dependency.

btimby:
Cool I did not know about that!

spool:
That would be a broad solution though a bit clunky to enforce on all users so if you like that solution we could make that an optional dependency.

btimby:
Documenting the django-extensions approach would probably be sufficient.

The heart of my issue is that the software did not respect the configuration I gave it. I asked it to use HTTP, and it refused to do so. Whether HTTPS is supported by X or Y provider is not really relevant. To me the only relevant fact is that the software chose to ignore what I asked it to do. It thought it was smarter than me (it is not, Amazon does not require https for testing). So in the end, I really did want what I was asking for. I knew the consequences, I could not get what I wanted without changing the code, even though there was an option purporting to do what I wanted.

IIOW, defaults should be "smart" but options should override this (un)intelligence, since we are all consenting adults.

ezhlobo:
I see this PR is very old however it's still a point. This lib does a bit more than promised in documentation. For example dropbox allows redirects to HTTP only for localhost, but with current implementation it's not possible to have.

There is a good parameter `ACCOUNT_DEFAULT_HTTP_PROTOCOL` but it's not taken into account for dropbox.

pennersr:
Closed via 44f4561b

PR #1624, opened 2017-02-19
Raise exception if Douban API return invalid data

guoqiao:
1. Douban profile API may return invalid data. Raise ProviderException for django-allauth to catch and display.
2. Return right data in extract_common_fields


pennersr:
The build is failing on the import order check (run make isort-fix)

pennersr:
Closed via e23a540e084d8f7159de4539dfec0db5f86d9295

PR #1625, opened 2017-02-20
Drop support for Django 1.7, Python 3.2; remove misc. compat hacks

jleclanche:


jleclanche:
@pennersr can I suggest not squashing loosely-related commits like this? Having clear and separate commits really helps when doing git blame / figuring out why a line changed / etc

pennersr:
Point taken -- this PR indeed did not require any squashing...

PR #1626, opened 2017-02-20
Feature/subject in context & LOGIN_ON_GET not working

rosscdh:


allauth/account/adapter.py
```diff
@@ -105,12 +105,16 @@ def render_mail(self, template_prefix, email, context):
...
                                    context)
+
         # remove superfluous line breaks
         subject = " ".join(subject.splitlines()).strip()
         subject = self.format_email_subject(subject)
 
         from_email = self.get_from_email()
 
+        # update context with the subject
+        context.update({'subject': subject})
```
pennersr: The key/lower change is okay, but I am not sure we should do this, for one it will break backwards compatibility in the case people already have subject in there. Also, I am wondering why you need this, you could just as well do {% include "message_subject.txt" %} in your template...

PR #1630, opened 2017-02-23
Run tests on Django 1.11b1

jleclanche:


jleclanche:
@pennersr Thoughts on dropping support for Python 3.3 altogether? Django 1.8 supports 3.4.

jleclanche:
Looks like one of the tests is failing on 1.11 :)

Issue #1632, opened 2017-02-24
Allow for easier subclassing of providers


lukeburden:
Hi there,

I'd like to propose a change and will follow approval/agreement here with a PR.

**The problem**

I've recently had to subclass Google, WindowsLive and PayPal providers to achieve a range of effects:
- OAuth2 flow without creation of a SocialAccount
- Overriding non-empty default scope
- Mark subclasses as not usable for login purposes, but OK for connection

My initial approach was to subclass the provided provider implementations to keep things DRY in terms of URLs and custom behaviour. This standard approach proves impossible due to the registration call for providers appearing in the same file as the provider class definitions.

Ie, if I define a customer provider as below:

```
from allauth.socialaccount.providers.google.provider import GoogleProvider
from wherever.mixins import NonLoginProviderMixin

class GoogleNonLoginProvider(NonLoginProviderMixin, GoogleProvider):
    id = 'google_non_login'

```

The import of GoogleProvider will result in that same class being registered, which is clearly not the intention. There is no way to subclass GoogleProvider without having it be unintentionally registered.

**My proposal:**

In each provider definition, expose the defined provider classes such that in the [load](https://github.com/pennersr/django-allauth/blob/master/allauth/socialaccount/providers/__init__.py#L31) method the registration can be performed. For example:

```
from allauth.socialaccount.providers.google.provider import GoogleProvider
from wherever.mixins import NonLoginProviderMixin

class GoogleNonLoginProvider(NonLoginProviderMixin, GoogleProvider):
    id = 'google_non_login'

provider_classes = [GoogleNonLoginProvider]

```

Then in allauth/socialaccount/providers/__init__.py:def load:

```
for app in settings.INSTALLED_APPS:
        provider_module = app + '.provider'
        try:
            for cls in importlib.import_module(provider_module).provider_classes:
                self.register(cls)
        except ImportError:
            pass
    self.loaded = True
```

This change would allow us to properly subclass providers. Let me know what you think!

pennersr:
This is indeed a paint point, I have been wanting to refactor all of this completely, perhaps even to go as far as to have just one Django app per protocol (e.g. OAuth, OAuth2, OpenID) and make things easier to subclass... but I am currently lacking the time to do all of that. Your proposal seems like a fair non-intrusive change that can easily be done without breaking too much, so :+1: for me

pennersr:
https://codeberg.org/allauth/django-allauth/issues/1632

PR #1634, opened 2017-02-25
Change provider registration to allow subclassing.

lukeburden:
Due to providers being registered in the same file as their definition
it is impossible to subclass a provider without having the parent be
registered. This change addresses this allowing for easy subclassing
of providers.

See #1632 for more info.


lukeburden:
@pennersr is this ever so slight reduction in coverage a problem?

pennersr:
Coverage is not a problem, it is hard failing on pep8/isort issues... Run make isort-fix

lukeburden:
Coverage is the only problem I can see. I ran `make isort-fix`, it only adjusted some whitespace that was not related to my changes. Where are you seeing pep8 failures?

lukeburden:
Coverage has only decreased as I've removed some lines of redundant but covered code, resulting in the ratio of covered to uncovered lines decreasing in a few files.

pennersr:
I meant that coverage is not a problem in the sense that it breaks the build -- it is only a soft-failing indicator.

pennersr:
Closes via 4039ab8e581d7e09ad354723d6dbe1c5d606db51

bittner:
@lukeburden I've seen your note on subclassing in the [ChangeLog](https://github.com/pennersr/django-allauth/commit/4039ab8e581d7e09ad354723d6dbe1c5d606db51#diff-a136a3ab0403c35496be1955335be9d9R4).

I think it would be good to add a section on "Subclassing Providers" to the [Advanced Usage](https://django-allauth.readthedocs.io/en/latest/advanced.html) chapter of the docs, explaining the approach with appropriate code snippets.

I feel, especially the introduction of `provider_classes` deserves an explanation, because - with all the advantages of the changes - the code has now stopped to be self-explanatory.
``` python
provider_classes = [SomeProvider]
```
used to be:
``` python
providers.registry.register(SomeProvider)
```

lukeburden:
@bittner Thanks for your comment. We could rename `provider_classes` to something a little more explicit like `discoverable_provider_classes`, potentially. It strikes me that anyone subclassing a provider would likely be:

1. basing their approach off an existing provider definition, in which case they'll see this convention
2. be familiar with the loading logic and so understand the convention

Changing the name would require a little more logic in the load method to still support the present naming; this doesn't seem worth it to me but will let @pennersr voice his thoughts.

On the special documentation for subclassing a provider; good idea, I'll do that.

PR #1644, opened 2017-02-28
[Docs:Showcase] Remove discontinued sites, add Painless, apply sort order

bittner:
Some showcases are either not using allauth anymore, or the sites have been discontinued.

- officecheese.com - _discontinued  (redirects to agency website)_
- superreceptionist.in - _uses python social-auth_
- kwatsi.com - _now a static website, no signup / no sign-in_
- smartgoalapp.com - _discontinued (domain parked)_
- healthifyme.com - _have stopped using allauth, site now made using [webflow](http://www.webflow.com)_

**NOTE:** Some other URLs of the list time out when trying to access them, but they cannot be confirmed to be offline or discontinued.

PR Details
------------
1. Remove discontinued sites, scams, advertisement
1. Add Painless Cloud
1. Apply alphabetical sort order (names first, then urls)

### DIFF
--- a/docs/showcase.rst
+++ b/docs/showcase.rst
@@ -5,22 +5,18 @@ Here are some sites that have `django-allauth` up and running:
 
 - `Drakdoo: Bitcoin trading app <http://www.drakdoo.com>`_
 - `Flatmate Rooms: UK flatshare <https://www.flatmaterooms.co.uk>`_
-- http://www.q-dance.com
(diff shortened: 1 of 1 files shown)


pennersr:
We independenty had the same thought about the issues in the showcase... I have removed that section now, it used to be a form of promotion that these days is not really needed anymore

PR #1648, opened 2017-03-03
handle validators being a list

alexstacey:
This fixes a bug when setting USERNAME_VALIDATORS.

USERNAME_VALIDATORS is a list of paths which fails when passed to import_attribute() - so instead pass each path in the list to import_attribute().

### DIFF
--- a/allauth/account/app_settings.py
+++ b/allauth/account/app_settings.py
@@ -296,10 +296,11 @@ class AppSettings(object):
         from django.core.exceptions import ImproperlyConfigured
         from allauth.utils import import_attribute
         from allauth.utils import get_user_model
-
-        path = self._setting('USERNAME_VALIDATORS', None)
-        if path:
-            ret = import_attribute(path)
+        ret = []
+        validators = self._setting('USERNAME_VALIDATORS', None)
+        if validators:
+            for path in validators:
+                ret.append(import_attribute(path))
             if not isinstance(ret, list):
                 raise ImproperlyConfigured(
                     'ACCOUNT_USERNAME_VALIDATORS is expected to be a list')
@@ -307,8 +308,6 @@ class AppSettings(object):
             if self.USER_MODEL_USERNAME_FIELD is not None:
                 ret = get_user_model()._meta.get_field(
                     self.USER_MODEL_USERNAME_FIELD).validators
-            else:
-                ret = []
         return ret
 
 


pennersr:
For now I have reverted this, according to the current documentation/implementation it was a path to a list, and not directly a list itself. I do agree, that may be unexpected, we need to reassess this, but as is now changing this would break backwards compatibility.

biddellns:
I referenced this in #1658. What's the best way to implement this? You said it's a path to a list, but where should that list reside, if not in the settings module?

biddellns:
I closed my other issue to avoid a lack of redundancy, but here are the contents:
------------------------------------------------------------------------------------------

I'm trying to use the Battlenet BattletagUsernameValidator However, Django is throwing errors no matter which way I write it.

If my settings.py contains:

`ACCOUNT_USERNAME_VALIDATORS = "socialaccount.some.validator"`

I get an error that ACCOUNT_USERNAME_VALIDATORS must be a list.

If I amend this I get an error about django six.string_types.

Here is the error I'm currently getting:

```
Internal Server Error: /accounts/battlenet/login/callback/
Traceback (most recent call last):
  File "/home/app_user/.virtualenvs/litsl-staging/lib/python3.4/site-packages/django/core/handlers/exception.py", line 39, in inner
    response = get_response(request)
  File "/home/app_user/.virtualenvs/litsl-staging/lib/python3.4/site-packages/django/core/handlers/base.py", line 249, in _legacy_get_response
    response = self._get_response(request)
  File "/home/app_user/.virtualenvs/litsl-staging/lib/python3.4/site-packages/django/core/handlers/base.py", line 187, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/home/app_user/.virtualenvs/litsl-staging/lib/python3.4/site-packages/django/core/handlers/base.py", line 185, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/app_user/.virtualenvs/litsl-staging/lib/python3.4/site-packages/allauth/socialaccount/providers/oauth2/views.py", line 69, in view
    return self.dispatch(request, *args, **kwargs)
  File "/home/app_user/.virtualenvs/litsl-staging/lib/python3.4/site-packages/allauth/socialaccount/providers/oauth2/views.py", line 139, in dispatch
    return complete_social_login(request, login)
  File "/home/app_user/.virtualenvs/litsl-staging/lib/python3.4/site-packages/allauth/socialaccount/helpers.py", line 153, in complete_social_login
    return _complete_social_login(request, sociallogin)
  File "/home/app_user/.virtualenvs/litsl-staging/lib/python3.4/site-packages/allauth/socialaccount/helpers.py", line 169, in _complete_social_login
    ret = _process_signup(request, sociallogin)
  File "/home/app_user/.virtualenvs/litsl-staging/lib/python3.4/site-packages/allauth/socialaccount/helpers.py", line 36, in _process_signup
    get_account_adapter(request).clean_username(username)
  File "/home/app_user/.virtualenvs/litsl-staging/lib/python3.4/site-packages/allauth/account/adapter.py", line 247, in clean_username
    for validator in app_settings.USERNAME_VALIDATORS:
  File "/home/app_user/.virtualenvs/litsl-staging/lib/python3.4/site-packages/allauth/account/app_settings.py", line 313, in USERNAME_VALIDATORS
    ret = import_attribute(path)
  File "/home/app_user/.virtualenvs/litsl-staging/lib/python3.4/site-packages/allauth/utils.py", line 142, in import_attribute
    assert isinstance(path, six.string_types)
AssertionError
```

My settings.py looks like:

```
ACCOUNT_USERNAME_VALIDATORS = [ 'socialaccount.providers.battlenet.validators.BattletagUsernameValidator',
 ]
```

alexstacey:
@biddellns I think settings needs a path and the path has to point to a list. So like:

foobar/settings.py
```
ACCOUNT_USERNAME_VALIDATORS = 'foobar.validators.username_validators'
```

then create foobar/validators.py with:
```
def username_validators():
    return [
        socialaccount.providers.battlenet.validators.BattletagUsernameValidator,
    ]
```

biddellns:
Ahh that makes sense! I know that's a really basic solution, but I was stuck on that problem for a while. Thank you for the guidance!

biddellns:
I'm actually still having trouble tracking this down and I'm not sure why python is being so weird about it. 

Django was still complaining when I used a method to return a list. So I just hardcoded a list instead and it was able to pick that up.

Unfortunately, it still won't pick up the validator. I'm getting this error and I've double and triple checked

```
  File "/home/app-user/sites/litsl_com/source/foobar/settings/validators.py", line 8, in <module>
    vs = [socialaccount.providers.battlenet.validators.BattletagUsernameValidator,]
AttributeError: 'module' object has no attribute 'validators'
```

My validators file looks like:
```
from allauth import socialaccount

# Doesn't work with all_auth
def username_validators():
    return [
            'socialaccount.providers.battlenet.validators.BattletagUsernameValidator',
           ]
# Is picked up by allauth but can't find validators
vs = [socialaccount.providers.battlenet.validators.BattletagUsernameValidator,]
```

My app structure looks like
```
myapp -
   - settings
        - base.py
        - staging.py
        -validators.py
```
It's definitely picking up my validators file. And I verified the 'socialaccount/providers/battlenet/validators' path.

Any ideas?

alexstacey:
yes... The reason the top one isn't working is that you're returning a list with a string in it rather than the actual validator object. I think you want something like:
```
from allauth.socialaccount.providers.battlenet import validators

def username_validators():
    return [
            validators.BattletagUsernameValidator,
           ]
```

biddellns:
I finally got it working! I had to write my import the validators as you described. However returning the function still didn't work. For anyone looking, I used this in 'myapp/settings/validators.py':

```
from allauth.socialaccount.providers.battlenet import validators

validator_list = [ validators.BattletagUsernameValidator, ]
```

xianfuxing:
@biddellns great! working fine! Thanks.

PR #1652, opened 2017-03-05
Expand Eventbrite provider implementation and correct test case

spool:
Just a little tweak removing an unneeded section. Tested it still works.


spool:
Hey, by (biased) guess is that the decrease in coverage is a function of reducing lines of code rather than making the tests worse.

allauth/socialaccount/providers/eventbrite/provider.py
```diff
@@ -39,5 +47,16 @@ def extract_common_fields(self, data):
             image_url=data.get('image_url')
         )
 
+    def extract_email_addresses(self, data):
+        """Extract email addresses if verified."""
+        ret = []
+        emails = data.get('emails')
+        for email_account in emails:
+            if email_account['verified']:
```
pennersr: The email address can/should be added even if it is not verified (but then ofcourse with verified=False)

allauth/socialaccount/providers/eventbrite/provider.py
```diff
@@ -11,6 +15,14 @@ def get_avatar_url(self):
         """Return avatar url."""
         return self.account.extra_data['image_id']
 
+    def to_str(self):
+        """Return string of account name."""
+        return self.account.extra_data['name']
+
+    def get_profile_url(self):
```
pennersr: You introduced this method but it seems not to be used?

spool: Ah sorry, I just saw that implemented in a number of other places and wondered if that was helpful. Will remove if not useful.

allauth/socialaccount/providers/eventbrite/provider.py
spool: Ah ok, sorry this was following another example (google's I think). Much prefer to keep all. Will fix.

allauth/socialaccount/providers/eventbrite/provider.py
pennersr: You can keep it, but then you should return the end-user URL linking to the user profile. Now you are returning the API URL endpoint.

spool:
Just copied the Google version, following the docstring on the super class.

allauth/socialaccount/providers/eventbrite/provider.py
spool: Just a nudge but apologies for not replying here initially: I think I've fixed this.

pennersr:
>  You can keep it, but then you should return the end-user URL linking to the user profile. Now you are returning the API URL endpoint.

Have you tackled the above issue? You are returning https://www.eventbriteapi.com/v3/users/me which does not seem like an end-user facing endpoint

spool:
Hey @pennersr thanks for the reply. I think I misunderstood that change request initially and now I think I do, thanks for clarifying. I'm a bit confused as to what to do next though, so advice would be appreciated. I think I got too confident when I could simply login with the current implementation using a test account, and pass the test with decent coverage.

https://www.eventbrite.co.uk/developer/v3/endpoints/users/ is the documentation I followed in choosing the `v3/users/me` endpoint, which refers to which user has logged in. Assuming they accept access to their account via the implemented `allauth` redirect, that should be the endpoint for their profile. The other option is `v3/users/:id/` which I could use if you prefer, and that may be a better option as once they have accepted access to their account that url will be persistently correct even if they haven't logged into Eventbrite. `v3/users/me` only works while they are logged into Eventbrite.

Unfortunately there's a further complication: successful querying of data like their `image_id` requires a token in the URL 
```python
'https://www.eventbriteapi.com/v3/users/me?token={0}'.format(token.token)
``` 
or in the request header 
```python
headers = {'Authorization': 'Bearer {0}'.format(token.token)} # django-allauth token object
requests.get('https://www.eventbriteapi.com/v3/users/me', headers=headers)
```

That's in my implementation of the `complete_login` method, but I haven't found a straightforward means of using tokens in the the `ProviderAccount` subclass, especially as tokens are not included in the `account.extra_data`.

I'm wondering about expanding this to include the token outside the `complete_login` method. I've also realised that the avatar implementation is only returning the `image_id` rather than URL so I'm trying to fix that too, but ideally that would also include the token in the URL. Or: is there a way of enforcing headers on the `get_avatar_url` and `get_profile_url` methods?

Sorry for the long message, hope that makes some sense. Will be pushing a new commit with a possible fix. Thanks.

PR #1662, opened 2017-03-08
Document provided URL names, and compare with django.contrib.auth.urls

Flimm:
This includes two commits:

- It documents in `installation.rst` whether `django.contrib.auth` is required, and whether the URLs provided by that package are required. It mentions the URL names provided by allauth
- It documents in `views.rst` the URL names of the views.


Flimm:
Thank you!

PR #1663, opened 2017-03-08
Document that the templates are overridable

Flimm:
I thought it would be useful to include this, for beginners to Django who aren't expecting to have to look in the source code of a package.


Flimm:
Thank you!

PR #1666, opened 2017-03-09
url(r'...')

JunYeongJang:
Avoiding regular expression escaping


pennersr:
Note that there are no special characters in the URL patterns, so there is nothing to escape really... but I am okay with picking this up. The PR is failing on pep8 issues though, can you have a look?

JunYeongJang:
fix pep8!

JunYeongJang:
hmm? what!

PR #1668, opened 2017-03-09
Add sociallogin and action to account connected message context

lukeburden:
I need to give users a little more precise feedback on what has happened when they've gone through a social auth flow, not just "The social account has been connected." regardless of the social account type, or whether the account previously existed or not.

There's no easy way to customize messages on social account connection or update at present. I could throw an ImmediateHttpResponse, but throwing exceptions from signal handling code to change the controlling code's behaviour is very undesirable.

The adapter classes have a `message_context` parameter available on their `add_message` method, but this is seldom used.

This PR simply injects the action and social login into the message context such that I can customize the message right where it should be customized; in the template.

Let me know what you think!

### DIFF
--- a/allauth/socialaccount/helpers.py
+++ b/allauth/socialaccount/helpers.py
@@ -100,6 +100,7 @@ def _add_social_account(request, sociallogin):
         return HttpResponseRedirect(reverse('socialaccount_connections'))
     level = messages.INFO
     message = 'socialaccount/messages/account_connected.txt'
+    action = None
     if sociallogin.is_existing:
         if sociallogin.user != request.user:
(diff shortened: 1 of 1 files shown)


pennersr:
Could you have a look at the merge conflict in this branch? Thanks

lukeburden:
Sure, done - PR was affected by your merge of my other PR on the same section of code. :)

lukeburden:
By the way, if you have any better alternatives or methods used elsewhere in this project for injecting "facts" like "added" or "updated" into the message context, let me know, happy to adjust. Don't think we need anything too fancy though.

PR #1669, opened 2017-03-10
UX improvement: Remove first mention of domain name in emails

Flimm:
Based on my limited UX testing, it seems that users don't read the confirmation and password reset emails. No surprise there. They just try to click the first link they see.

Because the first link they see is the domain name, I have seen them try to click that link, only to be confused that the email confirmation apparently did not work.

This commit makes the first (and biggest) link the confirmation link. Later on in the email, the site domain is displayed.

I realise that developers can override these templates, but why not make the UX a little bit nicer by default?


Flimm:
Thank you!

PR #1674, opened 2017-03-16
fixed #1673

### LINKED ISSUE #1673: Typo in flash message after incorrect login.
The `LoginForm` Class of `account/forms.py` ([link](https://github.com/pennersr/django-allauth/blob/master/allauth/account/forms.py#L69)) has `error_messages` for different login error situations.

`username_email_password_mismatch` key of `error_message` has value `"The login and/or password you specified are not correct."`. I think I should explicitly say that "The login username/email and/or password you specified are not correct".

Open for reviews.

rishabhiitbhu:
Addressed Issue #1673 in this PR

### DIFF
--- a/allauth/account/forms.py
+++ b/allauth/account/forms.py
@@ -163,8 +163,8 @@ class LoginForm(forms.Form):
             else:
                 raise forms.ValidationError(
                     self.error_messages[
-                    '%s_password_mismatch'
-                    % app_settings.AUTHENTICATION_METHOD])
+                        '%s_password_mismatch'
+                        % app_settings.AUTHENTICATION_METHOD])
         return self.cleaned_data
 
     def login(self, request, redirect_url=None):
--- a/allauth/account/forms.py
+++ b/allauth/account/forms.py
@@ -161,7 +161,7 @@ class LoginForm(forms.Form):
                 raise forms.ValidationError(
                     self.error_messages[error_type])
             else:
-                self.ValidationError(error_messages[
+                forms.ValidationError(error_messages[
                     '%s_password_mismatch'
                     % app_settings.AUTHENTICATION_METHOD])
         return self.cleaned_data
--- a/allauth/account/forms.py
+++ b/allauth/account/forms.py
@@ -82,9 +82,6 @@ class LoginForm(forms.Form):
 
         'username_password_mismatch':
         _("The username and/or password you specified are not correct."),
-
-        'username_email_password_mismatch':
-        _("The username/e-mail and/or password you specified are not correct.")
     }
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


pennersr:
Hmmm, sorry for coming back to this, but now that I see the full sentence:

    "The username/e-mail and/or password..."

... I find that it looks a bit weird, as it basically translates to this:

    "The username **or** e-mail **and/or** password..."

So all in all I think it would be better to drop the `'username_email_password_mismatch'` altogether. In case of authentication method `username_email`, we can simply check if there is a `@` in the `login` field... and if there is, render the existing `'email_password_mismatch'` message, else the `'username_password_mismatch'` one. Then we're truly explicit & user friendly (and as a bonus, one less string to translate).

rishabhiitbhu:
Okay, got it. That would indeed be truly explicit and user friendly.

rishabhiitbhu:
@pennersr I have made the changes, I would like to ask how should I run tests on my local dev environment because I need to add/modify some test methods to assert the changes.

rishabhiitbhu:
@pennersr 
Travis CI build gives "No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself." and Coverage/coveralls has also failed. I have tried to do my best, still I'm facing problems. Please help.

allauth/account/forms.py
```diff
@@ -155,10 +152,19 @@ def clean(self):
         if user:
             self.user = user
         else:
-            raise forms.ValidationError(
-                self.error_messages[
-                    '%s_password_mismatch'
-                    % app_settings.AUTHENTICATION_METHOD])
+            if app_settings.AUTHENTICATION_METHOD == 'username_email':
```
pennersr: I think you can shorten this down a bit, how about something like this:

```
auth_method = app_settings.AUTHENTICATION_METHOD
if auth_method == app_settings.AuthenticationMethod.USERNAME_EMAIL:
    login = self.cleaned_data['login']
    if '@' in login and '.' in login:
        auth_method = app_settings.AuthenticationMethod.EMAIL
    else:
        auth_method = app_settings.AuthenticationMethod.USERNAME
raise forms.ValidationError(
    self.error_messages['%s_password_mismatch' % auth_method])
```

rishabhiitbhu: Awesome!! Will do so.

pennersr:
@rishabhiitbhu To test locally, simply `pip install tox`... and then run just `tox` from the root repo directory.

You can make sure that there are no code style issues by running `make pep8` and `make isort-check` (requires `pip install isort` and `pip install flake8`)

rishabhiitbhu:
@pennersr I applied your suggested changes. The Travis CI build has successfully passed. But coverage has decreased. Do I need to write custom tests to assert the chages? Please help.

rishabhiitbhu:
Hey! @pennersr, we allow "@" and "." in usernames [doc link](https://docs.djangoproject.com/en/1.10/ref/contrib/auth/#django.contrib.auth.models.User.username) if we could restrict that this would work perfectly. What do you think?

pennersr:
@rishabhiitbhu You can already limit what characters are allowed within your project (e.g. by overriding `clean_username`) ... I do not want to go as far as to restrict the characters out of the box, it would break backwards compatibility, and the worst that could happen is that a user is presented with an ambiguous error message.

PR #1680, opened 2017-03-24
Form code literals in RST docs correctly

Flimm:
Unlike Markdown, you should surround code literals with double backticks
on either side, ``like this``. This goes through all the .rst files and corrects this.


Flimm:
Thanks!

PR #1687, opened 2017-03-31
Support for Shopify per-user access mode

ivanvenosdel:
This an optional feature that allows an all-auth site to tie the actual token, email, first and last of the Shopify user to the all-auth account, rather than the shop's token and email. 

As noted in code comments before this change, the extra_data for the user can not be obtained from an API request. However it can now be obtained by utilizing 'online' as an access_mode. This causes shopify to provide the user's details when passing back the auth token (and gives a token with a ~24hr expiration).

See Also: https://help.shopify.com/api/getting-started/authentication/oauth#api-access-modes

The code is entirely based off whether or not the 'associated_user' is given to us by Shopify. Which will be true if they put in the existing AUTH_PARAM setting. I thought this more explicit then adding a setting (e.g. ONLINE_ACCESS_MODE: True/False) that sets the auth param for them.

### DIFF
--- a/allauth/socialaccount/providers/shopify/provider.py
+++ b/allauth/socialaccount/providers/shopify/provider.py
@@ -1,3 +1,5 @@
+from django.conf import settings
+
 from allauth.socialaccount.providers.base import ProviderAccount
 from allauth.socialaccount.providers.oauth2.provider import OAuth2Provider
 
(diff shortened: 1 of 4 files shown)


ivanvenosdel:
Hah! Ok coveralls, I get it, thanks.

pennersr:
Thanks!

Issue #1688, opened 2017-03-31
different SocialApp providers on different sites


velis74:
I'm having problems where my app is handling multiple sites and not all of them use the same social providers:

INSTALLED_APPS = (
    ....
    'allauth.socialaccount.providers.facebook',
    'allauth.socialaccount.providers.google',
    'allauth.socialaccount.providers.twitter',
    'django.contrib.sites',
    ....
)

When looked at socialaccount_socialapp_sites table, I have the following mappings:
site 1: facebook, google, twitter (the "intended" scenario)
site 2: facebook, google
site 3: google

I have two distinct problems / issues with this:

1. Login screen lists all three providers regardless of what's specified in the socialapps_sites table. This is because get_providers tag simply gets list of all installed providers, not providers mapped to the site. My further inquiry shows that providing context to this tag (takes_context=True) would allow request to be passed into the registry.get_list() function which could then retrieve mapping-relevant list

2. while twitter provider makes no check on mapping (and only the actual attempt at logging-in with twitter would fail), facebook provider checks for the mapping immediately, and of course fails when mapping is not present (ImproperlyConfigured, provider.py, media_js, except SocialApp.DoesNotExist). I believe this one would be "solved" simply by solving issue 1. Not tested yet.

Question:
Is my reasoning that different sites might not all use the same providers correct? Would a patch solving this issue be even considered?

I will update the issue with my findings later.

velis74:
It would seem that my solution works nicely. I only get the icons for the providers, registered for a particular site. Additionally, I don't get ImproperlyConfigured any more.

Preparing a pull request for @pennersr to evaluate.

pennersr:
In the current main branch `{% get_providers %}` respects what the sites framework.

pennersr:
https://codeberg.org/allauth/django-allauth/issues/1688

PR #1689, opened 2017-03-31
Issue #1688: different SocialApp providers on different sites

velis74:
A solution to the issue explained in #1688


velis74:
Sorry, I have a column limit set to 120 characters per line and I forgot to check it before committing. If the patch is otherwise acceptable for you, I will fix it and resubmit.

allauth/socialaccount/providers/__init__.py
```diff
@@ -12,9 +12,16 @@ def __init__(self):
 
     def get_list(self, request=None):
         self.load()
+        try:
+            from django.contrib.sites.models import Site
+            site = Site.objects.get_current(request)
+            ids = [sa.provider for sa in site.socialapp_set.all()]
```
pennersr: This will break OpenID, as OpenID does not require the presence of a `SocialApp`.

allauth/socialaccount/providers/__init__.py
```diff
@@ -12,9 +12,16 @@ def __init__(self):
...
+        except:
+            ids = [provider_cls(request).id for provider_cls in self.provider_map.values()]
```
pennersr: This basically equals `provider_map.keys()`

allauth/socialaccount/providers/__init__.py
```diff
@@ -12,9 +12,16 @@ def __init__(self):
 
     def get_list(self, request=None):
         self.load()
+        try:
+            from django.contrib.sites.models import Site
+            site = Site.objects.get_current(request)
+            ids = [sa.provider for sa in site.socialapp_set.all()]
+        except:
```
pennersr: Please avoid Pokemon exception handling and only catch the exceptions that are likely to occur.

allauth/socialaccount/providers/__init__.py
velis74: I'm not sure what I could do to fix this. Maybe check if ids list is empty?

allauth/socialaccount/providers/__init__.py
velis74: This was only intended so that in case of **any** failure the default functionality would still be performed. Will check what exceptions could be thrown here.

allauth/socialaccount/providers/__init__.py
velis74: Agreed. Too little research on my part. Sorry.

velis74:
Sorry about the delay, was on vacation. Anyway, I amended the code, hopefully it's better now. Travis still doesn't like me, but the errors displayed don't seem to be related to my code?

velis74:
Is there something wrong with the amended code?

velis74:
What is the status of this PR?

pennersr:
It breaks the build so it is not ready for merging. Also, there is a hardcoded check in there to exclude OpenID. I would rather prefer a more general mechanism, e.g. by putting a propery `requires_socialapp` on the provider base class and use that.

velis74:
OK, I'd like to help with that, but the code changes work fine on my deployments & development environment. Can you direct me to a guide on tests so that I can see why pretty much all providers fail with reverse match failure?

I also agree with the requires_socialapp proposition, I was just afraid to make such a big change - my experience tells me that project leads don't like changes to base classes, so - given your feedback - I thought it less intrusive to only exclude the openid provider.

velis74:
OK, figured out tests:
The problem with reverse links seems to be tied exactly to the code changes: tests expect the provider to be listed in ProviderRegistry.get_list, but it is not as it's not set up in database.

Change the tests as well?

pennersr:
In the current main branch {% get_providers %} respects the sites framework configuration.

PR #1693, opened 2017-04-05
Naver's extract_common_fields should be deleted.

xncbf:
The following link is Naver's official update document.

https://developers.naver.com/notice/article/10000000000030652808

Naver login will not should be given user and email since April 7th.

This applies after a 6 month grace period.

sorry for my english.


xncbf:
I can't get through the coverage check, because it was just delete code.

Issue #1701, opened 2017-04-16
Unnecessary fbsr facebook cookie


vsemionov:
After visiting the login page, the fbsr_<appID> cookie is set if the facebook provider is available. It is never cleared and sent with every subsequent request to the application. This only happens if I currently have an active facebook session.

The cookie value is quite large and significantly increases the request size. The request becomes more difficult to read. I am not sure if setting the cookie has security implications.

I was able to workaround this by setting the cookie parameter of FB.init() to false in fbconnect.js.

pennersr:
See 5916e90487fe815acc60ff744ff05bbdb05fa3d4

vsemionov:
Thanks, Raymond

pennersr:
https://codeberg.org/allauth/django-allauth/issues/1701

PR #1702, opened 2017-04-16
Issue #1701: Do not set the fbsr cookie

vsemionov:
After visiting the login page, the fbsr_<appID> cookie is set if the facebook provider is available. It is never cleared and sent with every subsequent request to the application. This only happens if I currently have an active facebook session.

The cookie value is quite large and significantly increases the request size. The request becomes more difficult to read. I am not sure if setting the cookie has security implications.

I was able to workaround this by setting the cookie parameter of FB.init() to false in fbconnect.js.

### DIFF
--- a/allauth/socialaccount/providers/facebook/static/facebook/js/fbconnect.js
+++ b/allauth/socialaccount/providers/facebook/static/facebook/js/fbconnect.js
@@ -42,7 +42,7 @@
           appId: opts.appId,
           version: opts.version,
           status: true,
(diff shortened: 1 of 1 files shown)


pennersr:
The FB.init() values may vary from project to project, so I've now made it possible to configure them. By default, they are no longer passed, meaning FB's defaults are used.

See: 521af23e15b2fdb7a5da302ffafab00e7cd82f2f

PR #1705, opened 2017-04-17
Trigger social account signals on login and update docs

lukeburden:
During the social `login` process, the user adds or updates a social account. This PR ensures the relevant signal is fired in both those cases to be consistent with the signals fired during the `connect` process.

### DIFF
--- a/allauth/socialaccount/helpers.py
+++ b/allauth/socialaccount/helpers.py
@@ -177,6 +177,12 @@ def _complete_social_login(request, sociallogin):
     if sociallogin.is_existing:
         # Login existing user
         ret = _login_social_account(request, sociallogin)
+        signals.social_account_updated.send(
+            sender=SocialLogin,
+            request=request,
+            sociallogin=sociallogin
+        )
+
     else:
         # New social user
         ret = _process_signup(request, sociallogin)
--- a/docs/signals.rst
+++ b/docs/signals.rst
@@ -72,6 +72,13 @@ allauth.socialaccount
 
   Sent after a user connects a social account to a their local account.
 
+- `allauth.socialaccount.signals.social_account_updated(request, sociallogin)`
+
+  Sent after a social account has been updated. This happens when a user
+  logs in using an already connected social account, or completes a `connect`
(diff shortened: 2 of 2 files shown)


pennersr:
The added/removed signal were setup to correspond with the higher-level user account management actions of adding/removing a new socialaccount to an already existing local account. By firing the added signal also during signup we lose that meaning (and break backwards compatibility). Instead of having dedicated signals matching the higher level business logic, this results in lower level signals that are already offered by the built-in Django CRUD signals. Therefore, I do not think it adds value to introduce this change...

lukeburden:
I see your point, especially re. backward compatibility.

My specific use case: interact with Facebook's API to pull further information when a user registers, logs in, or connects a Facebook account.

I considered using the `post_save` signal on `SocialAccount` for this purpose, but this signal firing is not guaranteed to mean that a social auth flow has just occurred, only that the model has been persisted locally.

I just strongly favour distinct, custom signals for distinct phenomena like this. I think having them helps build systems with less concern for side-effects and gaps.

I can keep this extra signal in a personal fork. Will you accept this PR if I remove the extra firing of `social_account_added` and adjust the docs?

lukeburden:
This PR would end up just adding a legitimately missing `social_account_updated` firing, on login using an existing SocialAccount, without altering the semantics of `social_account_added` and `social_account_removed`.

pennersr:
You should be able to hook up to the `user_signed_up` signal. Then, inspect whether or not a `sociallogin` kwarg is present, and if so, if `sociallogin.account.provider == 'facebook'`. I do realise that you will have to hook up your code both in the `user_signed_up` signal as well as in the `socialaccount_added` signal, but that might even be handy given that you have knowledge on what the ongoing scenario is about.

Your code comments / docs are valuable, so if you can, please update the PR with just these...

lukeburden:
Sure thing, I'll update this PR shortly. And I'll consider taking the approach you suggest, despite the odd asymmetry of the signals there. Anything is better than `post_save`! ;)

pennersr:
Closed via 712b40117f7d5dd63710e634712a8cdbb8a4badc

PR #1709, opened 2017-04-19
Backport Form.order_fields(). Fixes #1226.

### LINKED ISSUE #1226: BaseSignupForm  overrides field_order of custom signup form
If I am not mistaken, the BaseSignupForm [overrides any `field_order`](https://github.com/pennersr/django-allauth/blob/9d8ab54de5eaa12eff2964ff5c994ac3ffed1dcd/allauth/account/forms.py#L226-L246) that one could have defined in the custom base form via `ACCOUNT_SIGNUP_FORM_CLASS`.

In our use case, we used a custom form to add additional fields, however we cannot affect the order of those, which is especially annoying with the `password` fields which always appear at the end.

My proposal is that BaseSignupForm should check if `field_order` is defined and if so only add additional fields if not previously defined.

If this makes sense I would happily provide a PR with test and fix.

jnns:
This commit should fix the issue with overridden `field_order` in a custom signup form.

### DIFF
--- a/allauth/account/forms.py
+++ b/allauth/account/forms.py
@@ -260,9 +260,13 @@ class BaseSignupForm(_base_signup_form_class()):
         username_field.widget.attrs['maxlength'] = str(
             username_field.max_length)
 
-        # field order may contain additional fields from our base class,
-        # so take proper care when reordering...
-        field_order = ['email', 'username']
+        default_field_order = [
+            'email',
+            'email2',  # ignored when not present
+            'username',
+            'password1',
+            'password2'  # ignored when not present
+        ]
         if app_settings.SIGNUP_EMAIL_ENTER_TWICE:
             self.fields["email2"] = forms.EmailField(
                 label=_("E-mail (again)"),
@@ -273,8 +277,6 @@ class BaseSignupForm(_base_signup_form_class()):
                     }
                 )
             )
-            field_order = ['email', 'email2', 'username']
-        merged_field_order = list(self.fields.keys())
         if email_required:
             self.fields['email'].label = ugettext("E-mail")
             self.fields['email'].required = True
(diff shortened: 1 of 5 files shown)


jnns:
@pennersr Could you give me a short feedback about this pull request please? 

I'd like to incorporate these changes into my project but don't want to maintain a seperate fork from this repo for this.

Thanks!

pennersr:
The custom signup form was designed to be a standard form (deriving from form.Form), containing just the additional fields that need to be asked during signup. The reason for this is that there are in fact two signup forms, one for a regular account, and one for completing a social account. These are two distinct forms. For ease of extensibility, these are things you don't want/need to be bothered with. 

Given that the signup form is a regular form there is no reason it would have access to fields not declared in the form (such as email, password, ...). However, due to the particular underlying implementation these fields did leak into the custom form, allowing people to customize e.g. the field order. But, this is fragile, changes in the implementation of allauth (such as moving fields declaration into the __init__) broke the leakage. 

So, all in all, this PR needs a design decision on how to proceed on this one. This PR may correct the situation to previous behavior, but do note that that behavior was already relying on internals and was never intended to behave like this.

allauth/account/forms.py
```diff
@@ -283,22 +285,29 @@ def __init__(self, *args, **kwargs):
...
+                    'password2'  # ignored when not present
+                ]
+
         if not self.username_required:
             del self.fields["username"]
 
+        if hasattr(self, 'field_order'):
+            set_form_field_order(self, self.field_order)
+        else:
+            default_field_order = [
```
pennersr: Setting the `default_field_order` here overrides the logic above, I think this needs to be dropped?

pennersr:
Given that Django explicitly documents "Unknown field names in the list are ignored." my initial objections are not really valid, so the changes proposed here are indeed sound.

pennersr:
PR tweaked and committed as f1b5d4bd7216dad9670be22ae0c2ea29ba0fccfa

jnns:
Thank you very much for reviewing and merging this commit. And I'm sorry for not responding in time to your comment about `default_field_order`.

PR #1711, opened 2017-04-20
Fixes to EmailAddress

jleclanche:
- First commit makes the email confirmation cooldown period available in the settings, because it really should be.
- Second commit fixes a potential albeit rare crash due to the race condition between the try and except block.

### DIFF
--- a/allauth/account/app_settings.py
+++ b/allauth/account/app_settings.py
@@ -73,6 +73,14 @@ class AppSettings(object):
                              settings.LOGIN_URL)
 
     @property
+    def EMAIL_CONFIRMATION_COOLDOWN(self):
+        """
+        The cooldown in seconds during which, after an email confirmation has
+        been sent, a second confirmation email will not be sent.
+        """
+        return self._setting("EMAIL_CONFIRMATION_COOLDOWN", 3 * 60)
+
+    @property
     def EMAIL_REQUIRED(self):
         """
         The user is required to hand over an e-mail address when signing up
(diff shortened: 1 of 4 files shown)


jleclanche:
@jnns you are incorrect, `get_or_create` is [guaranteed atomic](https://docs.djangoproject.com/en/1.11/ref/models/querysets/#get-or-create) on indexed values.

The main issue is I have to keep the try/except due to the lack of lowercasing of the email value... this is really something that needs to be fixed.

jnns:
You're right. I didn't look close enough at the source. The algorithm for get_or_create goes like this: 

1. try to get object 
2. if that fails:
    1. try to create object
    2. if that fails as well
        1. try to get object again

But you could easily get rid of that outer try-except and 

```
def add_email(self, request, user, confirm=False, signup=False):
    email_address, created = self.get_or_create(
        user=user, 
        email__iexact=email,  # iexact lookup in get_or_create
        default={'email': email}
    )
    if created and confirm:
        email_address.send_confirmation(request, signup=signup)
    return email_address
```

jleclanche:
Paging @pennersr

pennersr:
I think we should be able to simply all of this by:

```
EmailAddress.objects.get_or_create(user=..., email__iexact=..., defaults={'email': ...})
```

jleclanche:
@pennersr very sorry about the long delay, I have rebased, updated and squashed the PR.

pennersr:
@jleclanche No problem..could you have a look, the PR is breaking the build. Thanks.

jleclanche:
@pennersr sorry... more delays... build fixed.

PR #1718, opened 2017-04-24
Alter get_avatar_url for getting a profile_image(full-size image) on Kakao provider

JooeunAhn:
I added a simple feature that gets original profile image from KAKAO API.

The original version of django-allauth's kakao provider only supports the function to get thumbnail profile image.

### DIFF
--- a/allauth/socialaccount/providers/kakao/provider.py
+++ b/allauth/socialaccount/providers/kakao/provider.py
@@ -10,7 +10,7 @@ class KakaoAccount(ProviderAccount):
     def get_avatar_url(self):
         return self.properties['thumbnail_image']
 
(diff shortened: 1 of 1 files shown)


JooeunAhn:
When executing Python: 2.7, DJANGO="Django<1.11", The Travis said, "The job exceeded the maximum time limit for jobs, and has been terminated.".

What should I do for passing that error?

JooeunAhn:
According to KAKAO API Docs, 
the image's size in thumbnail_image is 110px * 110px. 
But the image's size in profile_image is 480px * 480px ~ 1024px * 1024px.

Because the changes would be more broadly applied, I hope the changes will be merged.

PR #1723, opened 2017-04-29
Add a feature getting user's email on KAKAO provider

JooeunAhn:
Today, KAKAO API start to provide the user's email information.
So, I added a feature that gets the email from kakao api

### DIFF
--- a/allauth/socialaccount/providers/kakao/provider.py
+++ b/allauth/socialaccount/providers/kakao/provider.py
@@ -25,22 +25,18 @@ class KakaoProvider(OAuth2Provider):
         return str(data['id'])
 
     def extract_common_fields(self, data):
-        email = None
-        if data.get("kaccount_email_verified"):
-            email = data.get("kaccount_email")
-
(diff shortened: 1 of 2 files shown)


allauth/socialaccount/providers/kakao/provider.py
```diff
@@ -23,5 +24,24 @@ class KakaoProvider(OAuth2Provider):
...
+        if data.get("kaccount_email_verified"):
+            email = data.get("kaccount_email")
+
+        return dict(email=email)
+
+    def extract_email_addresses(self, data):
+        ret = []
+        email = data.get("kaccount_email")
+        verified = data.get("kaccount_email_verified")
+        if verified:
```
pennersr: Can you change this so that the email address is also added when verified is not true?

JooeunAhn:
@pennersr Of course. Does this change you want?

JooeunAhn:
@pennersr Of course!  Is this the change you want?

pennersr:
Thanks

PR #1733, opened 2017-05-21
Test cleanups and fixes

jleclanche:
This PR achieves the following things:
- Drops support for Django 1.9 (1.8 is still supported)
- Cleans up some Django <1.8 hacks
- Adds per-Django-version pypi classifiers (each classifier advertises a supported Django major version)
- Upgrades the Django 1.11 test version, now that it's been fixed
- Sets up testing using Tox on Travis. This means the test matrix is handled by tox instead of Travis, and we just tell travis which environments to pair the matrix with. This pattern makes it possible to run the matrix locally or on other CIs easily.
- Creates new tox environments `qa` and `docs`, for QA checks and docs building, respectively. Instead of `make qa` / `make docs`, the command is now `tox -e qa` / `tox -e docs`. This is also separated on Travis so that docs/qa failures don't fail *all* the builds, but only their respective one (easier to understand what's wrong).
- Adds testing on django master (currently 2.0), with failures allowed. This gives us a heads up on what to expect in the future, without making the builds red. The `fast_finish` travis keyword tells travis not to wait for the djangomaster build to report results.
- ~~Switches to yarn instead of npm to install standardjs. This speeds up the build considerably.~~ *apparently this isn't possible on non-js travis envs...*

Please don't squash :)

### DIFF
--- a/docs/providers.rst
+++ b/docs/providers.rst
@@ -1206,7 +1206,7 @@ VK
 --
 
 App registration
-    http://vk.com/apps?act=settings
+    https://vk.com/editapp?act=create
 
 Development callback URL ("Site address")
     http://localhost


jleclanche:
@pennersr thoughts?

lorddaedra:
@jleclanche  LGTM

but I suggest to save 3.3 support for tests with Django 1.8

tox.ini
```diff
@@ -1,15 +1,38 @@
...
+	py27: mock >= 1.0.1
+	django18: Django>=1.8,<1.9
+	django110: Django>=1.10,<1.11
+	django111: Django>=1.11,<2.0
+	djangomaster: https://github.com/django/django/archive/master.tar.gz
 commands =
-    coverage run manage.py test allauth
-    coverage report
-    coverage html
+	coverage run manage.py test allauth
```
lorddaedra: Can we add "-Wall" here to show Deprecation Warnings?..

jleclanche: Can you pass -Wall to coverage? I think it's just for python

lorddaedra: You are right.

Then let's add 

```
[testenv]
setenv =
    PYTHONWARNINGS = all
```

jleclanche:
I did a bad push on my master and now I can't reopen the PR. I'll file it again.

PR #1752, opened 2017-05-30
Test cleanups and fixes

jleclanche:
Re-filing #1733 -- closed by mistake and unable to reopen.

No squashing please!


jleclanche:
Per recommendation from @lorddaedra I also enabled warnings for everything by default. This has allowed me to fix the tests, they are now green on Django 2.0.

.travis.yml
```diff
@@ -1,46 +1,57 @@
...
- - "3.3"
- - "3.4"
- - "3.5"
- - "3.6"
+
+python: "3.6"
+
 env:
   matrix:
-   - DJANGO="Django<1.9"
```
lorddaedra: Okay, same comment as in previous pull request: Python 3.3 with Django 1.8 is still supported until September, 2017, you can check it in version history table here https://en.wikipedia.org/wiki/CPython or directly in https://www.python.org/dev/peps/pep-0398/#lifespan ...

I hope this pull request will be merged before September, 2017 :-)))

jleclanche: Me too :) But that sounds unnecessary; it's a lot of extra time/effort to spend on the one and only 3.3 test which will no longer be supported in just 4 months. Requiring 3.4 is very reasonable for allauth to do.

Makefile
```diff
@@ -8,29 +8,10 @@ po:
...
-	( . $(VIRTUALENV)/bin/activate; cd docs; make html )
-
-isort-check:
-	$(ISORT) -c
-
 isort-fix:
 	$(ISORT)
 
-pep8:
-	flake8 --exclude=migrations allauth
```
lorddaedra: Hmm, changes in Makefile looks as regression, why would you remove that lines?..

pennersr:
Thanks for taking the time to prep this pull request! Dropping 3.3/1.8 is +1 for me...

pennersr:
It seems that standardjs checking is broken. In the qa output I see this:

`standard: Node v4 or greater is required. `standard` did not run.`

jleclanche:
I wish coverage wouldn't spam as much as it does...

jleclanche:
@pennersr standardjs issues fixed; I pulled it into its own tox env.

jleclanche:
@pennersr Should be good to land now. No squashing! :P

jleclanche:
Added fixes for isort 4.2.8, as well as tox.ini merging (timothycrosley/isort#487)

jleclanche:
@pennersr this is not an easy PR to keep rebasing; what's blocking it? can I help?

pennersr:
@jleclanche  Sorry, just me being swamped.

PR is looking good.. the only thing I am not fond of is moving the isort.cfg into the tox.ini -- these two should not be intertwined, for example, it will break my editor setup which isorts on save... could we move that to its original location again?

jleclanche:
@pennersr 👍 i dropped that particular commit then -- it should have been transparent though, merely require a `pip install isort --upgrade`.

pennersr:
Thanks!

PR #1757, opened 2017-06-10
Fixed potentional security issue with leaked password tokens

### LINKED ISSUE #1755: Leaking of password reset token through the reset url
I think it is possible to leak the password reset token since it is left in the url. 

In Django 1.11 the token is stripped during a redirect ([docs](https://docs.djangoproject.com/en/1.11/releases/1.11/#django-contrib-auth), [code](https://github.com/django/django/blob/1.11.2/django/contrib/auth/views.py#L461)) to prevent the token from being taken in the referrer header from 3rd party apps on the page. I haven't dug too deeply into the source for this project but at a first glance it seems that the vulnerability exists. 

If this is the case would be happy to help fix similar to [django-registration-redux](https://github.com/macropin/django-registration/pull/268), or at the very least alert you to the issue. Let me know if you guys need any help!

joshblum:
Django 1.11+ prevents password tokens from being leaked through the
HTTP Referer header if a template calls out to third-party resources
(i.e., JS or CSS) by setting token in session and redirecting.

Fixes #1755

### DIFF
--- a/allauth/account/tests.py
+++ b/allauth/account/tests.py
@@ -284,6 +284,9 @@ class AccountTests(TestCase):
         # Extract URL for `password_reset_from_key` view and access it
         url = body[body.find('/password/reset/'):].split()[0]
         resp = self.client.get(url)
+        # Follow the redirect the actual password reset page with the key
+        # hidden.
+        resp = self.client.get(resp.url)
         self.assertTemplateUsed(
             resp,
             'account/password_reset_from_key.%s' %
@@ -336,8 +339,11 @@ class AccountTests(TestCase):
         user = self._request_new_password()
         body = mail.outbox[0].body
         url = body[body.find('/password/reset/'):].split()[0]
+        resp = self.client.get(url)
+        # Follow the redirect the actual password reset page with the key
+        # hidden.
         resp = self.client.post(
-            url,
+            resp.url,
             {'password1': 'newpass123',
              'password2': 'newpass123'})
         self.assertTrue(is_authenticated(user))
--- a/allauth/account/tests.py
+++ b/allauth/account/tests.py
@@ -286,7 +286,8 @@ class AccountTests(TestCase):
         resp = self.client.get(url)
         # Follow the redirect the actual password reset page with the key
         # hidden.
-        resp = self.client.get(resp.url)
+        url = resp.url
+        resp = self.client.get(url)
         self.assertTemplateUsed(
             resp,
             'account/password_reset_from_key.%s' %
--- a/allauth/account/views.py
+++ b/allauth/account/views.py
@@ -39,6 +39,7 @@ from .utils import (
     url_str_to_user_pk,
 )
 
+
 INTERNAL_RESET_URL_KEY = "set-password"
 INTERNAL_RESET_SESSION_KEY = "_password_reset_key"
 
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -11,7 +11,7 @@ env:
    - DJANGO="Django<1.9"
    - DJANGO="Django<1.10"
    - DJANGO="Django<1.11"
-   - DJANGO="Django==1.11b1"
+   - DJANGO="Django>=1.11"
```
lorddaedra: `<2.0` is better here

setup.py
```diff
@@ -149,6 +149,9 @@ def find_package_data(where=".", package="", exclude=standard_exclude,
         'Programming Language :: Python :: 3.4',
         'Programming Language :: Python :: 3.5',
         'Framework :: Django',
+        'Framework :: Django :: 1.9',
```
lorddaedra: 1.9 is not supported, better do not include it, instead you can add 1.8, which is supported

.travis.yml
```diff
@@ -31,7 +31,7 @@ matrix:
     - python: "3.3"
       env: DJANGO="Django<1.11"
     - python: "3.3"
-      env: DJANGO="Django==1.11b1"
+      env: DJANGO="Django>=1.11"
```
lorddaedra: <2.0 is better here

tox.ini
```diff
@@ -8,7 +8,7 @@ deps =
     django18: Django < 1.9
     django19: Django < 1.10
     django110: Django < 1.11
-    django111: Django==1.11b1
+    django111: Django >= 1.11
```
lorddaedra: `<2.0` is better here

joshblum: done

setup.py
joshblum: my mistake, done!

.travis.yml
joshblum: done

.travis.yml
joshblum: done

joshblum:
@lorddaedra I'm getting the following error when testing:

```
isort $(find /home/travis/build/pennersr/django-allauth/allauth -not -path '*/migrations/*' -type f -name '*.py' -not -name '__init__.py' -print) -c
ERROR: /home/travis/build/pennersr/django-allauth/allauth/account/views.py Imports are incorrectly sorted.
ERROR: /home/travis/build/pennersr/django-allauth/allauth/socialaccount/providers/dwolla/urls.py Imports are incorrectly sorted.
ERROR: /home/travis/build/pennersr/django-allauth/allauth/socialaccount/providers/dwolla/views.py Imports are incorrectly sorted.
```

I didn't edit any import statements, do you have any suggestions for getting tests to pass?

pennersr:
Please rebase your changes on latest master, then the problem will be gone (git diff master..fix-referrer -- this will show that lots of stuff has changed between master and your branch)

joshblum:
@pennersr Updated, let me know if I should make any other changes!

joshblum:
@pennersr builds still seem to be failing due to an import sort error. Any insights? I'm not sure what to fix since this PR didn't change any imports. 

`ERROR: /home/travis/build/pennersr/django-allauth/allauth/account/views.py Imports are incorrectly sorted.`

pennersr:
The flake8 build output is a bit misleading.. it is actually this line that is the problem:

ERROR: /Users/pennersr/src/django-allauth/allauth/account/views.py Imports are incorrectly sorted.

So, you need two empty lines before INTERNAL_RESET_URL_KEY = "set-password"

joshblum:
Thanks!!

pennersr:
Thanks in return!

PR #1771, opened 2017-06-30
Updates dropbox provider to OAuth2 with Dropbox v2 API endpoints

### LINKED ISSUE #1760: Dropbox Oauth1 and v1 API is obsolete at June 28, 2017
The `dropbox` OAuth1 provider will no longer work after June 28.

The `dropbox_oauth2` provider will also probably crash after June 28 as the `v1` API URIs are changing for both oauth2 flow and accessing user account info:
https://www.dropbox.com/developers/reference/migration-guide

My own project is hard-coded to rely on the provider name being `dropbox` and not `dropbox_oauth2` so I opted to just clobber the `dropbox/` provider with the workable `dropbox_oauth2/` code and then update this new oauth2 provider to use the newer Dropbox v2 API endpoints.

https://github.com/pztrick/django-allauth/commit/03f27e174d58b9204777186f6d15a6109ff1e21a

The commit above is working with Dropbox v2 API/OAuth2 but few problems:

1. I'm several hundred commits behind
2. We probably want to remove the redundant `dropbox_oauth2` provider OR refactor so that it uses same code / is backwards-compatible if people just to upgrade `allauth`.

I don't have time today to go all the way and a simple rebase against `upstream/master` produced errors so I can't make a clean P/R right now. Just wanted to share my commit here in case anyone else wants to take it up before I can get back to making a P/R. (I do expect the `dropbox_oauth2` to throw error retrieving account info like `display_name` after June 28.)

pztrick:
Fixes #1760

### DIFF
--- a/allauth/socialaccount/providers/dropbox/provider.py
+++ b/allauth/socialaccount/providers/dropbox/provider.py
@@ -10,7 +10,6 @@ class DropboxOAuth2Account(ProviderAccount):
 class DropboxOAuth2Provider(OAuth2Provider):
     id = 'dropbox'
     name = 'Dropbox'
-    package = 'allauth.socialaccount.providers.dropbox'
     account_class = DropboxOAuth2Account
 
     def extract_uid(self, data):
--- a/allauth/socialaccount/providers/dropbox/provider.py
+++ b/allauth/socialaccount/providers/dropbox/provider.py
@@ -1,23 +1,24 @@
+from allauth.socialaccount import providers
 from allauth.socialaccount.providers.base import ProviderAccount
-from allauth.socialaccount.providers.oauth.provider import OAuthProvider
+from allauth.socialaccount.providers.oauth2.provider import OAuth2Provider
 
 
-class DropboxAccount(ProviderAccount):
+class DropboxOAuth2Account(ProviderAccount):
     pass
 
 
-class DropboxProvider(OAuthProvider):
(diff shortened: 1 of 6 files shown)


pztrick:
I do think it makes sense to just remove the `dropbox_oauth2` provider after September 28 when `v1` API goes down as:

1. The `uid` field no longer exists
2. The OAuth2 URLs have changed
3. Anyone developing against `providers.dropbox_oauth2` probably also wrote their app against Dropbox `v1` API to create/read/delete Dropbox files; their app is already going to be broken anyway. :)

I did add a `DeprecationWarning` to `dropbox_oauth2` and updated the docs to reflect as much.

cheers.

pztrick:
Rebased against `upstream/master` and fixed for `pyflakes` linting.

allauth/socialaccount/providers/dropbox/provider.py
```diff
@@ -1,23 +1,24 @@
...
+class DropboxOAuth2Account(ProviderAccount):
     pass
 
 
-class DropboxProvider(OAuthProvider):
+class DropboxOAuth2Provider(OAuth2Provider):
     id = 'dropbox'
     name = 'Dropbox'
-    account_class = DropboxAccount
+    package = 'allauth.socialaccount.providers.dropbox'
```
pennersr: This property can be dropped -- it is a legacy property

PR #1775, opened 2017-07-08
Kakao can use client_secret now

punkyoon:
Kakao now supports `client_secret`.

so we don't need to have `client.consumer_secret = None` anymore.


punkyoon:
image


check please

@pennersr

PR #1778, opened 2017-07-10
Add Dataporten provider

JakobGM:
Hi, there!

[Dataporten](https://www.uninett.no/en/service-platform-dataporten) is an oAuth2 provider for the education sector in Norway, and I have used it as a custom allauth provider for a couple of months now. If you are interested, I would like to add it as a supported provider.

Unfortunately, I have not been able to [pass the test suite](https://travis-ci.org/JakobGM/django-allauth) (EDIT: Not the case any longer). It fails at trying to fetch a user with multiple logins, possibly related to the handling of refresh tokens.

I have been debugging the issue for a while now, but I hope that you might be able to help me understand where the issue lies before I invest a lot more time on fixing this issue. If PRs with failing test suits are undesirable, I will delete this PR and create an issue instead.

Thanks in advance,
Jakob

### DIFF
--- a/allauth/socialaccount/providers/dataporten/provider.py
+++ b/allauth/socialaccount/providers/dataporten/provider.py
@@ -1,0 +1,94 @@
+from allauth.socialaccount.providers.base import ProviderAccount
+from allauth.socialaccount.providers.oauth2.provider import OAuth2Provider
+
+
+class DataportenAccount(ProviderAccount):
+    def get_avatar_url(self):
+        '''
+        Returns a valid URL to an 128x128 .png photo of the user
+        '''
+        # Documentation for user profile photos can be found here:
+        # https://docs.dataporten.no/docs/oauth-authentication/
+        base_url = 'https://api.dataporten.no/userinfo/v1/user/media/'
+        return base_url + self.account.extra_data['profilephoto']
+
+    def to_str(self):
+        '''
+        Returns string representation of a social account. Includes the name
+        of the user.
+        '''
+        dflt = super(DataportenAccount, self).to_str()
+        return '%s (%s)' % (
+            self.account.extra_data.get('name', ''),
+            dflt,
+        )
+
+
+class DataportenProvider(OAuth2Provider):
+    id = 'dataporten'
+    name = 'Dataporten'
+    account_class = DataportenAccount
+
+    def extract_uid(self, data):
+        '''
+        Returns the primary user identifier, an UUID string
+        See: https://docs.dataporten.no/docs/userid/
+        '''
+        return data['userid']
+
+    def extract_extra_data(self, data):
+        '''
+        Extracts fields from `data` that will be stored in
+        `SocialAccount`'s `extra_data` JSONField.
+
+        All the necessary data extraction has already been done in the
+        complete_login()-view, so we can just return the data.
+        PS: This is default behaviour, so we did not really need to define
+            this function, but it is included for documentation purposes.
+
+        Typical return dict:
+        {
+            "userid": "76a7a061-3c55-430d-8ee0-6f82ec42501f",
+            "userid_sec": ["feide:andreas@uninett.no"],
+            "name": "Andreas \u00c5kre Solberg",
+            "email": "andreas.solberg@uninett.no",
+            "profilephoto": "p:a3019954-902f-45a3-b4ee-bca7b48ab507",
+        }
+        '''
+        return data
+
+    def extract_common_fields(self, data):
+        '''
+        This function extracts information from the /userinfo endpoint which
+        will be consumed by allauth.socialaccount.adapter.populate_user().
+        Look there to find which key-value pairs that should be saved in the
+        returned dict.
+
+        Typical return dict:
+        {
+            "userid": "76a7a061-3c55-430d-8ee0-6f82ec42501f",
+            "userid_sec": ["feide:andreas@uninett.no"],
+            "name": "Andreas \u00c5kre Solberg",
+            "email": "andreas.solberg@uninett.no",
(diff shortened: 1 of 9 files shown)


allauth/socialaccount/providers/dataporten/provider.py
```diff
@@ -0,0 +1,96 @@
...
+    account_class = DataportenAccount
+
+    def extract_uid(self, data):
+        '''
+        Returns the primary user identifier, an UUID string
+        See: https://docs.dataporten.no/docs/userid/
+        '''
+        return data['userid']
+
+    def extract_extra_data(self, data):
```
JakobGM: I can remove this unecessary function definition if desired. I added it as it felt more explicit at the time.

allauth/socialaccount/providers/dataporten/provider.py
```diff
@@ -0,0 +1,96 @@
...
+            "username": "andreas",
+        }
+        '''
+        # Make shallow copy to prevent possible mutability issues
+        data = dict(data)
+
+        # If a Feide username is available, use it. If not, use the "username"
+        # of the email-address
+        username_set = False
+        for userid in data.get('userid_sec'):
```
JakobGM: This loop here feels a bit unpythonic, but I can't really think of a better way to write this...

allauth/socialaccount/providers/dataporten/tests.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def get_login_response_json(self, with_refresh_token=True):
+        rt = ''
+        if with_refresh_token:
+            rt = ',"refresh_token": "testrf"'
+        return '''{
+            "access_token":"testac",
+            "expires_in":3600,
+            "scope": "userid profile groups"
+            %s
```
JakobGM: I can't find a single mention of refresh tokens in the [Dataporten documentation](https://docs.dataporten.no). But as I understand it, it is meant to be included here according to the oAuth2 specification. The default implementation seems to have almost the same behaviour.

allauth/socialaccount/providers/dataporten/views.py
```diff
@@ -0,0 +1,75 @@
...
+        groups_data.raise_for_status()
+
+        # The endpoint returns json-data and it needs to be decoded
+        extra_data = userinfo_response.json()['user']
+        extra_data['groups'] = groups_data.json()
+
+        # Finally test that the audience property matches the client id
+        # for validification reasons, as instructed by the Dataporten docs
+        # if the userinfo-response is used for authentication
+        if userinfo_response.json()['audience'] != app.client_id:
```
JakobGM: I think maybe this is not the correct exception to raise here. Do you prefer a specific one?

JakobGM:
I found the bug!

The bug was caused by making **two** http requests, not one, during the `complete_login` call. The test framework only allows a custom adapter to return _one_ `MockedResponse` instance from the `get_mocked_response` test method, making it 'impossible' to mock out the second request.

For the moment I have 'fixed' this bug by not making the second request in the first place. The second request was made to the Dataporten groups API endpoint, in order to retrieve additional group membership information related to the user, such as enrolled courses, university name, faculty status, etc.. This information is not strictly necessary for authentication, as it contains none of the fields used by `extract_common_fields()`.

If we want to hit the Groups API endpoint in order to populate the `extra_data` JSONField of the SocialAccount, we have to enable providers to provide _several_ `MockedResponse` instances. I see that you have already laid most of the ground work for this with the `mocked_response` class. I propose two possible ways of finishing the last part of this functionality:

1) Make all `get_mocked_response()` methods return a `tuple` of `MockedResponse` objects and rename the method to `get_mocked_responses()`. Then we can splat this list in the `OAuth2TestMixin.test_login` context manager. This breaks backwards compatibility, though...

2) Keep everything as it is, but allow returning a `tuple` of `MockedResponse` objects in addition to just returning a single object. Then we have to add conditional branching logic when receiving `__iter__`-able items in the `test_login()` test. This has the drawback of introducing an additional branch, but this way it becomes backwards compatible.

I think option 2 is the best one, as it seems that it only needs to be checked at one, single place in the code. Are you able to comfirm this? If you are interested, I can take a shot at implementing this in a seperate PR, let me know.

Lastly, this bug was pretty nasty to debug, as it was non-deterministic on my local machine. If the aforementioned functionality is not desired, I think we should atleast add a test which screams if you are doing more than 1 request in the `complete_login` view. If we use `unittest.mock`, we can check how many times `requests.request` is invoked, and assert that it is equal to 1.

allauth/socialaccount/providers/dataporten/tests.py
pennersr: I must admit, the with_refresh_token is something that needs to be refactored at some point in time.. for now what you are doing here is okay -- others are doing it the same way.

allauth/socialaccount/providers/dataporten/views.py
pennersr: No, this is fine... a specific one would make sense if it somehow could be handled in a special way, but that's not the case.

pennersr:
Have a look at the get_mocked_response in the BitbucketTests ... it shows an example of returning multiple mocked responses. So I think what you need with respect to mocking is already in?

JakobGM:
@pennersr Oh, that's great! I would prefer to add back the functionality in a separate PR, if at all. I have been reconsidering adding the Group functionality anyway, since it seems a bit outside the scope of allauth. Would you agree?

The Group-API endpoint changes every semester anyway, so I think it is a bit weird to save eventually stale data in `extra_data`.

pennersr:
Agreed... going ahead and merging this as is. Thanks for contributing.

JakobGM:
@pennersr Likewise, thanks for a great app!

PR #1779, opened 2017-07-10
Add support for Authentiq

skion:
This should add support for passwordless sign-in via Authentiq ID.


skion:
Can see if I can add a test soon...

skion:
Should be better now!

PR #1786, opened 2017-07-21
Customizable redirect_connect_url

magnunleno:
Adds the possibility to customizes the redirect URL after a successful social account connection.


pennersr:
The socialaccount adapter has this entry point that can be used:

```
    def get_connect_redirect_url(self, request, socialaccount):
```

PR #1790, opened 2017-07-27
Fix user_logged_out signal behavior

### LINKED ISSUE #1787: About user_logged_out signal
Hi everyone,

Using the default account adapter, allauth [sends](https://github.com/pennersr/django-allauth/blob/81af6a54be579a565b3eef1766642cbaca7ee5a3/allauth/account/adapter.py#L395) the ``allauth.account.signal.user_logged_out`` signal on logout. But this signal is sent after the call to the classical django' logout function. So it misses a few things, e.g. `request.session` has already been flushed when this signal is emitted.

Maybe the docs could:
- specify when the signal is sent, i.e. at the (very) end of the logout flow,
- say [`django.contrib.auth.signals.user_logged_out`](https://docs.djangoproject.com/en/dev/ref/contrib/auth/#django.contrib.auth.signals.user_logged_out) is still sent on logout. This signal is sent earlier in the logout flow, such as listeners can still accessed the content of `request.session` before it has been flushed.

Just to say, initially i was doubting this signal (allauth' one) is worthwhile, because it seems like a messier copy of the django' one. But in fact they provide multiple listening points in the logout flow, namingly pre/post logout. With django' signal, you can access the session before it has been flushed, but it's useless to set session value. With allauth' one, session will be empty, but you can set session values and they will be kept. So "why not".

aureplop:
Changes the behavior of allauth' user_logged_out signal to the same as [django' one](https://docs.djangoproject.com/en/1.11/ref/contrib/auth/#django.contrib.auth.signals.user_logged_out). To perform this, django' signal is forwarded as allauth' one if the logout comes from the allauth' logout view (to be precise, from logout method of the default AccountAdapter).

It also corrects unexpected test failures when tests are run from a fresh copy of the repo.

Fixes #1787.

### DIFF
--- a/allauth/account/signals.py
+++ b/allauth/account/signals.py
@@ -1,8 +1,8 @@
+from django.contrib.auth.signals import user_logged_out  # noqa
 from django.dispatch import Signal
 
 
 user_logged_in = Signal(providing_args=["request", "user"])
-user_logged_out = Signal(providing_args=["request", "user"])
 
 # Typically followed by `user_logged_in` (unless, e-mail verification kicks in)
 user_signed_up = Signal(providing_args=["request", "user"])
--- a/allauth/account/adapter.py
+++ b/allauth/account/adapter.py
@@ -33,7 +33,6 @@ from ..utils import (
     get_user_model,
     import_attribute,
 )
-from .signals import user_logged_out
 
 
 try:
@@ -389,12 +388,7 @@ class DefaultAccountAdapter(object):
         django_login(request, user)
 
     def logout(self, request):
-        user = request.user
         django_logout(request)
-        user_logged_out.send(
-            sender=user.__class__,
-            request=request,
-            user=user)
 
     def confirm_email(self, request, email_address):
         """
--- a/allauth/account/tests.py
+++ b/allauth/account/tests.py
@@ -22,13 +22,14 @@ from allauth.account.models import (
     EmailConfirmation,
     EmailConfirmationHMAC,
 )
-from allauth.tests import TestCase, patch
+from allauth.tests import Mock, TestCase, patch
 from allauth.utils import get_user_model, get_username_max_length
 
 from . import app_settings
 from ..compat import is_authenticated, reverse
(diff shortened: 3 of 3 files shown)


tox.ini
```diff
@@ -43,6 +43,7 @@ commands =
 [testenv:standardjs]
 skip_install = True
 commands =
+	/usr/bin/env bash -c "mkdir -p {toxinidir}/node_modules"
```
pennersr: Why is this needed in the scope of this PR?

allauth/account/signals.py
```diff
@@ -1,9 +1,22 @@
...
+from django.contrib.auth.signals import user_logged_out as dj_user_logged_out
+from django.dispatch import Signal, receiver
 
 
 user_logged_in = Signal(providing_args=["request", "user"])
 user_logged_out = Signal(providing_args=["request", "user"])
 
+
+@receiver(dj_user_logged_out, dispatch_uid="forwards_user_logged_out")
+def forwards_user_logged_out(sender, **kwargs):
```
pennersr: Wondering, is this really necessary? What would go wrong if you just imported the django signal as is, without doing any forwarding/intervention at all.

aureplop: If you use another application which uses the django logout function, it means receivers of allauth signal will also receive signals initiated from this other application.
I think it's a better behavior that the allauth' signal is only fired from something which happened in allauth.

tox.ini
aureplop: It's correct, I just opened a separate PR (#1791) for this.

allauth/account/signals.py
pennersr: I can't really think of a use case where it is really required for the signal not to fire. Suppose there is another third party app that hooks up the django logout directly -- given that the user is actually being logged out as would be the case when logging out via allauth, I do think it makes more sense to always fire the signal. Also, for sake of simplicitly of the code, simply relying on the Django signal is a one liner, versus quite some intricate machinery introduced in this PR.

PR #1791, opened 2017-07-27
Fix testenv standardjs

aureplop:
Due to the lack of a `package.json` and `node_modules` directory on the
repo, npm could install standard in another directory than
`node_modules/bin/`, which means the standardjs doesn't start.
This PR fixes this behavior.

If a package.json file is added, the mkdir instruction from tox.ini can be
deleted.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ build
 dist
 *.egg*
 example/local_settings.py
+node_modules/
 
 # Unit test / coverage reports
 htmlcov/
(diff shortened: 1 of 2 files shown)


pennersr:
Where do you see the standardjs build failing without these changes? When I run things locally node_modules gets automatically created. Additionally, standardjs seems to be running fine on travis?: https://travis-ci.org/pennersr/django-allauth/jobs/258289036

aureplop:
This seems to happens because npm find a `node_modules` directory in a parent directory of my local repo. Indeed if I rename this directory, `tox -e standardjs` succeeds.
I'm not sure the wrong is on my side or not...

pennersr:
Ok, makes sense... let's merge this.

PR #1793, opened 2017-07-27
Make Norwegian local

ggjersund:
Created Norwegian translation, Bokmål, 'no-nb', which is one of the following ['Bokmål', 'Nynorsk'] written languages in Norway.

First ever GitHub contribution, so correct me if I'm doing something wrong here :)


pennersr:
Thanks!

PR #1796, opened 2017-08-04
Document SOCIALACCOUNT_ENABLED setting

jonashaag:


pennersr:
This is not a setting, it is an internal switch hardcoded to the following logic:

```
SOCIALACCOUNT_ENABLED = 'allauth.socialaccount' in settings.INSTALLED_APPS
```

If you want to disable the social account functionality simply remove the socialaccount app from the installed apps list.

jonashaag:
I see. Not adding `allauth.socialaccount` breaks the default templates, though. For example, the login template has `{% load account socialaccount %}`

pennersr:
The provided templates are only meant as a starting point. Change them to match your project configuration...

PR #1812, opened 2017-08-28
Update information to match dev page

melvyn-sopacua:
The names mentioned do no longer match names in LinkedIn developer backend.
Further more, Oauth 2, will not work without explicit redirect URI's and
the documentation is misleading.

Up to 200 redirect URI's can be specified per application, so
development URI's are no issue and non-existing domains or only locally
resolvable domains are accepted (tested).

Reworded the Oauth 1 bit, but no testing has been done on my part to verify if
empty callbacks are still accepted.

### DIFF
--- a/docs/providers.rst
+++ b/docs/providers.rst
@@ -760,9 +760,20 @@ fetching the access token::
 App registration (get your key and secret here)
     https://www.linkedin.com/secure/developer?newapp=
 
-Development callback URL
-    Leave the OAuth redirect URL empty.
+Authorized Redirect URLs (Oauth2)
+*********************************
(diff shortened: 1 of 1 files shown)


docs/providers.rst
```diff
@@ -760,9 +760,20 @@ fetching the access token::
 App registration (get your key and secret here)
     https://www.linkedin.com/secure/developer?newapp=
 
-Development callback URL
-    Leave the OAuth redirect URL empty.
+Authorized Redirect URLs (Oauth2)
```
pennersr: Sorry for nitpicking, but can you rename Oauth2 to OAuth2 (below as well), thanks.

PR #1822, opened 2017-09-09
Add Yahoo Provider

theju:


theju:
There is one problem (if you use multiple scopes) that Yahoo uses request parameter 'scopes' instead of the standard parameter 'scope'.

PR #1825, opened 2017-09-12
add autocomplete attributes

xi:
see https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofilling-form-controls:-the-autocomplete-attribute

### DIFF
--- a/allauth/account/forms.py
+++ b/allauth/account/forms.py
@@ -71,12 +71,16 @@ class PasswordField(forms.CharField):
         kwargs['widget'] = forms.PasswordInput(render_value=render_value,
                                                attrs={'placeholder':
                                                       kwargs.get("label")})
+        autocomplete = kwargs.pop('autocomplete', None)
+        if autocomplete is not None:
+            kwargs['widget'].attrs['autocomplete'] = autocomplete
         super(PasswordField, self).__init__(*args, **kwargs)
 
 
 class SetPasswordField(PasswordField):
 
     def __init__(self, *args, **kwargs):
+        kwargs['autocomplete'] = 'new-password'
         super(SetPasswordField, self).__init__(*args, **kwargs)
         self.user = None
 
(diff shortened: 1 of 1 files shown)


davidsomnus:
Tested this with django-allauth (0.41.0) and it works nicely.

allauth/account/forms.py
```diff
@@ -127,8 +134,9 @@ def __init__(self, *args, **kwargs):
             assert app_settings.AUTHENTICATION_METHOD \
                 == AuthenticationMethod.USERNAME_EMAIL
             login_widget = forms.TextInput(attrs={'placeholder':
-                                                  _('Username or e-mail'),
-                                                  'autofocus': 'autofocus'})
+                                                  _('Username e-mail'),
```
pennersr, line 137: Why did you remove "or" here?

allauth/account/forms.py
```diff
@@ -127,8 +134,9 @@ def __init__(self, *args, **kwargs):
...
+                                                  'autofocus': 'autofocus',
+                                                  'autocomplete': 'username'})
```
pennersr, line 139: Wouldn't autocomplete=login make more sense, given that it is either the username or the email?

allauth/account/forms.py
```diff
@@ -111,14 +116,16 @@ def __init__(self, *args, **kwargs):
             login_widget = forms.TextInput(attrs={'type': 'email',
                                                   'placeholder':
                                                   _('E-mail address'),
-                                                  'autofocus': 'autofocus'})
+                                                  'autofocus': 'autofocus',
```
pennersr, line 119: autofocus was dropped, so it should be removed from this PR as well.

pennersr (changes requested):

xi:
This PR is 3 years old and I am no longer using django-allauth. I don't have time to work on this so it is up for grabs.

allauth/account/forms.py
xi: Probably unintentional. Feel free to restore the original string.

pennersr:
Closed via 239a7235

PR #1851, opened 2017-10-14
create base email template for easier customizability

protux:
Hi,

it would be pretty useful to have a `base.txt` to overwrite for easier extensability, similar to `base.html`.

There are countries out there (including mine) which demand by law that you have an imprint in every mail you send as an organization.

Right now you need to copy all relevant mail templates into your project and edit every single template. Additionally, you need to repeat that if the mail templates will be updated.

This PR is my suggestion to add a `base.txt` to `templates/` which only contains `{% block content %}{% endblock content %}`. This way one can add custom stuff (e.g. an imprint).


allauth/templates/account/email/email_confirmation_message.txt
```diff
@@ -1,8 +1,8 @@
-{% load account %}{% user_display user as user_display %}{% load i18n %}{% autoescape off %}{% blocktrans with site_name=current_site.name site_domain=current_site.domain %}Hello from {{ site_name }}!
+{% extends 'base.txt' %}{% load account %}{% block content %}{% user_display user as user_display %}{% load i18n %}{% autoescape off %}{% blocktrans with site_name=current_site.name site_domain=current_site.domain %}Hello from {{ site_name }}!
```
pennersr, line 1: Perhaps it would be better to have a `allauth/templates/account/email/base_message.txt` so that it is clear that this base is email message specific, and, it is scoped within the allauth template context.

pennersr:
Closed via 14108d70

Issue #620, opened 2014-05-21
Use the shortcut get_current_site instead of get_current (from the model)


jensenbox:
We would prefer to not have to do a full install of the sites framework.

There is a simple work around, use https://docs.djangoproject.com/en/dev/ref/contrib/sites/#get-current-site-shortcut

This simple change will allow for using the framework or not easily - but there could be implications that I am not aware of of course.

pennersr:
https://codeberg.org/allauth/django-allauth/issues/620

PR #1856, opened 2017-10-23
Don't require django.contrib.sites

jonashaag:
If not using social accounts, do not depend on django.contrib.sites. Refs #620.


allauth/utils.py
```diff
@@ -264,7 +264,7 @@ def build_absolute_uri(request, location, protocol=None):
     from .account import app_settings as account_settings
 
     if request is None:
-        site = Site.objects.get_current()
+        site = get_current_site(request)
```
pennersr: `None` is not an acceptable input to `get_current_site`

jonashaag: It is if `django.contrib.sites` is installed. If you don't have it installed, you can't use `build_absolute_uri` outside a request context, which makes perfect sense (where would it get domain information from?).

pennersr: I understand,  but we need to be certain that allauth is properly handing over a request everywhere. Also, this method is documented as:

    Like request.build_absolute_uri, but gracefully handling
    the case where request is None.

With these changes, that is no longer true. So the docstring needs updating...

jonashaag: I'm not sure what you mean, to be honest.

The only place where this is called with `request=None` is from `get_email_confirmation_url` if you're doing email confirmation emails outside a request context. So in this case, we'll have to construct a URL from the current Django site. If you don't have one set up, or don't have `django.contrib.sites` installed, it won't work. This also didn't work before my patch.

The only thing I see that we could maybe explain a bit better is that you can't do outside-request-context email confirmations without `django.contrib.sites` installed. Is this what you want me to do? :)

jonashaag:
Bump

jonashaag:
This is now fine in terms of CI. Coveralls -0.001% is stupid.

greyhare:
So I'm a strong believer that site information should be pulled from an environment variable and not a database, because 12factor and I have One Site and "runserver" shouldn't be dialing out to some other site.

And I see a lot of things about Coverage on this pull request and have no idea what the problem is.

Why can't the sites framework be optional? "if 'django.contrib.sites' in settings.INSTALLED_APPS" or something like that?

iMerica:
100% agree.

The site name and domain should be in the environment, not in the database. 

Currently this Sites requirements forces Django applications to be "Stateful" e.g you can't bootstrap a new environment without manually creating your default site object. Otherwise all emails come from "example.com".

pennersr:
Closed via 16ee5458

PR #1867, opened 2017-11-12
Correct HTML head title of the email template

radwon:


radwon:
@pennersr Can you take a quick look at this. It's just a simple template correction. Thanks.

PR #1893, opened 2017-11-29
Add ACCOUNT_SIGNUP_ADD_MESSAGE option

jonashaag:
Use case: Hide the "confirmation email sent" message **only directly after signing up** (if optional email verification)

### DIFF
--- a/allauth/account/tests.py
+++ b/allauth/account/tests.py
@@ -7,6 +7,7 @@ from datetime import timedelta
 from django import forms
 from django.conf import settings
 from django.contrib.auth.models import AbstractUser, AnonymousUser
+from django.contrib import messages
 from django.contrib.sites.models import Site
 from django.core import mail, validators
 from django.core.exceptions import ValidationError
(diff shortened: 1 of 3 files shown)


jonashaag:
I surrender. No idea how flake8 expects me to indent that line here.

Rest of the tests pass.

jonashaag:
Hi @pennersr, any feedback on this?

pennersr:
There is no need to introduce additional settings for this. If you override the default email_confirmation_sent.txt template with a blank one, this automatically disables that message. I prefer this above introducing another setting.

jonashaag:
No, this also disables the message when using the "re-send confirmation message" feature.

jonashaag:
@pennersr bump

jonashaag:
@pennersr Can I get some feedback on this?

pennersr:
Adding a one off setting for just this specific message introduces inconsistency, you would expect a similar setting for the other messages as well. Instead, opted for adding more context to the template so that you can decide inside the template how you want to handle this. See: fc772f29

PR #1901, opened 2017-12-05
update all urls.py files to Django 2.0.

wardbradt:


pennersr:
This cannot be merged, we need to support Django 1.x as well.

PR #1919, opened 2018-01-03
Add signal for failed user authentication 

jvlomax:
Emits a new signal when a user failed to authenticate. This is so we can run our own custom code to block users permanently if they fail login a certain number of attempts.

### DIFF
--- a/allauth/account/adapter.py
+++ b/allauth/account/adapter.py
@@ -29,6 +29,7 @@ from django.utils.encoding import force_text
 from django.utils.translation import ugettext_lazy as _
 
 from . import app_settings
+from .signals import user_auth_failed
 from ..utils import (
     build_absolute_uri,
     email_address_exists,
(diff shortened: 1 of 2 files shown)


codebreach:
@pennersr 
would you be open to merging this if I open a new MR with the build fixed?

allauth/account/adapter.py
```diff
@@ -492,6 +493,7 @@ def authentication_failed(self, request, **credentials):
         dt = timezone.now()
         data.append(time.mktime(dt.timetuple()))
         cache.set(cache_key, data, app_settings.LOGIN_ATTEMPTS_TIMEOUT)
+        user_auth_failed.send(sender=self.__class__, request=request)
```
pennersr: A few thoughts:

- A signal is not needed per se, you can simply provide your own adapter and override the authentication_failed() method.

- If a signal is provided, it would be more consistent to name it after the adapter method (so authentication_failed, not user_auth_failed).

- To be truly useful, you would need to know details about the authentication attempt that failed -- as in, pass along the credentials into the signal.

PR #1938, opened 2018-02-02
Fix email on Eventbrite provider

trecouvr:
Hi,

I am using the `EventbriteProvider` on the new version of my application (https://www.pygmento.com) and I noticed it does not quite well handle the email field. I updated to provider to match the `GoogleProvider` mechanics.

- the provider returns a dict with a field `email` for `extract_common_fields`, using the primary address returned by Eventbrite
- the provider implements the method `extract_email_addresses`


pennersr:
Fixed via 16a5221f64384c95d7de43291a92c72b1e0aebb1

Issue #915, opened 2015-04-15
linkedin_oauth2 response changed


amyth:
Hey,

I was trying to setup linkedin oauth2 social login using "Django-Allauth" and "[Django-rest-auth](https://github.com/Tivix/django-rest-auth)" and for some reason whenever I try to verify a wrong access token, Instead of getting the error response I receive a 500 caused by a `KeyError` on `"id"` in the method `extract_uid`. Following is the traceback:

```
Traceback:
File "/Users/amyth/.virtualenvs/hungry/lib/python2.7/site-packages/django/core/handlers/base.py" in get_response
  132.                     response = wrapped_callback(request, *callback_args, **callback_kwargs)
File "/Users/amyth/.virtualenvs/hungry/lib/python2.7/site-packages/django/views/decorators/csrf.py" in wrapped_view
  58.         return view_func(*args, **kwargs)
File "/Users/amyth/.virtualenvs/hungry/lib/python2.7/site-packages/django/views/generic/base.py" in view
  71.             return self.dispatch(request, *args, **kwargs)
File "/Users/amyth/.virtualenvs/hungry/lib/python2.7/site-packages/rest_framework/views.py" in dispatch
  452.             response = self.handle_exception(exc)
File "/Users/amyth/.virtualenvs/hungry/lib/python2.7/site-packages/rest_framework/views.py" in dispatch
  449.             response = handler(request, *args, **kwargs)
File "/Users/amyth/.virtualenvs/hungry/lib/python2.7/site-packages/rest_auth/views.py" in post
  51.         if not self.serializer.is_valid():
File "/Users/amyth/.virtualenvs/hungry/lib/python2.7/site-packages/rest_framework/serializers.py" in is_valid
  187.                 self._validated_data = self.run_validation(self.initial_data)
File "/Users/amyth/.virtualenvs/hungry/lib/python2.7/site-packages/rest_framework/serializers.py" in run_validation
  370.             value = self.validate(value)
File "/Users/amyth/.virtualenvs/hungry/lib/python2.7/site-packages/rest_auth/registration/serializers.py" in validate
  30.                                                 response=access_token)
File "/Users/amyth/.virtualenvs/hungry/lib/python2.7/site-packages/allauth/socialaccount/providers/linkedin_oauth2/views.py" in complete_login
  21.         return self.get_provider().sociallogin_from_response(request, extra_data)
[...]

PR #1956, opened 2018-02-21
FIX Linkedin provider HTTP requests format to API

Brunux:
This related to issue #915 . Looks like linkedin API calls has change as mention in https://developer.linkedin.com/docs/oauth2  step4. I Added the right headers. Now it works.


Brunux:
Hi @pennersr was the conflict the improvement needed? if not would you give me more feedback, I'll be happy to improve it.

PR #1974, opened 2018-03-08
Quickbooks custom Provider

tjcrowley:
This is a custom provider I wrote to do oauth2 auth with Intuit Quickbooks and Payments APIs. 

It requires that you set a QBO_TEST variable in your settings file. If it's true, it uses the sandbox URL. If not, it assume production and tries that URL. In my system this file is at config.settings.base, probably need help making that more generic. But it works.

### DIFF
--- a/allauth/socialaccount/providers/quickbooks/provider.py
+++ b/allauth/socialaccount/providers/quickbooks/provider.py
@@ -45,7 +45,11 @@ class QuickBooksOAuth2Provider(OAuth2Provider):
         return fields
 
     def get_default_scope(self):
-        scope = ['openid', 'com.intuit.quickbooks.accounting', 'profile', 'phone']
+        scope = ['openid',
+                 'com.intuit.quickbooks.accounting',
+                 'profile',
+                 'phone',
+                 ]
         if app_settings.QUERY_EMAIL:
             scope.append('email')
         return scope
--- a/allauth/socialaccount/providers/quickbooks/provider.py
+++ b/allauth/socialaccount/providers/quickbooks/provider.py
@@ -17,6 +17,7 @@ class QuickBooksAccount(ProviderAccount):
             name = first_name + ' ' + last_name
         return name
 
+
 class QuickBooksOAuth2Provider(OAuth2Provider):
     id = 'quickbooks'
     # Name is displayed to ordinary users -- don't include protocol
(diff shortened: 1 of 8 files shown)


pennersr:
You can have a look at the base provider class. It features a method `get_settings`. This basically allows you to organize the provider settings, like so:

```
SOCIALACCOUNT_PROVIDERS = { 'quickbooks': { 'SANDBOX': True } }
```
Can you make use of that? Also, please add the snippet above in the `providers.rst` so that others know how to enable the sandbox mode.

tjcrowley:
Thanks, I will push a fix later today!

allauth/socialaccount/providers/quickbooks/provider.py
```diff
@@ -0,0 +1,60 @@
...
+        first_name = self.account.extra_data.get('givenName', None)
+        last_name = self.account.extra_data.get('familyName', None)
+        if first_name and last_name:
+            name = first_name + ' ' + last_name
+        return name
+
+class QuickbooksOAuth2Provider(OAuth2Provider):
+    id = 'quickbooks'
+    # Name is displayed to ordinary users -- don't include protocol
+    name = 'Quickbooks'
```
pennersr: Sorry for nitpicking, can you name this one "QuickBooks" as that seems to be the official style. Please rename the classes as well (QuickbooksAccount -> QuickBooksAccount). Thanks.

allauth/socialaccount/providers/quickbooks/provider.py
```diff
@@ -0,0 +1,60 @@
+from allauth.socialaccount import app_settings
```
pennersr: Can you add this provider to `test_settings.py`, `overview.rst` (list of supported providers) and `installation.rst` ?

allauth/socialaccount/providers/quickbooks/views.py
```diff
@@ -32,7 +32,9 @@ def complete_login(self, request, app, token, **kwargs):
     def get_user_info(self, token):
         auth_header = 'Bearer ' + token.token
         headers = {'Accept': 'application/json', 'Authorization': auth_header, 'accept': 'application/json'}
-        if app_settings.PROVIDERS.quickbooks.SANDBOX:
+        QBO_sandbox = getattr(app_settings, 'SOCIALACCOUNT_PROVIDERS', {}).get(
```
pennersr: You can replace this with self.get_provider().get_settings().get('SANDBOX', False)

tjcrowley: Just figured this out from looking at other providers, thank you!

tox.ini
```diff
@@ -32,6 +32,7 @@ commands =
 
 [testenv:flake8]
 skip_install = True
+max-line-length=90
```
pennersr: I am actually quite fond of pep8 and its line length limitations. Normally there is always a way to reduce the line length, and should you really have a case where readability is in danger, add a `# noqa` at the end of the line.

tjcrowley: Slowly figuring this out, I've never used Travis before. Thanks for the guidance.

PR #1998, opened 2018-04-06
Fix a serious error in German translation

phihag:
The German wording for "you must enter the same email twice" was "you must enter the same password twice", which can be really confusing for users.
This should probably be fixed on Transifex as well, but my request to join the de team on Transifex has not been answered yet.


phihag:
Thank you very much for accepting this! When can we expect a new release? I'd very much like to be able to depend on a new version rather than monkey-patching or similar atrocities.

Issue #1911, opened 2017-12-18
Duplicate Key Value Error When User Creates Account Twice


jplehmann:
I've been running AllAuth for more than a year now in a production environment (Django 1.8), and this error occurs for about 1.5% of the signups that I get.  I was able to upgrade to a recent version (0.32) to make sure this issue wasn't fixed before reporting it. `IntegrityError: duplicate key value violates unique constraint...DETAIL:  Key (provider, uid)=(facebook, 1234...) already exists.` 

Until this week I was unable to reproduce the error but I came about a sequence that seems to be what happens:
* Click Sign up
* Click Facebook logo to connect account
* Fill in username/confirm email
* Account is successfully created and presumably shows verification email sent page. (And verification email is successfully sent).
* *User enters a second username and email combination and clicks submit* (presumably after backing up or perhaps after browser never going to the verification page)
* In my Database, I can see the second account sometimes created in the same minute as the first one, so I know these come in quick succession.

I do not exactly know why the users are choosing to enter a second username/password combination, but one theory is they get a second though about the email they put in and decide they need it sent to another email address so they go back and change it.  Or, for some reason their browser never takes them to the verification page. I have emailed several users to try to get more info on what is happening to them. 

Other settings: 
* `ACCOUNT_EMAIL_REQUIRED=True`
* `SOCIALACCOUNT_AUTO_SIGNUP=False`
* This last week I converted my usernames to the preferred `ACCOUNT_PRESERVE_USERNAME_CASING=False` so that is in no way related.
* This error is the same as #411 but I'm only running a single PG database, so cause seems unrelated.

Questions:
* This "flow" seems to be totally allowed in the system for years, leading to the error.  Why am I the only one who has ever reported this problem?
* I am trying to find out the underlying cause why users do this.
* What behavior should `allauth` exhibit in this situation?  If nothing else, I think it should catch the integrity error, and report to the user that their Social account has already been assigned to another user, and to contact the admin if they need it reassigned (this is obviously far from ideal).

Full stack:
```
Internal Server Error: /accounts/social/signup/
Traceback (most recent call last):
  File "/app/.heroku/python/lib/python2.7/site-packages/django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/app/.heroku/python/lib/python2.7/site-packages/newrelic-2.60.0.46/newrelic/hooks/framework_django.py", line 499, in wrapper
[...]

PR #2001, opened 2018-04-10
Test case for #1911: social account taken.

jplehmann:
See comments in the PR (or issue #1911) for explanation of test case and suggested solutions to this problem.

### DIFF
--- a/allauth/socialaccount/tests.py
+++ b/allauth/socialaccount/tests.py
@@ -587,3 +587,68 @@ class SocialAccountTests(TestCase):
             'An account already exists with this e-mail address.'
             ' Please sign in to that account first, then connect'
             ' your Google account.')
+
+    @override_settings(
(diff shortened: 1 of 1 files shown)


pennersr:
Thanks for going through the trouble and providing the test!

I do not think it is the best way forward to introduce special logic inside the social signup view to recognize and handle this particular case. As you put it, the user "realizes that email address is unusable for them, and so backs up". But, that realization can also come at a time the session is gone, meaning there is no way to back up at all. So, IMHO, the social signup view is not the proper place to address this user realization as whatever we do it would not fully cover the scenario anyway.

For now, I will integrate your test and make a minor fix to pop the stashed social login once it is used. That avoids the crash.

pennersr:
825920d9f0e5461b3ebb29f8e84f530d0fd98635

PR #2006, opened 2018-04-17
Updating 23andMe endpoints.

owen9825:
Old URLs no longer return responses.


owen9825:
Fyi I've tested this at bevrly.com. The signup is still restricted though, so others won't be able to try it there.

PR #2007, opened 2018-04-17
when changing password distinguish between empty password and incorrect password being given

beda42:
Hi, 

At present, when the users changes password and inputs an invalid old password, he is not notified about it and instead gets a generic message "Please type your current password."

This pull request adds a more user-friendly error message for this case.

### DIFF
--- a/allauth/account/forms.py
+++ b/allauth/account/forms.py
@@ -462,9 +462,13 @@ class ChangePasswordForm(PasswordVerificationMixin, UserForm):
         self.fields['password1'].user = self.user
 
     def clean_oldpassword(self):
-        if not self.user.check_password(self.cleaned_data.get("oldpassword")):
+        old_password = self.cleaned_data.get("oldpassword")
+        if not old_password:
             raise forms.ValidationError(_("Please type your current"
                                           " password."))
+        if not self.user.check_password(old_password):
+            raise forms.ValidationError(_("The provided password is "
+                                          "not valid."))
         return self.cleaned_data["oldpassword"]
 
     def save(self):


allauth/account/forms.py
```diff
@@ -462,9 +462,13 @@ def __init__(self, *args, **kwargs):
         self.fields['password1'].user = self.user
 
     def clean_oldpassword(self):
-        if not self.user.check_password(self.cleaned_data.get("oldpassword")):
+        old_password = self.cleaned_data.get("oldpassword")
+        if not old_password:
             raise forms.ValidationError(_("Please type your current"
                                           " password."))
+        if not self.user.check_password(old_password):
+            raise forms.ValidationError(_("The provided password is "
```
pennersr: Sorry for nitpicking on details. Elsewhere, this style is used:

"The username and/or password you specified are not correct."

Can you change this error message to:

"The password you specified is not correct."

beda42: Sorry for the delay. I am quite new to this stuff - is there a way to make the suggested changes in the merge request directly or is it better to create a new merge request from a different branch altogether?

varunsaral:
@pennersr can I do it?

pennersr:
@varunsaral Sure, but wondering if it is really necessary to introduce new texts for this. For one, the the password field should be marked as "required" meaning the browser won't POST the form. And in case it stll does happen, I would just make sure this Django error pops up:

```
class Field:
    default_error_messages = {
        "required": _("This field is required."),
    }
```

... there really is no need add a custom string for this, that only increases translation burden.

varunsaral:
Aggreed @pennersr, we can close this.

PR #2011, opened 2018-04-18
delete unverified accounts if verification is mandatory

makerGeek:
Before this commit, when a user creates an account and doesn't verify the email that account stays registered and the email can't be used again (for example by its rightful owner).
This patch deletes the accounts with unverified emails **when ACCOUNT_EMAIL_VERIFICATION== "mandatory"** because it's senseless to keep an unverified account when it's useless.

### DIFF
--- a/allauth/utils.py
+++ b/allauth/utils.py
@@ -122,9 +122,23 @@ def valid_email_or_none(email):
     return ret
 
 
+def remove_email_if_unverified(email):
+    users_with_this_email = get_user_model().objects.filter(email=email)
+    if users_with_this_email.exists():
+        for user in users_with_this_email:
+            for emailaddress in user.emailaddress_set.all():
+                if not emailaddress.verified:
+                    emailaddress.delete()
+            if not user.emailaddress_set.all():
+                user.delete()
+
+
 def email_address_exists(email, exclude_user=None):
     from .account import app_settings as account_settings
     from .account.models import EmailAddress
+
+    if account_settings.EMAIL_VERIFICATION == account_settings.EmailVerificationMethod.MANDATORY:
+        remove_email_if_unverified(email)
 
     emailaddresses = EmailAddress.objects
     if exclude_user:


pennersr:
I am not really in favor of taking this direction. With these changes, a relatively innocent method to check if an email already exists becomes a potential problem as the side effect of calling it is to delete a complete user account.

9mido:
If you want to delete unverified accounts you can use a custom management command like so:

Settings.py:

```
#Recommended to avoid deleting users with more than one email address
ACCOUNT_MAX_EMAIL_ADDRESSES = 1
```

```
#Command: python3 manage.py file
from django.core.management.base import BaseCommand

import datetime
from django.utils import timezone
import pytz

from allauth.account.models import EmailAddress
from django.contrib.auth.models import User

class Command(BaseCommand):
    help = 'Delete users who have unverified email addresses that are between 1-2 days old'

    def handle(self, *args, **kwargs):
        recent_joined_users = User.objects.filter(date_joined__gt=timezone.now() - timezone.timedelta(days=2), date_joined__lt=timezone.now() - timezone.timedelta(days=1))
        for x in recent_joined_users:
            EmailAddress.objects.filter(user=x, verified=False).delete()
```

This example will delete unverified users who are unverified 1-2 days after signing up. But you could customize the query days=number portion to be however long you want to wait before deleting new unverified users.

9mido:
If you want to delete users with unverified emails you can use a custom management command:

Settings.py:
```
#Recommended to prevent deleting users who have more than one email address
ACCOUNT_MAX_EMAIL_ADDRESSES = 1
```

```
#Command: python3 manage.py file

from django.core.management.base import BaseCommand

import datetime
from django.utils import timezone
import pytz

from allauth.account.models import EmailAddress
from django.contrib.auth.models import User

class Command(BaseCommand):
    help = 'Delete users who have unverified email addresses that are between 1-2 days old'

    def handle(self, *args, **kwargs):
        recent_joined_users = User.objects.filter(date_joined__gt=timezone.now() - timezone.timedelta(days=2), date_joined__lt=timezone.now() - timezone.timedelta(days=1))
        for x in recent_joined_users:
            emails = EmailAddress.objects.filter(user=x, verified=False).values_list('user', flat=True)
            for y in emails:
                User.objects.get(id = y).delete()
```

The code above will delete unverified users who are created between 1-2 days ago. You can customize the days=number in the query to be whatever you want.

PR #2026, opened 2018-05-20
Bugfix: UUIDField pk - allow confirm email login

asfaltboy:
Prior to this fix, a user that used a UUIDField primary key would not be logged in due to the failing comparison of user pk with the "stashed" pk.


pennersr:
Can you please rebase?

asfaltboy:
Sure thing, check it out now

PR #2038, opened 2018-06-01
Add the ability to have maximum length for username

petersobhi:
I read this [question](https://stackoverflow.com/q/50548685/6086865) on Stack Overflow about customizing the maximum length of the `username` field in the signup form. I found it frustrating that you have to customize the `User` model and change the `username` field `max_length` or to override the `SignupForm` to do such a simple thing.

I added a new setting: `ACCOUNT_USERNAME_MAX_LENGTH` that allows customizing the username's maximum length easily without making any conflicts with the maximum length defined in the `User` model.

`ACCOUNT_USERNAME_MAX_LENGTH` will be only used if its value is less than the `max_length` defined in the `username` field of the `User` model.

### DIFF
--- a/allauth/account/app_settings.py
+++ b/allauth/account/app_settings.py
@@ -176,6 +176,13 @@ class AppSettings(object):
         return self._setting("USERNAME_MIN_LENGTH", 1)
 
     @property
+    def USERNAME_MAX_LENGTH(self):
+        """
+        Maximum username Length
+        """
+        return self._setting("USERNAME_MAX_LENGTH", 150)
+
+    @property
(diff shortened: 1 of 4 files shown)


gajeshbhat:
There is no need to customize user model. This can be achieved by extending `DefaultAccountAdapter.`
I have illustrated an example [here](https://stackoverflow.com/a/50941366/6206961) in a StackOverflow question.

pennersr:
A setting for this is not really in order. First, it would work only one way (having a shorter limit than your actual database model allows), and, as @gajeshbhat indicates it can be easily accomplished without altering allauth.

PR #2054, opened 2018-07-02
fix(signup): form autofocus use setting logic

afaur:
Login forms currently autofocus based on the `ACCOUNT_AUTHENTICATION_METHOD` setting.
The Signup form is currently not making a determination based on this. 

- Login Form Logic: https://github.com/pennersr/django-allauth/blob/330bf899dd77046fd0510221f3c12e69eb2bc64d/allauth/account/forms.py#L110-L131
- Docs: https://django-allauth.readthedocs.io/en/latest/configuration.html?highlight=ACCOUNT_AUTHENTICATION_METHOD

The signup form field order places email before username. 

- Signup Form Field Order Logic: https://github.com/pennersr/django-allauth/blob/330bf899dd77046fd0510221f3c12e69eb2bc64d/allauth/account/forms.py#L283-L289

In the case of `ACCOUNT_AUTHENTICATION_METHOD = 'username_email'` the email field would display first but autofocus would be on the username field.

This PR addresses this problem by allowing the correct field to obtain autofocus based on your settings.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -119,3 +119,4 @@ Will Ross
 William Li
 Yuri Kriachko
 Yaroslav Muravsky
+Adam Faur
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -11,6 +11,8 @@ Note worthy changes
 
 - New providers: Disqus.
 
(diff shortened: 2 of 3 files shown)


pennersr:
Autofocus has now been drop altogether (b9ca281c)

PR #2062, opened 2018-07-11
fix(baidu): Fixed baidu's get_profile_url bug.

seieewkx:
Old: The function always return fixed url `https://openapi.baidu.com//rest/2.0/passport/users/getLoggedInUser`. This is not the correct url point to user's profile page.

New: `https://www.baidu.com/p/<uname>/` Is the correct user's profile page. Like https://www.baidu.com/p/seieewkx.


seieewkx:
Could anyone review my PR?

PR #2067, opened 2018-07-25
Fix a translation error in Chinese translation

JayTurnr:
There's a bug in translation which is preventing compilemessages from running


JayTurnr:
It seems a few other languages have this issue. I'll submit a PR within the week hopefully fixing any broken references.

PR #2107, opened 2018-09-10
This should fix #2106

### LINKED ISSUE #2106: ACCOUNT_LOGIN_ON_EMAIL_CONFIRMATION does not log the user in after confirmation since 0.37
First of all I'd like to thank you and the maintainers for the hard work you're been doing making it easy for me to add social login. 

I just upgraded to `django-allauth==0.37.1` and noticed that users are no longer signed in after confirming their email address, with the `ACCOUNT_LOGIN_ON_EMAIL_CONFIRMATION` option set. Instead I end up with the sign-in page, which I was trying to prevent. It used to work on 0.36.

I browsed through the code and found that the `login_on_confirm` expects the user id to be stashed prior to logging the user in (https://github.com/pennersr/django-allauth/blob/0.37.0/allauth/account/views.py#L326). I did not find any reference in the confirmation view that actually stashes the user id. I worked around it by adding the stashing code in the adapter which is specific for the project I'm working on:

```
    def confirm_email(self, request, email_address):

        self.stash_user(
            request,
            user_pk_to_url_str(
                email_address.user
            )
        )
        return super(InviteAccountAdapter, self).confirm_email(request, email_address)
```

Is this correct and intentional? I can think of a PR which eliminates the need for stashing in the custom adapter. I'll be happy to create it.

bastbnl:
# Submitting Pull Requests

## General

 - [ ] Make sure yo use [semantic commit messages](https://seesparkbox.com/foundry/semantic_commit_messages). 
 - [ ] All Python code must be 100% pep8 and isort clean.
 - [ ] Javascript code should adhere to [StandardJS](https://standardjs.com).
 
 ## Provider Specifics
 
- [ ] Add an entry to the list of supported providers over at `docs/overview.rst`

### DIFF
--- a/allauth/account/views.py
+++ b/allauth/account/views.py
@@ -36,6 +36,7 @@ from .utils import (
     passthrough_next_redirect_url,
     perform_login,
     sync_user_email_addresses,
+    user_pk_to_url_str,
     url_str_to_user_pk,
 )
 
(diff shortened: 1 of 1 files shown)


dgilge:
Can this one be merged?

PR #2139, opened 2018-10-26
OpenID refactor to make way for a Steam provider

jleclanche:
This took me a while! I wanted to implement a Steam provider. Steam is OpenID, but gives no access to even the most basic things (no username!). That's all behind a custom API which requires its own key.

To implement Steam, I wanted to do this right, but there were no other instances of a custom OpenID-based provider.
More specifically, to avoid code duplication, I had to reimplement the OpenID login and callback views into class-based views. They are now much more extensible and I make heavy use of that in the Steam provider.

Tested in production.

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -1,3 +1,12 @@
+0.39.0 (Unreleased)
+*******************
+
+Note worthy changes
+-------------------
+
+- New provider: Steam (OpenID)
+
+
 0.38.0 (2018-10-03)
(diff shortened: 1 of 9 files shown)


jleclanche:
@pennersr ping, can I get feedback on this please

jleclanche:
Rebased against master and added a tiny docs fix.

jleclanche:
@pennersr new ping?

PR #2158, opened 2018-12-03
Email verification required as a Mixins

simo97:
feat(account): Added a new mixins to  enable email verification control on a class-based views.

### DIFF
--- a/allauth/account/mixins.py
+++ b/allauth/account/mixins.py
@@ -9,9 +9,9 @@ Example of use :
 view.py
 
 from django.views.generic import ListView
-from allauth.account.mixins import VerifiedEmailRequired
+from allauth.account.mixins import VerifiedEmailRequiredMixin
 
-class IndexView(VerifiedEmailRequired, ListView):
+class IndexView(VerifiedEmailRequiredMixin, ListView):
     ...
 """
 
@@ -22,7 +22,7 @@ from .models import EmailAddress
 from .utils import send_email_confirmation
 
 
-class VerifiedEmailRequired(LoginRequiredMixin):
+class VerifiedEmailRequiredMixin(LoginRequiredMixin):
     """
     This class is a version of the 
     allauth.account.decorators.verified_email_required() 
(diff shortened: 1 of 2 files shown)


simo97:
Oh ok will change that quickly.

allauth/account/mixins.py
```diff
@@ -0,0 +1,39 @@
+"""
+This module contain a class mixin design to bring email verification
+to class based views in django-allauth
+It is used like django's LoginRequiedMixin, so when using this
```
pennersr, line 4:
```suggestion
It is used like django's LoginRequiredMixin, so when using this
```

allauth/account/mixins.py
```diff
@@ -0,0 +1,39 @@
...
+there is no need to use LoginRequiedMixin fro django
```
pennersr, line 5:
```suggestion
there is no need to use LoginRequiredMixin fro django
```

docs/decorators.rst
```diff
@@ -24,3 +24,20 @@ The behavior is as follows:
...
+
+Mixins
+======
+
+Verified E-mail Required
+------------------------
+
+In the case you are using a class-based view, there is a mixin available
+in ``allauth.account.mixins.VerifiedEmailRequiredMixin`` which is use in 
+any class-based view from django, you can use it as follow ::
```
pennersr, line 36:
```suggestion
any class-based view from django, you can use it as follows ::
```

docs/decorators.rst
```diff
@@ -24,3 +24,20 @@ The behavior is as follows:
...
+
+
+    from django.views.generics import ListView
+    from allauth.account.mixins import VerifiedEmailRequiredMixin
+
+    class YouClassName(VerifiedEmailRequiredMixin, ListView):
```
pennersr, line 42:
```suggestion
    class YourClassName(VerifiedEmailRequiredMixin, ListView):
```

pennersr (changes requested):

allauth/account/mixins.py
```diff
@@ -0,0 +1,39 @@
+"""
+This module contain a class mixin design to bring email verification
+to class based views in django-allauth
+It is used like django's LoginRequiredMixin, so when using this
+there is no need to use LoginRequiredMixin fro django
```
pennersr, line 5:
```suggestion
there is no need to use LoginRequiredMixin from Django
```

allauth/account/mixins.py
```diff
@@ -0,0 +1,39 @@
+"""
+This module contain a class mixin design to bring email verification
+to class based views in django-allauth
+It is used like django's LoginRequiredMixin, so when using this
```
pennersr, line 4:
```suggestion
It is used like Django's LoginRequiredMixin, so when using this
```

pennersr (changes requested):

PR #2159, opened 2018-12-04
Fixed potentially faulty if statement

Jasper-Koops:
The ```PRESERVE_USERNAME_CASING``` - when True - ran  a case insensitive search on username. 
I propose changing the if statement so that such a search would only be done when ```PRESERVE_USERNAME_CASING``` is False

This potential bug was discovered when allauth started throwing errors in our app where both  'Bob' and 'bob'  would throw a 
``` MultipleObjectsReturned: get() returned more than one User -- it returned 2! ``` error during a failed login.

### DIFF
--- a/allauth/account/utils.py
+++ b/allauth/account/utils.py
@@ -357,9 +357,9 @@ def sync_user_email_addresses(user):
 
 
 def filter_users_by_username(*username):
-    if not app_settings.PRESERVE_USERNAME_CASING:
+    if app_settings.PRESERVE_USERNAME_CASING:
         qlist = [
-            Q(**{app_settings.USER_MODEL_USERNAME_FIELD + '__iexact': u})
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by pennersr** [comment]
Please check the documentation:

```
ACCOUNT_PRESERVE_USERNAME_CASING (=True)
  This setting determines whether the username is stored in lowercase
  (``False``) or whether its casing is to be preserved (``True``). Note that when
  casing is preserved, potentially expensive ``__iexact`` lookups are performed
  when filter on username. For now, the default is set to ``True`` to maintain
  backwards compatibility.
```

When preserve casing is false, Bob is turned into bob, and then saved. So, there is no need to do an iexact lookup. Wen preserve casing is true, Bob is stored as Bob, but any user attempting to claim boB as a username should fail, hence, an iexact is required.

What you are encountering is likely not an issue in allauth. Possibly, you created a user manually, or, you switched the value of the setting while users where already created.

Issue #2142, opened 2018-11-01
Network Login Failure error reporting is QUITE unhelpful


derek-adair:
Currently if my social account integration is not set up properly, I'm left having to dig into the code to debug.  It'd be a fantastic improvement for debugging if we could make this generic fail more helpful. 

- {{ auth_error }} returned in debug mode
- Exact errors reported via the error console/logs (even w/o debug=True)

I'm sure there are other improvements that can be made in this regard.

I'd be glad to dive in and get this done if no one else had the bandwidth.

derek-adair:
Specifically using twitter.  Which seems to NOT be related to #1120? unsure tho.

kov:
This is very frustrating and it is a bit surprising that it has been an issue for so long (the first issue I hit on my search was from 2014). I am working on a dockerized application so it's not very easy to debug things or make changes. Why not simply provide at least some detail in the logs at least so that you don't risk exposing something you shouldn't to users. It really shouldn't be that hard to do, right?

pennersr:
Dupe of #1120

derek-adair:
my bad dude.  Thanks for your work.

Would you be open to someone help you triage issues?

pennersr:
Sure, there is nothing preventing anyone from hopping onto an issue and doing the research.

derek-adair:
Sure, I'll try to comment where I can.  I meant in a more official capacity -- e.g. closing issues for you or escalating valid ones -- which would require us being on the same page with your ambitions and approach to open source.

pennersr:
https://codeberg.org/allauth/django-allauth/issues/2142

PR #2175, opened 2019-01-14
feat(oauth): Show an informative log when OAuthError is raised

ykdojo:
This is similar to the way [facebook/views.py](https://github.com/pennersr/django-allauth/blob/41f84f5530b75431cfd4cf2b89cd805ced009e7d/allauth/socialaccount/providers/facebook/views.py) logs an exception.

This feature was requested on #2142.

Example log:
`allauth.socialaccount.providers.oauth.client.OAuthError: Invalid response while obtaining request token from "api.twitter.com". Response content: <?xml version="1.0" encoding="UTF-8"?><errors><error code="415">Callback URL not approved for this client application. Approved callback URLs can be adjusted in your application settings</error></errors>`

### DIFF
--- a/allauth/socialaccount/providers/oauth/client.py
+++ b/allauth/socialaccount/providers/oauth/client.py
@@ -77,7 +77,9 @@ class OAuthClient(object):
                 raise OAuthError(
                     _('Invalid response while obtaining request token'
                       ' from "%s".') % get_token_prefix(
-                          self.request_token_url))
+                            self.request_token_url) +
+                    ' Response content: %s' % response.content.decode("utf-8")
+                )
             self.request_token = dict(parse_qsl(response.text))
             self.request.session['oauth_%s_request_token' % get_token_prefix(
                 self.request_token_url)] = self.request_token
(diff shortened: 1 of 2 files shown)


allauth/socialaccount/providers/oauth/client.py
```diff
@@ -77,7 +77,9 @@ def _get_request_token(self):
                 raise OAuthError(
                     _('Invalid response while obtaining request token'
                       ' from "%s".') % get_token_prefix(
-                          self.request_token_url))
+                            self.request_token_url) +
+                    ' Response content: %s' % response.content.decode("utf-8")
```
pennersr: content need not be in utf-8 format, you should simply use response.text to have the decoding transparently handled for you.

rraihansaputra:
hi @ykdojo, are you interested to continue working on this PR? This is an improvement that is sorely needed. It would be a shame for your work to not be continued with a bit of correction..

allauth/socialaccount/providers/oauth/client.py
ykdojo: Thank you for the feedback

Issue #345, opened 2013-07-31
How to disable form login/signup?


fjcaetano:
Hey guys, first of all, thank you for this app. It's amazing and it has already saved me a bunch of time.

I have the following problem because I couldn't find anything in the docs, so I thought it would be better to ask.

How can I completely disable form login/signup and enable only the providers I set? In my business model, the user can only login/signup with Facebook and, whenever I use the @login_required decorator, it redirects the user to a login page where he can input emal/password.

Thanks in advance.

markunsworth:
There's a potential security loophole here. 

If you setup allauth to use a social auth provider as the signup/signin process. 

Then use a custom SocialAccountAdapter to check the domain of email address to remove the need for email confirmation  (which was set to optional) to remove friction from the proces. You can do this with the `pre_social_login` method of `DefaultSocialAccountAdapter`

As the DefaultAccountAdapter always returns `True` from the `is_open_for_signup` method, someone can just hit the `/accounts/signup/` page and create an account which then doesn't need to be verified. 

Whilst this is obviously down to the developer to configure correctly, it's not so clear that this would be the case. 

I think adding a configuration option to turn on/off the standard account signup page would be sensible and setting everything to off by default might also be wise, then developers can just switch on what they need.

wetneb:
I support this feature request! :-)

aliva:
I believe direct sign up, should be defined as another optional backends

aliva:
here is my adapters as @pennersr mentioned in his comment:

```python
class AccountAdapter(DefaultAccountAdapter):
    def is_open_for_signup(self, request):
        if request.path.rstrip("/") == reverse("account_signup").rstrip("/"):
            return False
        return True


class SocialAccountAdapter(DefaultSocialAccountAdapter):
    def validate_disconnect(self, account, accounts):
        raise ValidationError("Can not disconnect")
```

but there is another problem here, It's with `EmailView` (accounts/email) in this view users can add/remove email addresses to their account - I think if you want completly use providers (disable direct login/singup) email management may cause some problems

also `password` mangement doesn't make sense. (set password/change password/reset password)

bittner:
I too think configuring social login + registration _only_ should be made easier. I'd even prefer that the `Account` models/tables were not populated/created at all. In fact, let's [look at the setup](https://django-allauth.readthedocs.io/en/latest/installation.html):
``` python
INSTALLED_APPS = [
    # ...
    'allauth',
    'allauth.account',
[...]

PR #2181, opened 2019-01-16
feat: ACCOUNT_SIGNUP_OPEN and ACCOUNT_NEW_ACCOUNTS_ACTIVE settings

olivierdalang:
This PR adds two settings, allowing to toggle `is_open_for_signup` (allowing to easily close new registrations) and `user.is_active` (allowing for manual activation of new accounts by admins).

I guess it's quite common to need to switch those, and subclassing `allauth.account.adapter.DefaultAccountAdapter` just for that is a bit cumbersome.

If you like the feature, I can write some tests.


wgordon17:
It might also be beneficial to apply the `SIGNUP_OPEN` logic to https://github.com/pennersr/django-allauth/blob/4a9d07c512d0247f940262038b7258ea986e1e16/allauth/socialaccount/adapter.py#L176 as well.

olivierdalang:
@wgordon17 is this (last commit) what you meant ?

Is this really wanted ? (I'm asking as to me it seems it makes it more complicated to change the behaviour, since subclassing `DefaultAccountAdapter` would not be enough anymore...).

If this is ok, please advise of next steps to get this merged.

Thanks !!

wgordon17:
I apologize...it's been way too long to remember what on earth I was thinking 😂

But, I _think_ something along the lines of another option, `SOCIALACCOUNT_SIGNUP_OPEN`, was what I had in mind. You could default it to None, and then do something like

```python
    def is_open_for_signup(self, request, sociallogin):
        """
        Checks whether or not the site is open for signups.
        Next to simply returning True/False you can also intervene the
        regular flow by raising an ImmediateHttpResponse
        """
        if app_settings.SOCIALACCOUNT_SIGNUP_OPEN is None:
            return get_account_adapter(request).is_open_for_signup(request)
        else:
            return app_settings.SOCIALACCOUNT_SIGNUP_OPEN
```

This would at least give you the option to control socialaccount signups separately, otherwise it would just default to the current behavior...that way it doesn't affect any legacy expectations.

wgordon17:
I'm also not a maintainer of this project...so I'm not sure what would be needed to get it merged 😞

olivierdalang:
> I apologize...it's been way too long to remember what on earth I was thinking 😂
> 
> But, I _think_ something along the lines of another option, `SOCIALACCOUNT_SIGNUP_OPEN`, was what I had in mind. You could default it to None, and then do something like
> 
> ```python
>     def is_open_for_signup(self, request, sociallogin):
>         """
>         Checks whether or not the site is open for signups.
>         Next to simply returning True/False you can also intervene the
>         regular flow by raising an ImmediateHttpResponse
>         """
>         if app_settings.SOCIALACCOUNT_SIGNUP_OPEN is None:
>             return get_account_adapter(request).is_open_for_signup(request)
>         else:
>             return app_settings.SOCIALACCOUNT_SIGNUP_OPEN
> ```
> 
> This would at least give you the option to control socialaccount signups separately, otherwise it would just default to the current behavior...that way it doesn't affect any legacy expectations.

Ok I see what you mean. I find these two similar and interdependent settings make things a bit complicated/confusing, as most of the time, if you want to disable signup, you want to do it both for social accounts and for regular accounts. For now, I'll revert the commit and we'll see whatever the maintainer thinks (if he comes by at some point). Thanks for the detailed review though !

wgordon17:
Sounds good! 👍 My use case was that I actually wanted to rely on _only_ social account login...and it would be nice to have this controllable in settings...but subclassing the default Adapter was easy enough.

9mido:
Linking the following issue since it relates:

https://github.com/pennersr/django-allauth/issues/345

pennersr:
It's already easy enough to close down signup using the adapter. As the discussion in this PR shows, the changes here are also not fully trivial (e.g. the `True` / `False` / `None` issue). All in all, I don't think this is a problem we need to fix. Closing.

PR #2196, opened 2019-02-02
feat(signal): Add post auth signal during social auth after token request

iarp:
Some providers do not return all information during the callback and must be obtained through the authorization call using the code supplied.

There is currently no way to obtain that response information other than to re-create the provider as a custom provider and override the view where complete_login is called.

My instance of this is needing to confirm scopes were granted or not. Google returns the scopes at the same time as the auth code so I was able to use a pre_social_login signal which has the response object. Microsoft Graph and Windows Live only returns scopes in the auth response.

### DIFF
--- a/allauth/socialaccount/providers/oauth/views.py
+++ b/allauth/socialaccount/providers/oauth/views.py
@@ -2,7 +2,7 @@ from __future__ import absolute_import
 
 from django.urls import reverse
 
-from allauth.socialaccount import providers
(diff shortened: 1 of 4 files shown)


pennersr:
This signal is fired just before `complete_social_login` is called. Though, on of the first things that `complete_social_login` invokes is: `get_adapter(request).pre_social_login(request, sociallogin)`, so I think this use case is already covered by having that adapter method.

Issue #2039, opened 2018-06-05
ACCOUNT_FORMS['signup'] != ACCOUNT_SIGNUP_FORM_CLASS


dyve:
I needed a django-allauth signup form with a few extra fields. I'd done this a few times, and thought it wouldn't be a problem. There is also a good explanation by @pennersr here: https://stackoverflow.com/a/12308807/117831

Having noticed the new `ACCOUNT_FORMS` setting, I thought it woud be better to start using this instead of the `ACCOUNT_SIGNUP_FORM_CLASS` I'd used before.

```
# Old method in `settings.py`
ACCOUNT_SIGNUP_FORM_CLASS = 'app.users.forms.MySignUpForm'

# New method in `settings.py`
ACCOUNT_FORMS = {
    'signup': 'app.users.forms.MySignUpForm'
}
```

The form seemed to work, but only my extra fields showed up. After trying some stuff, I couldn't get the new method to work. I reverted to using `ACCOUNT_SIGNUP_FORM_CLASS`, and everything worked like a charm.

From the documentation, I gathered that both settings should provide the same results, but the documentation does not actually say this. Or I missed something.

I would like to know if these settings are supposed to do the same thing.

If yes, then there is a bug to be fixed.
If no, then the documentation needs to be made more explicit.

Thanks for a great job on this project!

gajeshbhat:
You can use the new `ACCOUNT_FORMS` method by extending the `SignupForm` class into your custom form class.
You can follow the below example.
```from allauth.account.forms import SignupForm
from django import forms

class SignupForm(SignupForm):
    first_name = forms.CharField(max_length=30, label='First Name')
    last_name = forms.CharField(max_length=30, label='Last Name')

    def signup(self, request, user):
        user.first_name = self.cleaned_data['first_name']
        user.last_name = self.cleaned_data['last_name']
        user.save()
        return user
```
`return user` in the above code is very important as it passes the user objects to other modules for other verification.
Finally, add the settings in `settings.py`
`ACCOUNT_FORMS = {'signup': 'YourProject.forms.SignupForm'}`

KrazyMax:
I am not sure this is what @dyve meant. 

Like him, I think having this new `ACCOUNT_FORMS` is way more explicit (makes more sense to override the `save` method of the form than create a custom `signup` one) and concised (having everything at the same place).

[...]

PR #2203, opened 2019-02-08
docs(forms): Add clarification to forms about inheriting base forms

iarp:
Updated documentation about forms and ACCOUNT_SIGNUP_FORM_CLASS to clarify how forms are loaded, unherited, or overridden. #2039

### DIFF
--- a/docs/forms.rst
+++ b/docs/forms.rst
@@ -1,16 +1,21 @@
-Forms
-=====
+Customizing Forms
+=================
 
-The following forms can be overridden as needed in order to:
+All forms are based on ``django.forms.Form`` class, any custom fields you add
+will require custom logic to save the data. Some forms do not explicitly call
+the built-in save method, below lists all forms used and the final method
+called where you would want to save your custom data.
 
-- Add extra fields for extra required information
-- Override save to add extra functionality on save
+It is recommended that you review the default form classes provided that you
+wish to extend or replace in order to understand the fields and logic already
+supplied.
 
-Overriding Save
----------------
+To add additional fields or override functionality, create a custom class and
+inherit from the matching form listed below to supply your customizations.
 
-If you decide to add fields to a form, you will need to
-manually save the custom fields' data.
+To replace a form completely, create a custom class and inherit
+from ``forms.Form``. You will need to reimplement the original forms methods
+you have overridden
 
 ACCOUNT_FORMS
 -------------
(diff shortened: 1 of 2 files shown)


docs/forms.rst
```diff
@@ -75,6 +80,10 @@ Used on `account_signup <views.html#signup-account-signup>`__ view.
             # You must return the original result.
             return user
 
+The configuration setting ``ACCOUNT_SIGNUP_FORM_CLASS`` is the same as
+inheriting the SignupForm class. It is recommended to inherit and use the
+setting below instead.
```
pennersr: Why would overriding the `signup` form complete be preferable over using ACCOUNT_SIGNUP_FORM_CLASS ?

docs/forms.rst
```diff
@@ -1,16 +1,21 @@
...
-Overriding Save
----------------
+To add additional fields or override functionality, create a custom class and
+inherit from the matching form listed below to supply your customizations.
 
-If you decide to add fields to a form, you will need to
-manually save the custom fields' data.
+To replace a form completely, create a custom class and inherit
+from ``forms.Form``. You will need to reimplement the original forms methods
+you have overridden
```
pennersr: First of, I appreciate the effort you spent in trying to up the documentation. But, I am wondering whether or not the above truly helps in explaining things:

- "All forms are based on django.forms.Form" -- that holds for any Django project with forms.

- "Custom fields require custom logic" -- true for any project with forms, does not convey much information.

- "... recommended to review" -- while true, from a documentation point of view, not very helpful.

iarp: Mentioning Form was to differentiate between Form and ModelForm types.

I suppose the rest was to inform someone that if they add custom fields, said data will not just automatically be saved. I had seen too many posts (not directly related to allauth) where its expected that data on the form is just magically saved somehow. That's more so with ModelForms than Form types.

docs/forms.rst
iarp: The project currently has 2 separate ways of supplying the signup form class, `ACCOUNT_SIGNUP_FORM_CLASS` and `ACCOUNT_FORMS['signup']`.

Is the single dict `ACCOUNT_FORMS = {...}` not a cleaner way of supplying the forms that are being overridden throughout the application. Now all forms that are overridden are in a single location in the settings file rather than potentially scattered and using different slightly naming schemes.

Issue #2235, opened 2019-03-11
Password value sent back on response when form invalid


gabriel-amram:
When using AJAX, the `DefaultAccountAdapter.ajax_response_form` adds each field's value to the response that will be sent back to the client when the form is invalid.

One of these fields might be sensitive - such as password, which is bad practice to send back to the client for security reasons.

It doesn't respect the `ACCOUNT_PASSWORD_INPUT_RENDER_VALUE` setting which would normally prevent the value from being rendered in the template.

pennersr:
Why is it bad to send it back to the client? The client already knows the value, and either the connection is encrypted or not. If not, then not sending it back won't address a security issue, and if it is encrypted, there is no issue.

pennersr:
https://codeberg.org/allauth/django-allauth/issues/2235

PR #2236, opened 2019-03-11
Issue 2235 password value sent back on response

gabriel-amram:
fixes issue #2235 
added attribute to `PasswordField` that keeps boolean of `render_value` and overrided `prepare_value` to respect the `render_value` and return `None` if value should not be rendered

### DIFF
--- a/allauth/account/forms.py
+++ b/allauth/account/forms.py
@@ -64,14 +64,20 @@ class PasswordVerificationMixin(object):
 
 
 class PasswordField(forms.CharField):
+    _render_value = False
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


varunsaral:
Should i complete this @pennersr ? I understood the scope, we have to return none in case render_value=false

pennersr:
See my comment here: https://github.com/pennersr/django-allauth/issues/2235#issuecomment-1694748622 -- I would like to see some arguments first why this is an issue.

varunsaral:
Can't think of any reason why it's bad. Maybe just a bad user experince when a user see his password again on screen afte something went wrong. Although here we can close it if it is not neccesary.

PR #2244, opened 2019-03-17
fix: updating the twitch provider to use new authentication API

MattBroach:
# What

Updates the `twitch` provider to user the new API's authentication framework.  This is still OAuth2-based, but uses a different set of authentication URLs and has completely changed the user data and content.

# Why

Although the previous API still technically works, the V5 API has been deprecated and could therefore be turned off at any time (it was actually supposed to be removed at the end of 2018, but, for now, has been granted a zombie status).  From Twitch's dev docs:

> Note: URLs have been updated to replace https://api.twitch.tv/kraken with https://id.twitch.tv. Code that uses the old kraken domain for Twitch authentication will continue to work until the removal of Twitch API v. 5 functionality. Twitch API v. 5 is currently deprecated.

# Caveats

The new API's post-login profile data is different in both structure and content than the old `V5` endpoint.  Although logins should function seamlessly when transitioning from the old API to the new one, any project that relies on data from 'SocialAccount.extra_data` should refer to the new API [user endpoint](https://dev.twitch.tv/docs/api/reference/#get-users) documentation for information on the new profile data format and content.

# Submitting Pull Requests

## General

 ## Provider Specifics
 

MattBroach:
I had to do a number of additional commits to deal with some lingering Travis errors.  Would the preference be for me to close this PR, squash the commits, and open a new PR with a single commit?  I'd be happy to do whatever makes the maintainers happiest.

pennersr:
Thanks. Can you add the "Caveats" to the ChangeLog.rst (Backwards incompatible changes) ?

MattBroach:
Changelog updated!

pennersr:
Merged via 72b6cd5

beheh:
Based on my tests the default scope needs to be either updated or removed. The email scope for the new endpoint is now `user:read:email`, but it's not required for just authenticating with Twitch. [List of scopes](https://dev.twitch.tv/docs/authentication/#scopes).

MattBroach:
Whoops, that's correct -- the old V5 scope didn't get updated.  I think my personal preference would be to add the updated `user:read:email` scope rather than omit it -- it's true that it isn't required, but I think most people building Django apps would want the access to the email address (for password resets and such).  But I'm not wedded to it...

beheh:
Yeah, right, it should probably be included, I think that's what most providers do. While we don't need it in our app we can always override the scope.

PR #2263, opened 2019-04-08
feat(discord): Added verified email detection for Discord provider

pnwparksfan:
Discord performs email verification, and indicates this in the extra data returned by their OAuth2 API. 

[Discord API documentation](https://discordapp.com/developers/docs/resources/user#user-object): 

image

I also tested this by looking at the API data returned for my own already-verified account, and looking at the API data returned for a new test account that I created with an email address entered but not verified. As you can see in the screenshot below, you are unable to add a phone number to the account or enable 2FA until after your email has been verified. 

image

The `extract_email_addresses` code is copied from the `google` adapter and modified just to change which key to `get` from the `data`.


pnwparksfan:
Looks like there are some issues with some of the sqlite tox configs for Django 2.2?

PR #2271, opened 2019-04-19
Handle user.email=None case when generating password token

Anton-Shutik:
When `user.email` equals `None` they're not able to reset their password because of a bug in `EmailAwarePasswordResetTokenGenerator`. This tiny PR fixes it.

People still able to reset their password even if they don't have an email address on django's `user`, but they have it on django allauth's `EmailsAddress`

[Commit](https://github.com/pennersr/django-allauth/commit/3fe584e1a3998e244e608d097ca4fececa77aa30#diff-3b9310d6e2c9a729f9c65c16647b3e1aR47) where the issue was introduced

Thanks

### DIFF
--- a/allauth/account/forms.py
+++ b/allauth/account/forms.py
@@ -39,7 +39,7 @@ class EmailAwarePasswordResetTokenGenerator(PasswordResetTokenGenerator):
             EmailAwarePasswordResetTokenGenerator, self)._make_hash_value(
                 user, timestamp)
         sync_user_email_addresses(user)
-        emails = set([user.email])
+        emails = set([user.email] if user.email else [])
         emails.update(
             EmailAddress.objects
             .filter(user=user)


Anton-Shutik:
@pennersr Would you mind taking a look at the PR ?

al-the-x:
Let's add some tests to our fork, @Anton-Shutik.

al-the-x:
You called out the test that should fail in your comment on the commit: https://github.com/groveco/django-allauth/blob/master/allauth/account/tests.py#L318

Anton-Shutik:
@al-the-x Yep it will, but for another reason. They have `email` as required field on `User` model, we can't set it to `None`. I've [tried that](https://github.com/pennersr/django-allauth/pull/2271/commits/742078a10e6aa6d63e181b825c7df8890f9aaeea).
So probably we should leave this one-line PR without a test

allauth/account/forms.py
```diff
@@ -44,7 +44,7 @@ def _make_hash_value(self, user, timestamp):
             EmailAddress.objects
             .filter(user=user)
             .values_list('email', flat=True))
-        ret += '|'.join(sorted(emails))
+        ret += '|'.join(sorted((str(e) for e in emails)))
```
pennersr: I would rather see the issue addressed on this stop (instead of using the str() workaround):

 `emails = set([user.email] if user.email else [])`

Anton-Shutik: @pennersr  Done!

Anton-Shutik:
@pennersr Thanks a lot!
Is there any approximate release date of this ?

PR #2279, opened 2019-04-25
chore: ci: Travis: use dist=xenial

blueyed:
This is good in general (more recent), and likely to fix builds with
Django 2.2, which currently fail:

> django.core.exceptions.ImproperlyConfigured: SQLite 3.8.3 or later is
> required (found 3.8.2).


blueyed:
@pennersr 
Thanks for merging it so quickly - I'm glad CI is good again now.

PR #2281, opened 2019-04-25
fix: cases where the 'request' context item may not be present

wgordon17:
# Submitting Pull Requests

## General

 - [ X ] Make sure you use [semantic commit messages](https://seesparkbox.com/foundry/semantic_commit_messages).
 - [ X ] All Python code must be 100% pep8 and isort clean.
 - [ X ] Feel free to add yourself to `AUTHORS`.

### DIFF
--- a/allauth/templates/tests/test_403_csrf.html
+++ b/allauth/templates/tests/test_403_csrf.html
@@ -1,0 +1,2 @@
+{% load socialaccount %}
+<a href="{% provider_login_url "google" next=next %}" class="nav-link">Sign In</a>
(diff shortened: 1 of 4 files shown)


allauth/templates/tests/test_403_csrf.html
```diff
@@ -0,0 +1,86 @@
+{% load socialaccount %}
+<!DOCTYPE html>
```
pennersr: I don't think there is a need to have a full template in here. It could be a minimal template with just  the provider_login_url tag in there.

wgordon17:
@pennersr Would you mind re-reviewing?

Issue #1789, opened 2017-07-26
Harden the tip on protecting admin with allauth


ishitatsuyuki:
See https://stackoverflow.com/a/45328687/2484972 for details. The docs were added in #1562.

blueyed:
See https://github.com/pennersr/django-allauth/pull/2287 - needed an adjustment to the StackOverflow answer (see my comment there).

pennersr:
https://codeberg.org/allauth/django-allauth/issues/1789

PR #2287, opened 2019-04-26
docs: use staff_member_required to protect admin

blueyed:
Fixes https://github.com/pennersr/django-allauth/issues/1789.
Ref: https://stackoverflow.com/a/45328687/15690

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -219,10 +219,14 @@ every instance of ``AdminSite``):
 
 .. code-block:: python
 
+    from django.conf import settings
     from django.contrib import admin
-    from django.contrib.auth.decorators import login_required
+    from django.contrib.admin.views.decorators import staff_member_required
 
-    admin.site.login = login_required(admin.site.login)
+    admin.site.login = staff_member_required(
+        admin.site.login, login_url=settings.LOGIN_URL
+    )
+
 
 Customizing providers
 ---------------------


pennersr:
`staff_member_required` does not use the allauth login (and its brute force protection as documented in the paragraphs about your changes).

blueyed:
@pennersr 
Please re-check: regarding brute-force it is the other way around.  `login_required` will display the login form (for brute forcing).
And `login_url` is used to redirect through allauth.  That should be clarified probably, i.e. that `settings.LOGIN_URL` should be set to allauth's.

pennersr:
If you look here...

https://github.com/django/django/blob/b9cf764be62e77b4777b3a75ec256f6209a57671/django/contrib/admin/views/decorators.py

... you'll see that it does not use the regular login_required decorator, and redirects to `admin:login`. As a result, the issue warned about in the documentation is not tackled.

blueyed:
`login_url='admin:login'` is the default - this PR uses `settings.LOGIN_URL` there, i.e. allauth.
It checks if it is a staff user, which AnonymousUser is not.. ;)

blueyed:
(I have a test for this actually, so it is working as expected really)

blueyed:
Can we have this then, please? :)

9mido:
This seems similar to this:

https://github.com/pennersr/django-allauth/commit/90a2eb917d435c40f70867632994e4e35d8ac784

PR #2288, opened 2019-04-26
chore(tox): rename flake8 to checkqa

blueyed:
Also uses `ignore_errors = True` to keep going on errors, but still exit
non-zero if one command fails.


blueyed:
@pennersr 
Anything needed here?

PR #2299, opened 2019-05-13
race condition fix

Anton-Shutik:
## Refactored small piece of code in order to prevent possible race condition.

I'm having error like below:

image

Sql logs (blurred emails are the same on the screenshot):

image

We can take advantage of django's `get_or_create` and prevent possible race condition, which I'm getting (see logs).

### DIFF
--- a/allauth/account/utils.py
+++ b/allauth/account/utils.py
@@ -359,10 +359,14 @@ def sync_user_email_addresses(user):
                 and EmailAddress.objects.filter(email__iexact=email).exists():
             # Bail out
             return
-        EmailAddress.objects.create(user=user,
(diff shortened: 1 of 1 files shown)


blueyed:
LGTM in general, but wonder how the race gets triggered for you?

Also commits should be squashed probably, but don't know if that would/could be done when merging already.

Anton-Shutik:
Probably email was created in another request or smth

Anton-Shutik:
@blueyed sorry for long response. Removed merge commits, would you mind taking another look ?

pennersr:
Closed via ad258875

PR #2308, opened 2019-06-06
Add allauth.urls.provider_urlpatterns

blueyed:
This is useful to reuse them without the other urlpatterns.

### DIFF
--- a/allauth/urls.py
+++ b/allauth/urls.py
@@ -12,6 +12,8 @@ urlpatterns = [url(r'^', include('allauth.account.urls'))]
 if app_settings.SOCIALACCOUNT_ENABLED:
     urlpatterns += [url(r'^social/', include('allauth.socialaccount.urls'))]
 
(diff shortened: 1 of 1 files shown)


allauth/urls.py
```diff
@@ -12,11 +12,14 @@
...
     try:
         prov_mod = import_module(provider.get_package() + '.urls')
     except ImportError:
         continue
-    prov_urlpatterns = getattr(prov_mod, 'urlpatterns', None)
-    if prov_urlpatterns:
-        urlpatterns += prov_urlpatterns
+    prov_mod_urlpatterns = getattr(prov_mod, 'urlpatterns', None)
+    if prov_mod_urlpatterns:
+        provider_urlpatterns += prov_mod_urlpatterns
```
blueyed: Do not really have to rename `prov_urlpatterns` here - happy to revert those lines (I was exposing `prov_urlpatterns` initially, but then thought `provider_urlpatterns` is more explicit).

PR #2329, opened 2019-07-17
Fixed links, .html to .rst

2O4:
Fixed links, from `.html` to `.rst` in the doc file `forms.rst`.


pennersr:
This does not work for me. After building the html using sphinx, I end up with .rst links in the .html

PR #2334, opened 2019-08-06
 add simple support to SITE_ID for email

beres:
Use case:
We have one project with different domains, we would like to permit to users to use the same email to register across the domains. The email is unique per site. I was unable to find a simple solution with allauth, so this patch add the support.

If the variable EMAIL_SITE_ID is specified EmailAddress.email is no longer unique. filter_users_by_email verifies that the email is not already present for the same domain.

EMAIL_SITE_ID must contain the path to the field that contains the user's site_id.

For example:

class User(AbstractUser):
  account = models.ForeignKey('Account'...

class Account(models.Model):
  site = models.ForeignKey(Site, default=settings.SITE_ID, on_delete=models.CASCADE)

EMAIL_SITE_ID must contains 'user__account__site'

It was tested with this settings:
ACCOUNT_USER_MODEL_USERNAME_FIELD = 'username'
ACCOUNT_EMAIL_REQUIRED = True
ACCOUNT_USERNAME_REQUIRED = False
ACCOUNT_AUTHENTICATION_METHOD = 'email'
ACCOUNT_UNIQUE_EMAIL = False
ACCOUNT_USER_DISPLAY = 'user.email'

ACCOUNT_EMAIL_SITE_ID = 'user__account__site'

Please let me know if there are easier solutions

### DIFF
--- a/allauth/account/adapter.py
+++ b/allauth/account/adapter.py
@@ -296,8 +296,8 @@ class DefaultAccountAdapter(object):
         validate_password(password, user)
         return password
 
-    def validate_unique_email(self, email):
-        if email_address_exists(email):
+    def validate_unique_email(self, email, site=None):
+        if email_address_exists(email, site=site):
             raise forms.ValidationError(self.error_messages['email_taken'])
         return email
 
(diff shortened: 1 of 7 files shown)


pennersr:
Within allauth scope, username would need to be handled as well (making it local per site), as well as social accounts, which will turn internals upside down. At this time I would prefer to decline supporting this -- an alternative that would not require changes to all models would be to use multiple databases -- store the user related models for each site in a different database. That would be less intrusive and maybe even cleaner from a deployment perspective.

PR #2336, opened 2019-08-08
fix(templates): Don't HTML-escape plain text emails

Foobartender:
This affects, for example, site names containing "&".


mario:
Looks decent.

Issue #1307, opened 2016-02-22
Users' email addresses can be leaked


Flimm:
If you attempt to sign up with an email address that has already been associated with a user, you get this error message:

> A user is already registered with this e-mail address.

This is a privacy leak. It is now possible for me to verify if a user is signed up to a website just by attempting to sign up with their email address. In theory, it also is possible for me to get a list of many of the email addresses of signed up users, by attempting to enumerate email addresses.

A better practise is to show a message similar to this to the user:

> We have sent an email to foo@bar.com with further instructions.

If the user is new, then the email will contain the validation link. If not, it can contain a message like this:

> Someone attempted to register a new user account on bar.com with the email address foo@bar.com. If this was you, and you are attempting to change your password, then please follow this link to the password reset form. If not, then you can ignore this email.

---

Another vector where users' email addresses can be leaked is through the password reset form. If you enter an invalid email address in the password reset form, you get this message:

> The e-mail address is not assigned to any user account

This is a privacy leak. Again, it is now possible for to me verify if a user is signed up to a website, just by attempting to reset their password with this form.

A better practise is to show a message similar to this to the user:

> We have sent an email to foo@bar.com with further instructions.

If the email address is valid, then the usual email is sent. If not, an email like this one should be sent:

> You (or someone else) entered this email address foo@bar.com when trying to change the password of an account at foobar.com. However, this email address is not on our database of registered users and therefore the attempted password change has failed.
> 
> If this was not you, please ignore this email.

---

Troy Hunt has a good article on this subject, look for the subheading [Username enumeration and the impact on anonymity](http://www.troyhunt.com/2012/05/everything-you-ever-wanted-to-know.html).

pennersr:
Just to add: the leak occurs in the signup page as well, depending on the settings. So avoid that, you would need to disable unique email or allow users to signup with a username instead. So the impact is not just on the password reset page.

destroyerofbuilds:
@pennersr pre-existing practice should not compromise the employment of good security practices.

[...]

PR #2342, opened 2019-08-14
Fix #2341

### LINKED ISSUE #2341: Password Reset form exposes registered E-Mails
The current reset form tells you if an address does not exist which is problematic if you are using your E-Mail to log in. By default you should get a message that an E-Mail was sent to that address regardless if the user exists or doesn't.

Would you be open for accepting a pull request for this issue?

mario:
Signed-off-by: Mario Danic <mario@lovelyhq.com>

### DIFF
--- a/allauth/account/forms.py
+++ b/allauth/account/forms.py
@@ -291,7 +291,7 @@ class BaseSignupForm(_base_signup_form_class()):
         ]
         if app_settings.SIGNUP_EMAIL_ENTER_TWICE:
             self.fields["email2"] = forms.EmailField(
-                label=_("E-mail (again)"),
+                label=_("Confirm Email"),
                 widget=forms.TextInput(
                     attrs={
                         'type': 'email',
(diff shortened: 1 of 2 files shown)


mario:
@giant995 I don't mind changing this (will do it in an hour or two) but am not a fan of the terminology used.

giant995:
I see why you're not a fan and I agree better words could be used. The idea is staying vague about the inner working of the system will reminding the user to provide a valid email otherwise they will never receive an answer.

mario:
@giant995 updated. I'll open another PR once I have a suggestion for better wording.

9mido:
@mario Can you add onto this PR and do the same thing for signups? See issue 1307 for more info.

mario:
@9mido I could, but I wonder what's the point when this PR has been sitting here ignore for ages?

9mido:
Yea I see what you mean @mario it can be frustrating to wait so long. 

I connected your PR to where the main issue 1307 is being tracked so it hopefully gets more attention. It seems the issue you originally connected to was closed because it was possibly a duplicate of 1307. 

Even some other core allauth PR feature/fixes is taking a while to merge :/

But if the demand for the community is high (which it is) it should eventually make it into the main library (at least we hope).

pennersr:
This mentions it fixes #2341, but #2341 is not the whole story -- see #1307.  Eventually, when #1307 is fixed, you would always send an email to the user. For example, in case the email address is not known and you request a password reset, an email would be sent stating "You requested a password reset for email X, yet, we have no such account on file. Please signup ...".  So the current text, "We have sent you an e-mail..." would still be more approriate.

Issue #2343, opened 2019-08-16
Unconfigured providers should be ignored instead of throwing an ImproperlyConfigured exception


olivierdalang:
Hi !

Currently, an exception is thrown when a provider is added to INSTALLED_APPS but not configured from the admin : `django.core.exceptions.ImproperlyConfigured: No Facebook app configured`

IMO, it would be much better to just ignore unconfigured providers from the providers registry. This would allow apps to work out of the box. This would also allow adding several providers extensions, and leaving it to the website manager to enable the providers he actually wants from django admin. We could still have a warning in the logs to help out beginners.

In my setup, this issue means that when I deploy my app from scratch, I get an error out of the box (as the homepage redirects guests users to the login screen). It also prevents using the social login screen for the admin if you want to configure the app from the admin.

Thanks

adamhooper:
This is a concern when configuring dev environments, too. And test environments.

I expect my project to have 100 collaborators, and I only expect 5 of those people to need to fiddle with SocialApps. I also expect tests to run without social accounts configured. How does one accomplish these tasks?

I tried a janky monkey-patch to back SocialApps with flat-file config. (The idea being, different devs would set up files as they need them, and we'd have a file on production.) That failed because OAuth/OAuth2 `SocialToken.app` is a foreign key.

I feel this is an urgent bug. The way I see it, every developer who uses a Django app with socialaccounts sees a 500 error by default. And it's my understanding that tests are similarly difficult to author. Yet here we are, buried at #2343 with a single participant. That means 2342 other bugs came first. I don't mean to complain -- rather, I want to help! But ... am I missing something obvious here? Are all the figurative cool kids high-fiving with some awesome solution, leaving me (and olivierdalang) in the dark?

Tomorrow I'll try an approach more similar to #2343.

adamhooper:
A-ha. This only happens for Facebook. I guess the cool kids aren't using Facebook :).

denisemauldin:
We're currently using an older version of django-allauth partially because of this functionality.  Is there a way to escalate this fix?  Can I help somehow?  Thanks!

pennersr:
Fixed via affcfa09

denisemauldin:
Thank you!

pennersr:
https://codeberg.org/allauth/django-allauth/issues/2343

PR #2344, opened 2019-08-16
Fix for #2343 (allow unconfigured providers)

olivierdalang:
Quickfix for #2343 

This PR filters the list returned by ProviderRegistry.get_list to only return providers that are configured.

This prevents exceptions for unconfigured providers on the login page, as they will simply be ignored. This in turn allows adding several providers extensions to INSTALLED_APPS, and leaving it to the website manager to enable the providers he actually wants from django admin.

I made this as a very quick fix, and the code could probably be improved (e.g. import at the top), but I was getting an `AppRegistryNotReady` error and don't have time to investigate now.

### DIFF
--- a/allauth/socialaccount/providers/__init__.py
+++ b/allauth/socialaccount/providers/__init__.py
@@ -1,6 +1,4 @@
 import importlib
-import warnings
-
 from collections import OrderedDict
 
 from django.conf import settings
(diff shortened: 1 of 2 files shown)


pennersr:
Not all providers feature a SocialApp -- take OpenID for example.

olivierdalang:
@pennersr ok I can give it another shot, do you prefer that I open a new PR or can you reopen this one and I'll add new commits ?

PR #2347, opened 2019-08-21
feat(gitlab): Set the default scope to read_user

mscherer:
This permit to handle gitlab as easily as the other providers.

By default, if no scope is provided, it default to use 'api',
which seems overkill for simple authentication.

# Submitting Pull Requests

## General

 - [X] All Python code must be 100% pep8 and isort clean.
 - [X] JavaScript code should adhere to [StandardJS](https://standardjs.com).
 - [X] Feel free to add yourself to `AUTHORS`.
 
 ## Provider Specifics
 
### DIFF
--- a/allauth/socialaccount/providers/gitlab/provider.py
+++ b/allauth/socialaccount/providers/gitlab/provider.py
@@ -21,6 +21,9 @@ class GitLabProvider(OAuth2Provider):
     name = 'GitLab'
     account_class = GitLabAccount
 
+    def get_default_scope(self):
+        return ['read_user']
+
     def extract_uid(self, data):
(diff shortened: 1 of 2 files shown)


mscherer:
I realize that this do change the default, so someone who rely on the default to be "api" would have their application broken after upgrade. I did a test, and if you have a read_user scope on the token, and ask for api, it fail. It you have a "api" scope (which is the default when nothing is checked) on the token, requesting a 'read_user' also fail. 

But I assume, after the change on https://github.com/pennersr/django-allauth/pull/1860, everybody is using a explicit scope, which shouldn't break.

PR #2357, opened 2019-09-06
fix(account): add reset_password link to password_change.html

Flimm:
This is to help users who cannot remember their current password. It may not be obvious to users that to change their password, they would have to log out, and then find the "forgot password" link. This provides a link so that they don't have to do that.


pennersr:
Closed via d6dbec110aa381ba721fadc4bf4f3f81f9715a49

PR #2358, opened 2019-09-12
Fixes to steam provider

jleclanche:


pennersr:
This looks a little bit suspicious. Why would the Steam API return invalid JSON? Either the data is returned in a different format, but I would like to know what data it does return... or, there is some other error ongoing that is now swept under the carpet.

jleclanche:
@pennersr I believe it's done the same way in other backends as well. As for why it would return invalid JSON, server errors may be the case. The thing is the login is still valid, but you may not get "extra data" from the steam API in such a case (the base API is OpenID and we don't touch that; this is an except on the proprietary API calls on top of it).

jleclanche:
@pennersr ping

pennersr:
I think what is missing is a resp.raise_for_status(). Likely, you are not receiving an 2xx at the time the JSON is absent.  I would like to avoid pokemon exception handling...

jleclanche:
@pennersr There's already status code checking earlier in the code; this is if the json() call fails (steam sends invalid json). I don't think there's a cleaner path to checking it.

pennersr:
I do not see any resp.raise_for_status() here?:

image

nebriv:
I am also having issues with the Steam Provider. It looks like this PR would resolve my issues.

I'm getting "JSONDecodeError at /accounts/steam/callback/" after the authorization.

ckarrie:
Any news on this? Still gettings JSONDecodeError with Steam auth.

ckarrie:
You need to put the Steam Web API key into the "secret" Field in your Social App - not in Client-ID, that fixes for me

MarechJ:
@ckarrie What do you add in Client-ID then? It's a mandatory field

MarechJ:
> 
> 
> I think what is missing is a resp.raise_for_status(). Likely, you are not receiving an 2xx at the time the JSON is absent. I would like to avoid pokemon exception handling...

This is correct. When this case is happening to me, it's because the response is a 500

pennersr:
Closed via 73295318

Issue #2084, opened 2018-08-18
Unable to treat password resets as email verification confirmations


immanuelnoel:
I'm using users to login into their auto-created accounts. They need to do a password reset since they are not provided with a password - the templates are worded accordingly. 

Now, since ACCOUNT_EMAIL_VERIFICATION is set to "mandatory", they get an email right after the password reset. I'm unable to double up password reset, as a email verification step. 

I'm thinking of listening for the signal, allauth.account.signals.password_reset(request, user), and calling - confirm_email(self, request, email_address). 

Is this the only way, or am I missing a configuration.

derek-adair:
@rokstrnisa it is courteous to close issues like this if you come to a resolution <3.  Thank you for your proactive solution I will likely be adding it to the docs as an example.

@pennersr close as not planned.  The user demonstrated a very simple and straightforward solution that doesn't require modifying allauth.

pennersr:
The password reset token does not store the email the password reset was sent to, meaning, you don't know what email address should be verified. Even in case of just 1 email address, there is no guarantee that that email address is the same as the one the reset was sent to if the user changes the email after sending the reset email. Therefore, this is a valid issue that needs a design decision -- whether or not to store the email (id) in the token.

pennersr:
Dupe of #1267

pennersr:
https://codeberg.org/allauth/django-allauth/issues/2084

PR #2365, opened 2019-10-02
fixed issue #2084 where password reset from email does not verify user's email

prdpx7:
this commit fixes issue #2084

### DIFF
--- a/allauth/account/forms.py
+++ b/allauth/account/forms.py
@@ -556,6 +556,14 @@ class ResetPasswordKeyForm(PasswordVerificationMixin, forms.Form):
 
     def save(self):
         get_adapter().set_password(self.user, self.cleaned_data["password1"])
+        try:
+            # verify user's EmailAddress on password reset via email
+            email_address = EmailAddress.objects.get_for_user(self.user, user_email(self.user))
+            email_address.verified = True
+            email_address.save()
+        except:
+            # do nothing if email, user does not exists in EmailAddress
+            pass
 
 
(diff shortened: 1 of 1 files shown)


allauth/account/forms.py
```diff
@@ -556,6 +556,14 @@ def __init__(self, *args, **kwargs):
 
     def save(self):
         get_adapter().set_password(self.user, self.cleaned_data["password1"])
+        try:
+            # verify user's EmailAddress on password reset via email
+            email_address = EmailAddress.objects.get_for_user(self.user, user_email(self.user))
```
pennersr: A user can have multiple email addresses. Here, you happen to return the one that is one user.email, but it is not 100% certain that this was the email address the password reset mail was sent to.

allauth/account/forms.py
```diff
@@ -556,6 +556,14 @@ def __init__(self, *args, **kwargs):
...
+            email_address.verified = True
+            email_address.save()
```
pennersr: `.save(update_fields=['verified'])`

pennersr:
Closing -- see my comments.

danlamanna:
@pennersr I'm not the original author, but I'm interested in this quality of life feature.

It seems like placing the email in the token itself is a strict requirement for this feature. If a PR were to be made that had some sort of `EmailBasedPasswordResetTokenGenerator` which instead of being a user id and timestamp were an email address and timestamp, would you be open to accepting it?

PR #2388, opened 2019-11-09
fix(templates): Fix #2387 Error when 500 template contains a django-allauth template tag

### LINKED ISSUE #2387: Error when 500 template contains a django-allauth template tag
**Error message**: 
- `AttributeError: 'NoneType' object has no attribute 'POST'`

**How to reproduce**:
1) Create a 500 template (`500.html` in your template directory) that includes a template tag from django-allauth. For me, it was `Google Sign In` button:
```html
<!--500.html-->
{% load socialaccount %}
<a class="nav-link" href="{% provider_login_url 'google' %}">Log In</a>
```

2. Add an endpoint that is handled by Django's default 500 handler (`handler500` in `django.conf.urls`, which by default points to `django.views.defaults.server_error`)
```python
# urls.py
from django.conf.urls import handler500
urlpatterns = [
    # ...
    path('500/', handler500, name='500'),
    # ...
]
```
- The handler (`server_error`) renders the template: `return HttpResponseServerError(template.render())`
- `render` in `allauth/socialaccount/templatetags/socialaccount.py` is called
- `get_request_param` in `allauth/utils.py` is called in the line `next = get_request_param(request, 'next')`
- `return request.POST.get(param) or request.GET.get(param, default)` in `get_request_param` causes an error because `request` is `None` in this case

**Solution**:
- Add a guard statement to the function `get_request_param` in `allauth/utils.py`
- before a patch: Use a custom 500 handler instead of Django's default 500 handler

Bartleby2718:
# Submitting Pull Requests

## General

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -66,6 +66,7 @@ Jeff Triplett
 Jeremy Satterfield
 Jerome Leclanche
 Jesse Gerard Brands
+Jihoon Park
 Jiyoon Ha
 Joe Vanderstelt
 John Bazik
--- a/allauth/utils.py
+++ b/allauth/utils.py
@@ -299,4 +299,6 @@ def get_form_class(forms, form_id, default_form):
 
 
 def get_request_param(request, param, default=None):
+    if request is None:
+        return default
     return request.POST.get(param) or request.GET.get(param, default)


Bartleby2718:
@pennersr Can you tell me why this is not merged yet? I saw another pull request that was created after this and merged before this, and I'm curious if there's anything I can do to get it merged.

Also, can you tell me how often a patch is released? I'd like to know whether I should wait for a patch or work around the bug another way.

Thank you in advance!

PR #2411, opened 2019-12-16
Pass allowed_hosts value to is_safe_url()

### LINKED ISSUE #1766: `next` parameter not respected on login, when using an absolute URL
When logged in, "next" after a GET to the login url is not respected if it starts with http://, e.g.: http://localhost:8000/accounts/login/?next=http://localhost:8000/foo/ redirects to http://localhost:8000 not to http://localhost:8000/foo/.

I think the reason is that in https://github.com/pennersr/django-allauth/blob/master/allauth/account/adapter.py#L420 allowed_hosts is not passed to django.utils.http.is_safe_url.

I fixed it in my custom adapter using:

    def is_safe_url(self, url):
        from django.utils.http import is_safe_url
        return is_safe_url(url, allowed_hosts=settings.ALLOWED_HOSTS)

radwon:
Fixes #1766

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -9,6 +9,7 @@ Contributors
 Aaron van Derlip
 Abhishek Kumar Jaiswal
 Adam McKerlie
+Adam Radwon
 Agustin Perez Paladini
 Ahmet Emre Aladağ
 Aldiantoro Nugroho
--- a/allauth/account/adapter.py
+++ b/allauth/account/adapter.py
@@ -419,7 +419,8 @@ class DefaultAccountAdapter(object):
                 is_safe_url as url_has_allowed_host_and_scheme,
             )
 
-        return url_has_allowed_host_and_scheme(url, allowed_hosts=None)
+        return url_has_allowed_host_and_scheme(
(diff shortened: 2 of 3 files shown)


pennersr:
Why is this change needed? Aren't you introducing a security risk by passing along the host from the request? The host header can easily be changed using a host header attack, and then the server side would blindly assume that this fake host is allowed.

radwon:
It is needed to fix #1766 since using `allowed_hosts=None` means no allowed hosts.

Using `request.get_host()` is safe because Django validates the request HOST header against settings.ALLOWED_HOSTS otherwise it raises a DisallowedHost exception. However, an alternative is to use `settings.ALLOWED_HOSTS` directly.

Since in Django 3.0 `is_safe_url()` has been renamed to `url_has_allowed_host_and_scheme()` I will update the fix and pull request.

radwon:
@pennersr I'm not sure why test coverage has decreased with this pull request as I've included a unit test so it should have increased.

radwon:
@pennersr input and improvements provided, can you update the pull request labels.

PR #2422, opened 2019-12-23
fix(socialaccount): Fix expression that store tokens

flowgunso:
# Submitting Pull Requests

## General

When a SocialApp is fetched from the settings, the object's instance is not saved and hence does not have an id/pk (https://docs.djangoproject.com/en/2.2/ref/models/instances/#auto-incrementing-primary-keys), the previous expression cannot work.

### DIFF
--- a/allauth/socialaccount/models.py
+++ b/allauth/socialaccount/models.py
@@ -230,7 +230,7 @@ class SocialLogin(object):
         user.save()
         self.account.user = user
         self.account.save()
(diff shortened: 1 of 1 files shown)


pennersr:
This MR is outdated, code currently uses app_id

goranmoomin:
@pennersr The PR is outdated, but looks like `self.token.app_id` is still None, which makes the original issue still a problem.
Would it be possible if you take a look at this one more time? Thanks so much.

PR #2424, opened 2019-12-25
Add «Sign in with Apple» provider

0leg5ergeev:
# Submitting Pull Requests
This pull request adding Apple provider to allauth package
PR has a breaking change now — update to store Apple config in the `SocialApp` model

## General

 ## Provider Specifics
 

0leg5ergeev:
Could anybody help with tests for it? Any tutorial on how to run tests for this package? 
Also, there are new dependencies, packages PyJWT and Cryptography, is it ok?

Thanks for any help and offers to make it better.

bashu:
@0leg5ergeev did you ever test this PR?

0leg5ergeev:
@bashu if you mean unit tests, as you see, no. I stopped on this.
If you talking about "does it work", yes, it works great in my project. 

Please share your error and let's fix it. 

Anyway, you are welcome to do it better.

bashu:
@0leg5ergeev you should update setup.py and list pyjwt as requirements

simon-gaia:
@0leg5ergeev Thanks for your hard work! I'm desperately waiting for your PR to get included - so would be amazing if it could get merged in @bashu :)

s1m-0n:
@0leg5ergeev Thanks for your hard work! I'm desperately waiting for your PR to get included - so would be amazing if it could get merged in @bashu :)

fpghost:
@0leg5ergeev thanks for this. However, when I tried your PR I'm seeing

``` 
(Pdb) jwt.decode(data['id_token'], public_key, algorithm="RS256", verify=True,audience=client_id)
*** jwt.exceptions.InvalidSignatureError: Signature verification failed
```

when you try and decode the `id_token`

0leg5ergeev:
@fpghost Thank you for your comment, yes, I noticed this too. Apple drops changes to their mechanism without any updates, I found [this thread](https://forums.developer.apple.com/thread/129047) at their forum and will fix it soon, now I trying to run a unit test.

0leg5ergeev:
okay then, I need help to run tests. [This line](https://github.com/pennersr/django-allauth/pull/2424/files#diff-579fe88cf513d91b95a7f0686be9bd05R30) return `None` instead [this url](https://appleid.apple.com/auth/keys) content in the test environment, but in "production" it works well, I do not understand how to mock _requests.get_ data. Any examples or ideas? @fpghost @bashu

fpghost:
@0leg5ergeev Maybe you could patch `requests` in your unit tests?

```
@patch('socialaccount.providers.apple.client.requests.get')
def some_unit_test(mock_get):
    # Configure the mock to return a response with an OK status code.
    mock_get.return_value.ok = True

   ....
```

metroid09:
I've been trying to use django rest framework and rest_auth with the apple provider, but I'm having some issues getting it to work. When I pass in an `access_token` only the view throws an exception. Here is the relevant part of the stack trace.

```
File "/opt/klyp_staging/lib/python3.6/site-packages/rest_framework/serializers.py" in is_valid
  235.                 self._validated_data = self.run_validation(self.initial_data)

File "/opt/klyp_staging/lib/python3.6/site-packages/rest_framework/serializers.py" in run_validation
  433.             value = self.validate(value)

File "/opt/klyp_staging/lib/python3.6/site-packages/rest_auth/registration/serializers.py" in validate
  119.         social_token = adapter.parse_token({'access_token': access_token})

File "/opt/klyp_staging/src/django-allauth/allauth/socialaccount/providers/apple/views.py" in parse_token
  49.             data['id_token'],
```

It appears that `parse_token` depends on `'id_token'` being in the data and the way that it's getting passed in from rest_auth doesn't allow for that.

fpghost:
@metroid09 You could maybe modify rest auth's `SocialLoginSerializer`  by overriding the `validate` method and have it take the auth code instead from your frontend and then finish the OAuth flow from there. Something like

```
            code = attrs.get('code')

            provider = adapter.get_provider()
            scope = provider.get_scope(request)
            client = self.client_class(
                request,
                app.client_id,
                app.secret,
                adapter.access_token_method,
                adapter.access_token_url,
                self.callback_url,
                scope
            )
           token_dict = client.get_access_token(code)

```

I was looking at getting rest_auth to work too, and I think even if you modified the apple provider to access access token and refresh token there is no "profile url" for apple to go grab the user details from using those tokens and create the user (unlike say google/facebook where if you have the tokens you can hit up  a profile url and grab the uid, name, email etc). Or am I wrong?

fpghost:
@0leg5ergeev I modified slightly to get the key and select the right public key from apple for the decode

```
class AppleOAuth2Adapter(OAuth2Adapter):
    ...

    def get_public_key(self, id_token):
        kid = jwt.get_unverified_header(id_token)['kid']
        apple_public_key = [d for d in requests.get(self.public_key_url).json()['keys'] if d['kid'] == kid][0]
        public_key = jwt.algorithms.RSAAlgorithm.from_jwk(
            json.dumps(apple_public_key)
        )
        return public_key

    def parse_token(self, data):
        token = SocialToken(token=data['access_token'])
        token.token_secret = data.get('refresh_token', '')

        public_key = self.get_public_key(data['id_token'])
        ...
```

Hope that helps

0leg5ergeev:
@fpghost nice to see, thanks!

stevejalim:
What are the outstanding sticking points on this PR, please? Is the test coverage complete enough and it's just the ChangeLog conflict that needs fixing, or can I/we/anyone contribute some additional work to get this across the line - with 30 April approaching, am sure lots of people (myself included) would be grateful for a built-in solution to Apple Sign-In

0leg5ergeev:
@stevejalim There are no unit tests for a new provider and I have some troubles with creating it. 
Of course, you are welcome to contribute to this branch, it will be very helpful.

allauth/socialaccount/providers/apple/client.py
```diff
@@ -0,0 +1,92 @@
...
+            url,
+            data=data,
+            headers=self.headers)
+        access_token = None
+        if resp.status_code in [200, 201]:
+            if (resp.headers['content-type'].split(
+                    ';')[0] == 'application/json' or resp.text[:2] == '{"'):
+                access_token = resp.json()
+            else:
+                access_token = dict(parse_qsl(resp.text))
```
benhowes: Have you ever had a json response?

allauth/socialaccount/providers/apple/client.py
```diff
@@ -0,0 +1,92 @@
...
+            raise OAuth2Error('Error retrieving access token: %s'
+                              % resp.content)
+        return access_token
+
+    def get_redirect_url(self, authorization_url, extra_params):
+        params = {
+            'client_id': self.consumer_key,
+            'redirect_uri': self.callback_url,
+            'response_mode': 'form_post',
+            'scope': ' '.join([Scope.NAME, Scope.EMAIL]),
```
benhowes: AFAICT, we cannot make use of the `name` scope because the name is only included in the `POST` from the `form_post` response from apple and we have no way to safely add that to the session (other than some wacky way to include that data in the redirect URL, which I can't endorse as being a good idea).

benhowes:
The build is failing here, and it looks like a caching issue because I have run it on a fork, (here: https://travis-ci.org/github/ismfg/django-allauth/builds/666440062) and it installs the packages which have been added to `setup.py`. Tox's listing of installed packages indicates that the new requirements are being missed.

Can you try removing the cache key from `.travis.yaml` (L35 -> L38), so we can rule that out @0leg5ergeev?

0leg5ergeev:
@benhowes sorry, I have no `cache key` at `.travis.yml (L35 -> L38)`. 
Maybe @pennersr can rerun Travis?

allauth/socialaccount/providers/apple/client.py
```diff
@@ -0,0 +1,100 @@
...
+        }
+        self._strip_empty_keys(data)
+        resp = requests.request(
+            self.access_token_method,
+            url,
+            data=data,
+            headers=self.headers)
+        access_token = None
+        if resp.status_code in [200, 201]:
+            if self.is_json_response(resp):
```
pennersr: Is this helper method really needed? Couldn't you just invoke resp.json() directly and catch the ValueError in case it is not proper JSON?

allauth/socialaccount/providers/apple/provider.py
```diff
@@ -0,0 +1,27 @@
...
+    def extract_common_fields(self, data):
+        return dict(
+            email=data.get('email'),
+        )
+
+    def extract_email_addresses(self, data):
+        ret = []
+        email = data.get('email')
+        if email and data.get('email_verified'):
+            ret.append(EmailAddress(email=email, verified=True, primary=True))
```
pennersr: If there is an email, we should always add it to the list, even if not verified. 
```suggestion
            ret.append(EmailAddress(email=email, verified=email.get('email_verified'), primary=True))
```

setup.py
```diff
@@ -121,7 +121,9 @@ def find_package_data(where=".", package="", exclude=standard_exclude,
     install_requires=['Django >= 2.0',
                       'python3-openid >= 3.0.8',
                       'requests-oauthlib >= 0.3.0',
-                      "requests"],
+                      "requests",
+                      "pyjwt >=1.7",
+                      "cryptography >= 2.8"],
```
pennersr: Is cryptography needed? I don't see it being referenced in your code?

0leg5ergeev: @pennersr sure it does, as it described [here](https://pyjwt.readthedocs.io/en/latest/installation.html#cryptographic-dependencies-optional), there is RS256 decoding requires `cryptography` package

allauth/socialaccount/providers/apple/client.py
0leg5ergeev: @pennersr updated by this [commit](https://github.com/pennersr/django-allauth/pull/2424/commits/25d094299f426d892a04fe521c7d186fe5510745)

allauth/socialaccount/providers/apple/provider.py
0leg5ergeev: @pennersr updated by this [commit](https://github.com/pennersr/django-allauth/pull/2424/commits/25d094299f426d892a04fe521c7d186fe5510745)

docs/providers.rst
```diff
@@ -121,6 +121,33 @@ App registration (get your key and secret here)
...
+    Client ID: Service ID identifier
+    Secret Key: Key ID
+
+In settings.py:
+.. code-block:: python
+
+    SOCIALACCOUNT_PROVIDERS = {
+        'apple': {
+            'MEMBER_ID': 'Member ID/App ID Prefix',
+            'SECRET_KEY' : """-----PRIVATE KEY-----"""
```
safwanrahman: I think it is better to take the value from the Database like other social applications. You need to increase the `secret_key` field `max_length` for that.

allauth/socialaccount/providers/apple/client.py
```diff
@@ -0,0 +1,88 @@
...
+
+        MEMBER_ID = APPLE_PROVIDER_SETTINGS.get('MEMBER_ID', None)
+        SECRET_KEY = APPLE_PROVIDER_SETTINGS.get('SECRET_KEY', None)
+
+        CURRENT_TIMESTAMP = int(time())
+        claims = {
+            'iss': MEMBER_ID,
+            'aud': 'https://appleid.apple.com',
+            'sub': self.consumer_key,
+            'iat': CURRENT_TIMESTAMP,
```
safwanrahman: It can  be written as following

```suggestion
            'iat': timezone.now(),
            exp': timezone.now() + timedelta(hours=1),
```

docs/providers.rst
benhowes: I did wonder about that when starting to work on this PR, but I thought that since it's a file, there is actually some benefit in being able to keep it as a file which can be loaded in the settings as needed.

allauth/socialaccount/providers/apple/client.py
benhowes: Thanks! I am going to update to use `datetime.utcnow()` though, since I don't want to give the impression that timezones matter in the JWT creation.

docs/providers.rst
benhowes: Thinking about this more, you are right. 

The auth for the token exchange is based on both the member id and the secret key. There is a field we're not using at the moment `key` which seems like it's used to access the profile API in other clients (e.g. Trello), so perhaps that's a better fit? 

The member_id and the secrey_key are a pair at the moment too - I don't know that it makes sense to have one global and the other per social application. Would you want to see an additional field on the social application model for this?

The final part of this is that  `OAuth2Client` doesn't receive `app.key` at the moment, so I can add that too?

benhowes: Here's an idea of what that would look like https://github.com/0leg5ergeev/django-allauth/pull/5/files (combines your other review comment at the moment)

allauth/socialaccount/providers/apple/views.py
```diff
@@ -0,0 +1,130 @@
...
+    def parse_token(self, data):
+        try:
+            token = SocialToken(token=data["access_token"])
+            token.token_secret = data.get("refresh_token", "")
+
+            public_key = self.get_public_key(data["id_token"])
+            provider = self.get_provider()
+            allowed_auds = self.get_client_id(provider)
+
+            token.user_data = jwt.decode(
```
safwanrahman, line 58: The implementation of sign in with apple is weired. We are getting the JWT key form apple server, why do we need to verify again with the key they have provided? It makes the thing much slower unfortunately!

0leg5ergeev: Sure we have, it makes verification that no one has interfered to logging in.
That's how I understand the process. [here are the docs](https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_rest_api/verifying_a_user#overview)
@safwanrahman

safwanrahman: @0leg5ergeev if anyone interfered, we would not get the access token while generating the token. That is what I understand. Correct me if I am wrong.

allauth/socialaccount/providers/apple/client.py
```diff
@@ -0,0 +1,84 @@
...
+            payload=claims, key=self.cert, algorithm='ES256', headers=headers
+        ).decode('utf-8')
+        return client_secret
+
+    def get_client_id(self):
+        """ We support multiple client_ids, but use the first one for api calls """
+        return self.consumer_key.split(',')[0]
+
+    def get_access_token(self, code):
+        url = self.access_token_url
```
safwanrahman, line 48: I think we need to verify the token before getting the access token as it is mentioned in the [doc](https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_rest_api/verifying_a_user#overview). Correct me if I am wrong

allauth/socialaccount/providers/apple/client.py
```diff
@@ -0,0 +1,84 @@
...
+                'Error retrieving access token: %s' % resp.content
+            )
+        return access_token
+
+    def get_redirect_url(self, authorization_url, extra_params):
+        params = {
+            'client_id': self.get_client_id(),
+            'redirect_uri': self.callback_url,
+            'response_mode': 'form_post',
+            'scope': ' '.join([Scope.EMAIL]),
```
safwanrahman, line 78: I think it is beteer to use `self.scope`

allauth/socialaccount/providers/apple/provider.py
```diff
@@ -0,0 +1,31 @@
+from allauth.account.models import EmailAddress
+from allauth.socialaccount.providers.base import ProviderAccount
+from allauth.socialaccount.providers.oauth2.provider import OAuth2Provider
+
+
+class AppleProvider(OAuth2Provider):
+    id = 'apple'
+    name = 'Apple'
+    account_class = ProviderAccount
+
```
safwanrahman, line 10: We should include `name, email` in `get_default_scope` method.

allauth/socialaccount/providers/apple/views.py
```diff
@@ -0,0 +1,130 @@
...
+        * returns AppleOAuth2Client
+        * Apple requests callback by POST
+    """
+
+    def dispatch(self, request, *args, **kwargs):
+        if request.method == "POST":
+            url = request.build_absolute_uri(request.get_full_path())
+            params = {
+                "code": request.POST.get("code"),
+                "state": request.POST.get("state"),
```
safwanrahman, line 120: You should pass the scope data as well.

allauth/socialaccount/providers/apple/provider.py
pennersr: Whether or not `email` is in the default scope should depend on `from allauth.socialaccount.app_settings import QUERY_EMAIL`.

allauth/socialaccount/providers/apple/views.py
```diff
@@ -0,0 +1,130 @@
...
+
+class AppleOAuth2CallbackView(AppleOAuth2ClientMixin, OAuth2CallbackView):
+    """
+    Custom OAuth2CallbackView because `Sign In With Apple`:
+        * returns AppleOAuth2Client
+        * Apple requests callback by POST
+    """
+
+    def dispatch(self, request, *args, **kwargs):
+        if request.method == "POST":
```
safwanrahman, line 116: What is the reason to redirect to the same path with `GET` method? We need to fix the `OAuth2CallbackView` to use `get_request_param` for getting the code.
Change the follwoing
`access_token = client.get_access_token(request.GET['code'])`
to
`access_token = client.get_access_token(get_request_param('code'))`

0leg5ergeev: I am not sure to update `OAuth2CallbackView`, that's why. If you are ready to help, please commit changes to my branch 🤷 @safwanrahman

allauth/socialaccount/providers/apple/client.py
benhowes: In that doc it says "Web apps must obtain the identity token from their server by validating the authorization code" so I think they're kind of saying that we're verifying the code by exchanging it for the access_token? That was my reading of it anyway. There are no other API endpoints that seem to fit the bill for verifying the code.

allauth/socialaccount/providers/apple/views.py
benhowes: @safwanrahman I think you're right, but as @0leg5ergeev points out the docs say you _must_ verify the signature. It does seem a little heavy-handed to verify it when it's come straight from an apple endpoint. 

I'd vote to keep it, because a) it follows the apple guidelines, b) it is easier for hooking in to the auth process via an API (In which case the id_token may be from an iOS SDK)

allauth/socialaccount/providers/apple/client.py
safwanrahman: @benhowes It is interesting. The docs are very much misleading with their terminology.
The code we are getting here is actually "authorization code" as mentioned in the [doc](https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_js/incorporating_sign_in_with_apple_into_other_platforms) and it is not **Identity Token**


image


And the token we are getting after calling the `access_token_url` is besically `refresh token` as mentioned in the [doc](https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_rest_api/verifying_a_user#3385187). As per the doc "**After verifying the identity token on your server, call the Generate and validate tokens endpoint with the client_id, client_secret, and nonce information**", but we are trying to get the refresh token before validing the **`identity token`**

0leg5ergeev: @safwanrahman I just can say, that docs were very poor at December when I published this PR :)

allauth/socialaccount/providers/apple/views.py
safwanrahman: @benhowes It was really confusing actually as per the docs. But the identity token that we should decode is passed in `id_token` key while the user [got redirected to our call back url](https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_js/incorporating_sign_in_with_apple_into_other_platforms#3332115). As it is coming directly from the client, we need to validate the `id_token` when it is coming from the callback.

allauth/socialaccount/providers/apple/views.py
safwanrahman: @0leg5ergeev Yeah,I will make a PR to your fork.

allauth/socialaccount/providers/apple/client.py
safwanrahman: @0leg5ergeev Yeah, I can endorse that the docs were worse and I was almost gone while implementing sign in with apple with a short deadline back in december or january. It is good that it is better.

safwanrahman:
@0leg5ergeev I have made a PR https://github.com/0leg5ergeev/django-allauth/pull/7 at your fork. The tests are failing, can you check?

allauth/socialaccount/providers/oauth2/views.py
```diff
@@ -16,22 +16,20 @@
...
-)
+from allauth.socialaccount.providers.oauth2.client import OAuth2Client, OAuth2Error
 from allauth.utils import build_absolute_uri, get_request_param
 
 from ..base import AuthAction, AuthError
 
 
 class OAuth2Adapter(object):
-    expires_in_key = 'expires_in'
+    client_cls = OAuth2Client
```
safwanrahman, line 26: @benhowes Excellent idea! 💯🚀 I have also thought about it, but somehow did not take the courage to do this!

docs/overview.rst
```diff
@@ -45,6 +45,8 @@ Supported Providers
 
 - AngelList (OAuth2)
 
+- Apple (OAuth2)
```
safwanrahman, line 48: I am actually hesitate to call it a OAuth2 😅😅

safwanrahman:
@pennersr I think the Pull Request is in good state to review. Can you please review?

0leg5ergeev:
> @pennersr I think the Pull Request is in good state to review. Can you please review?

tests have failed because of formatting, I can fix it in 2-3 hours

benhowes:
I think there is a problem here actually - it seems that `request.session` is empty when the POST from apple redirects to the site. I'm investigating.

0leg5ergeev:
@benhowes you can watch a report [here](https://travis-ci.org/github/pennersr/django-allauth/jobs/671652852)

benhowes:
Update - got to the bottom of the problem.

When there is a `form_post` request back to the site, django's `SESSION_COOKIE_SAMESITE = "Lax"` means that the cookies are not sent on that request. This means that we do not have access to the session in this case.

The only broadly applicable solution to this problem at the moment is to set `SESSION_COOKIE_SAMESITE = None` in django settings.

The old method of redirecting the initial POST had the unintended benefit of avoiding this issue.

0leg5ergeev:
>The old method of redirecting the initial POST had the unintended benefit of avoiding this issue. 
Huh, I've just remembered about it right now. it wasn't `unintended`, I and a guy from my team spent a day for tries to avoid this problem. In the end, we decided that redirect from `GET` to `POST` is the best way

safwanrahman:
@benhowes Nice finding. But where do we need the request session? is not it stateless?

docs/providers.rst
```diff
@@ -142,6 +142,12 @@ In Social App:
 page, or it’s your App ID Prefix in your App ID)
     Cert: The contents of your private key, including "-----BEGIN/END PRIVATE KEY-----" and line breaks
 
+In your django settings, the means by which apple returns data to your app requires that cross-site cookies are allowed. This can be done in your Django settings:
+
+.. code-block:: python
+
+    SESSION_COOKIE_SAMESITE = None
```
0leg5ergeev, line 149: Is it «secure» to post this flag for web apps?

benhowes: Right, so I've looked in to this a bit, it should not be a problem since the setting only applies to the session cookie. If it was global then this would be bad (i.e. the csrf token must have at least "Lax").

I'm adding to the docs to say that `SESSION_COOKIE_SECURE = True` should also be set too

benhowes:
@safwanrahman It's not stateless because we save some state with `SocialLogin.stash_state()` which is then loaded in the callback. That's not specific to apple, but for things like storing a `next_url`, it's the only way.

docs/providers.rst
0leg5ergeev: Hmm, okay. But I am still not sure that this decision is better than redirect `get->post`. If you can, please convince me. 
Anyway, @pennersr has the last word.

safwanrahman: I am also thinking about the tradeoff. Maybe we can make the thing stateless for apple!

safwanrahman: I think we should implement a different approach for apple as they support [state using query params](https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_js/incorporating_sign_in_with_apple_into_other_platforms#3332115) instead of session. We need to add the state according to it. Setting this value in settings may lead unintended security risk.

benhowes: So, a moderate problem with the previous approach was that the user's name is only provided in the `user` key of the`form_post` body. With the `POST`->`GET` redirect, there is nowhere to safely store the user's name.

Making this particular flow stateless would be a large deviation from the way that allauth works for all other providers at the moment. 

As mentioned, the `SESSION_COOKIE_SAMESITE` only applies to the session cookie, so it would not prevent django's CSRF protection, which to my mind severely limits the ability of this to cause problems, especially if CORS is correctly configured. 

That said, I am not a web security expert.

benhowes: I've created a new (and arguably better) approach here: https://github.com/0leg5ergeev/django-allauth/pull/9

The TL;DR is:
- Use a `POST`->`GET` redirect as the apple form_post return endpoint
- Create an apple login specific session (since the normal `request.session` is not available with default/secure cookie settings with a cross-origin post). This stores the `id_token` and `user` from the response. You can think of this endpoint as making the apple response behave more like a normal oauth2 response
- The subsequent endpoint is much more normal for allauth and fetches the data it needs back from the apple specific session (then removes the apple session)

It's not perfect, but it does avoid changing cookie settings to be potentially insecure and it also keeps the apple specific code very contained to this new endpoint.

Keen for feedback!

madprime:
Mostly FYI to let you know how valuable this is & noting urgency/timeliness. An enormous 💖 to everyone that's contributing here!

I need to implement Apple login as this is blocking deployment of community-developed apps on iOS developed in response to the covid-19 crisis that want to use our platform for login & data management (Open Humans, citizen science with personal data). I'm likely to pick one of these solutions and hope it's merged, or hope that switching to the different, merged solution isn't going to be a problem with the database. (Advice appreciated!)

babu-upcomer:
@madprime For now you can install it from the forked project and test. I am using this one and hoping that I could make it work. 

`pipenv install -e git+https://github.com/ismfg/django-allauth.git@apple-specific-session#egg=django-allauth`

0leg5ergeev:
@pennersr I think, it were final edits to PR. Can you review it please?

benhowes:
I've created a diagram of the flow: (cc @babu-upcomer)

Apple signin flow

Diagram source: https://swimlanes.io/u/Ccjq9w1pF

madprime:
@0leg5ergeev & others – thanks so much for all the work here! I ran into an issue when trying this out for our site, and I shared a fix that worked for me: https://github.com/0leg5ergeev/django-allauth/pull/10

allauth/socialaccount/providers/apple/views.py
```diff
@@ -0,0 +1,180 @@
...
+        """ We need to gather the info from the apple specific login """
+        add_apple_session(request)
+
+        # Exchange `code`
+        code = get_request_param(request, "code")
+        access_token_data = client.get_access_token(code)
+
+        return {
+            **access_token_data,
+            **self.get_user_scope_data(request),
```
safwanrahman, line 133: I think we should keep the unverified `user` scope data in separate key so that it can does not overwrite the `access_token_data`. The `user` data can be altered from the client

safwanrahman:
@benhowes The approach seems great. But I was thinking what about saving the state in the apple specific session and pass that session key in the `state` parameter of the apple login? In that way we can avoide the additional redirect and complexity.

madprime:
Have you tested this on an apple phone or ipad? The behavior is different. I ran into one issue, got around it, and then another.

First issue – the encoding of response_type and scope need to be different when sent to Apple, see here: https://forums.developer.apple.com/message/381068#381068 (note the device-specific behavior in that discussion: _"Somehow, the Sign In with Apple from an iPhone/iPad using Face ID/Touch ID isn't working if the parameters aren't encoded, but the ones from a non-Apple device are working fine."_ I was able to fix this issue with this code change: https://github.com/madprime/django-allauth/commit/69d5c73b17c293db9026e0c779200f10bf1d088b

Second issue – (once the above fix was done) the user is sent back to my site, but authorization fails. I've attempted to investigate, and it seems like the added session data (`id_token` and `name`) isn't available in the post_callback (but is fine on a non-Apple device; I don't know why this might happen).

(Note, to get this all working I've also been using [a fix I shared earlier](https://github.com/0leg5ergeev/django-allauth/pull/10) regarding a `ValidationError` due to needing a boolean rather than string.)

allauth/socialaccount/providers/apple/views.py
benhowes: Ah, yes. Well, 2 things.

1. We _should_ whitelist the keys from `user` so that we know we're not passing in bad content.
2. The actual processing of the keys should depend on the signed id_token being from apple as expected. This does mean that any auth attempt will bail around this point if the request is invalid.

benhowes:
> @benhowes The approach seems great. But I was thinking what about saving the state in the apple specific session and pass that session key in the `state` parameter of the apple login? In that way we can avoide the additional redirect and complexity.

@safwanrahman this goes back to the problem of having something more than a one-time code in the URL which is not something we can consider safe.

benhowes:
@madprime the second issue you mention seems to relate to cookie same-site policies. I have access to an iPhone so I can see this issue, but I do not have access to a mac to be able to more thoroughly debug :(

Something I did notice is that on safari iOS, a popup touch ID confirmation comes up, which I assume should auth you, however it then redirects anyways. I guess the JS framework may be needed in page with the meta data configured etc? https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_js

babu-upcomer:
@benhowes @madprime Have you guys come across the following error? I have copy pasted .p8 key file content into certificate field for the Apple provider.

Screenshot 2020-04-22 at 8 31 30 PM

babu-upcomer:
@benhowes I can confirm the issue tested with my iPhone X. @madprime Were you able to figure that out?

babu-upcomer:
@benhowes @madprime Any updates please? Thanks in advance!

babu-upcomer:
@madprime @benhowes Sorry for the false alarm, I think the problem is with the rest auth libraries like dj-rest-auth and it's predecessor django-rest-auth in my case. Here is the equivalent code line  which fails with KeyError "id_token" in dj-rest-auth, https://github.com/jazzband/dj-rest-auth/blob/master/dj_rest_auth/registration/serializers.py#L118. Ideally it should be passing the entire `token` object. I customised the serializer class and fixed it myself.

allauth/socialaccount/providers/apple/apple_session.py
```diff
@@ -0,0 +1,35 @@
...
+    patch_vary_headers(response, ('Cookie',))
+    request.apple_login_session.save()
+    response.set_cookie(
+        APPLE_SESSION_COOKIE_NAME,
+        request.apple_login_session.session_key,
+        max_age=None,
+        expires=None,
+        domain=settings.SESSION_COOKIE_DOMAIN,
+        # The cookie is only needed on this endpoint
+        path=response.url,
```
mhsparks, line 31: I've done some digging and I think this is causing the issues with login on safari and presumably apple devices. `response.url` is a full URI whilst we should just be setting a path, I presume other browsers are more forgiving in their cookie handling.
 
I'll investigate further and submit a PR next week unless someone gets a chance sooner.

gedankenstuecke: Thanks, that would be amazing!

madprime: The fix works for me – I left a note saying the same on the PR you later opened, thank you! :sparkling_heart:

madprime: The fix works for me – I left a note saying the same on the PR you later opened, thank you! :sparkling_heart: hopefully others find it's good as well.

madprime:
Well, the version in @mhsparks [PR to this branch](https://github.com/0leg5ergeev/django-allauth/pull/11) fixed the remaining issues I ran into and seems to be working – and thanks to that, our app finally got through Apple app store review. :tada: 

:crossed_fingers: that this is merged into master!

pennersr:
:+1: for putting this together.  There are two things that are holding me back from merging this. The first is an overall allauth design issue, where keys & secrets are stored in the database (`SocialApp`). That way of working may need to be deprecated over time, and this PR contributes to the issue by introducing another field (`cert`) that may be better off served elsewhere. The other is trick with the Apple session cookie. I really want to see this running myself and experiment with it a bit. For that I simply need time, which is a bit hard to come by at the moment...

fonar:
Hi, is it possible to use this provider with native iOS application?
I successfully get it working with web application, but can't with iOS native.

gedankenstuecke:
@fonar If that helps: We're using the branch of this PR to login inside a native iOS app, the login is handled through a Web view inside the app.

fonar:
> @fonar If that helps: We're using the branch of this PR to login inside a native iOS app, the login is handled through a Web view inside the app.

@gedankenstuecke thanks you so much i'm trying to do the same, but stuck, when i need to specify redirectURI for AppleID.auth.init javascript method, but i can't specify capactiror://localhost, as apple don't allow it as redirectURI. What is the solution could be here? One more question, do you have some sample code or do you know tutorial for such implementation?

gedankenstuecke:
Ah, sorry, our use-case is slightly different I think, as we use the native web view to login people on our own website but using _Sign In with Apple_ as the login-method there. So our redirect for the Apple auth goes to a "normal" URI on our website. Once a person is then logged in into our website this triggers our own OAuth2 flow that redirects people back to the native application. 

If you want to see it in action you can have a look at the app that implements that at https://apps.apple.com/us/app/oh-data-port/id1512384252?ls=1 (and the source code for it is public at https://github.com/OpenHumans/qf-heartrate-apple-health/)

Arideno:
What SOCIALACCOUNT_PROVIDERS parameters should I use for sign in with apple?

Arideno:
Trying to use Apple Provider. Get this error. Do you have any thoughts?
`File "/home/deploy/spinevoxenv/lib/python3.7/site-packages/rest_framework/serializers.py" in is_valid
  235.                 self._validated_data = self.run_validation(self.initial_data)

File "/home/deploy/spinevoxenv/lib/python3.7/site-packages/rest_framework/serializers.py" in run_validation
  433.             value = self.validate(value)

File "/home/deploy/spinevoxenv/bin/spinevox/authuser/serializers.py" in validate
  410.         social_token = adapter.parse_token({'access_token': access_token})

File "/home/deploy/spinevoxenv/lib/python3.7/site-packages/allauth/socialaccount/providers/apple/views.py" in parse_token
  93.         identity_data = self.get_verified_identity_data(data["id_token"])

Exception Type: KeyError at /api/auth/social/apple/
Exception Value: 'id_token'
`

Arideno:
Trying to use Apple Provider. Get this error. Do you have any thoughts?

```
File "/home/deploy/spinevoxenv/lib/python3.7/site-packages/rest_framework/serializers.py" in run_validation
  433.             value = self.validate(value)

File "/home/deploy/spinevoxenv/bin/spinevox/authuser/serializers.py" in validate
  410.         social_token = adapter.parse_token({'access_token': access_token})

File "/home/deploy/spinevoxenv/lib/python3.7/site-packages/allauth/socialaccount/providers/apple/views.py" in parse_token
  93.         identity_data = self.get_verified_identity_data(data["id_token"])

Exception Type: KeyError at /api/auth/social/apple/
Exception Value: 'id_token'
```

babu-upcomer:
> Trying to use Apple Provider. Get this error. Do you have any thoughts?
> 
> ```
> File "/home/deploy/spinevoxenv/lib/python3.7/site-packages/rest_framework/serializers.py" in run_validation
>   433.             value = self.validate(value)
> 
> File "/home/deploy/spinevoxenv/bin/spinevox/authuser/serializers.py" in validate
>   410.         social_token = adapter.parse_token({'access_token': access_token})
> 
> File "/home/deploy/spinevoxenv/lib/python3.7/site-packages/allauth/socialaccount/providers/apple/views.py" in parse_token
>   93.         identity_data = self.get_verified_identity_data(data["id_token"])
> 
> Exception Type: KeyError at /api/auth/social/apple/
> Exception Value: 'id_token'
> ```

The problem is in, `social_token = adapter.parse_token({'access_token': access_token})`. Check my finding in https://github.com/pennersr/django-allauth/pull/2424#issuecomment-620543627

nomhoi:
How do I use this provider with django-rest-auth?

soymonitus:
Hi, I have read so many issues and different changes over the original pull-request... Is there any pull request or fork that has everything working without issues? With this, can you do native sign in with Apple so I can later send the token to my backend using REST auth?

paulnicolet:
@soymonitus I also need to implement this ASAP since the deadline is coming fast.
I managed to put everything together using this PR and Rest Auth, but it needs some tweaking on the Rest Auth side, as @babu-upcomer said.
It is as simple as overriding a serializer though, which is pretty straightforward, but invokes some copy pasting if you do not wish to fork Rest Auth... so still a temporary solution and not quite elegant and permanent.

@pennersr I understand the issue with introducing one more secret stored in the DB, however, since it's designed this way since a long time now, should we really let this PR stuck, in prevision of future `allauth` design changes ? It's ok since we all manage with workarounds, but maybe more people really need this feature, since the deadline is in less than 10 days now 😉

Thanks all !

alexcican:
@pennersr Apple’s deadline to implement `Sign in with Apple` is looming… any updates on this?

pennersr:
Sorry, I don't have any ETA on this...  if you rely on Apple Sign-In, you can always pip install a the repo/branch of this PR. I am aware that this is not ideal, but it does work for the time being...

babu-upcomer:
Do anybody else have the bug that doesn't populate the user first name and last name from the apple sign-in? I'm on iPhone X with Face ID and debugged all the way down but couldn't figure out if the issue is at frontend or backend. At both we send `email` and `name` as the scope.

Arideno:
@babu-upcomer try to reinstall the app. Should work.

whit:
@babu-upcomer I can't get how to modify `django-rest-auth`'s serializer to make it work. Please can you help me (and others) and provide code snippet? My view is:

```python
class AppleLogin(SocialLoginView):
    adapter_class = AppleOAuth2Adapter
    client_class = AppleOAuth2Client
    callback_url = getattr(settings, 'API_APPLE_CALLBACK_URL', 'https://fitrecepty.info/apple/login/callback/')
    serializer_class = serializers.AppleLoginSerializer
```

babu-upcomer:
```
class SocialLoginSerializer(serializers.Serializer):
    access_token = serializers.CharField(required=False, allow_blank=True)
    code = serializers.CharField(required=False, allow_blank=True)

    def _get_request(self):
        request = self.context.get('request')
        if not isinstance(request, HttpRequest):
            request = request._request
        return request

    def get_social_login(self, adapter, app, token, response):
        """
        :param adapter: allauth.socialaccount Adapter subclass.
            Usually OAuthAdapter or Auth2Adapter
        :param app: `allauth.socialaccount.SocialApp` instance
        :param token: `allauth.socialaccount.SocialToken` instance
        :param response: Provider's response for OAuth1. Not used in the
        :returns: A populated instance of the
            `allauth.socialaccount.SocialLoginView` instance
        """
        request = self._get_request()
        social_login = adapter.complete_login(request, app, token, response=response)
        social_login.token = token
        return social_login

    def validate(self, attrs):
        view = self.context.get('view')
        request = self._get_request()

        if not view:
            raise serializers.ValidationError(
                _('View is not defined, pass it as a context variable')
            )

        adapter_class = getattr(view, 'adapter_class', None)
        if not adapter_class:
            raise serializers.ValidationError(_('Define adapter_class in view'))

        adapter = adapter_class(request)
        app = adapter.get_provider().get_app(request)

        # More info on code vs access_token
        # http://stackoverflow.com/questions/8666316/facebook-oauth-2-0-code-and-token

        # Case 1: We received the access_token
        if attrs.get('access_token'):
            access_token = attrs.get('access_token')
            token = {'access_token': access_token}

        # Case 2: We received the authorization code
        elif attrs.get('code'):
            self.callback_url = getattr(view, 'callback_url', None)
            self.client_class = getattr(view, 'client_class', None)

            if not self.callback_url:
                raise serializers.ValidationError(
                    _('Define callback_url in view')
                )
            if not self.client_class:
                raise serializers.ValidationError(
                    _('Define client_class in view')
                )

            code = attrs.get('code')

            provider = adapter.get_provider()
            scope = provider.get_scope(request)
            client = self.client_class(
                request,
                app.client_id,
                app.secret,
                adapter.access_token_method,
                adapter.access_token_url,
                self.callback_url,
                scope,
                key=app.key,
                cert=app.cert,
            )
            token = client.get_access_token(code)
            access_token = token['access_token']

        else:
            raise serializers.ValidationError(
                _('Incorrect input. access_token or code is required.'))

        social_token = adapter.parse_token(token)
        social_token.app = app

        try:
            login = self.get_social_login(adapter, app, social_token, access_token)
            complete_social_login(request, login)
        except HTTPError:
            raise serializers.ValidationError(_('Incorrect value'))

        if not login.is_existing:
            # We have an account already signed up in a different flow
            # with the same email address: raise an exception.
            # This needs to be handled in the frontend. We can not just
            # link up the accounts due to security constraints
            if allauth_settings.UNIQUE_EMAIL:
                # Do we have an account already with this email address?
                if get_user_model().objects.filter(email=login.user.email).exists():
                    raise serializers.ValidationError(_('E-mail already registered using different signup method.'))

            login.lookup()
            login.save(request, connect=True)

        attrs['user'] = login.account.user
        return attrs
```

@whit That's my custom serializer. Changes are at `self.client_class()` and `social_token = adapter.parse_token(token)`.

ayushin:
Anybody got this working with djang-rest-auth so far?

ayushin:
I'm getting stuck with the id_token key error

ayushin:
sorry, do i miss something but if we have identity token (from a native app) we don't need to pass the code?

ashrafemad:
@ayushin 
you can try sending only the `code` field which is the Authorization Code from the mobile app

myles:
I have a small bug fix PR here, https://github.com/0leg5ergeev/django-allauth/pull/12. It fixes a bug we found when using `response_mode=form_get`. Basically what happens is the response won't contain an email or email_verified parameters and the library will raise an exception:

```python
File "/usr/local/lib/python3.7/site-packages/allauth/socialaccount/providers/apple/provider.py" in extract_email_addresses
  33.             verified = verified.lower() == 'true'

Exception Value: 'NoneType' object has no attribute 'lower'
```

jsbls:
Hey everybody, I'm trying to get this to work with DRF but I'm having some trouble, hopefully someone can shed some light.
Error: `Incorrect input. access_token or code is required.`

AFAIK, this error has something to do with my `Social App` config, but I did as required on the docs:
```
Client ID: Service ID identifier, App ID (using the Bundle ID since the Prefix is required below)
Secret Key: Key ID
Key: App ID Prefix
Cert: The contents of your private key, including "-----BEGIN/END PRIVATE KEY-----" and line breaks
```

I've also added the corresponding class to `views.py` & url to `urls.py`:
```
View
---
class AppleLogin(SocialLoginView):
	adapter_class = AppleOAuth2Adapter
	permission_classes = []

Url
---
re_path(
	r'^rest-auth/apple/$', 
	user_views.AppleLogin.as_view(), 
	name='apple_login'
),
```

Any help is greatly appreciated. :)

skumarcm:
To use django-rest-auth for Apple signin from mobile App, I had to do the following:
1. Send the "authrizationCode" from the mobile app to server
2. Use the modified SocialLoginSerializer that @babu-upcomer provided above
3. The above serializer needs to be updated slightly to ensure that the right client_id is used to verify the "AuthorizationCode" with the server. Currently, the serializer uses the "Service ID" as the client_id which is good for web based authentication. For mobile authentication, we need to use the app ID as the client id which is stored at the index 1. I used the following code to determine the client_id while instantiating the client_class in the above serializer:
client_id= app.client_id.split(',')[1]
Use the above client_id instead of the "app.client_id" for instantiating the client_class

itaisteinherz:
@0leg5ergeev Shouldn't `cert` also be added to the array here (line 200)?

https://github.com/pennersr/django-allauth/blob/c19a212c6ee786af1bb8bc1b07eb2aa8e2bf531b/allauth/socialaccount/adapter.py#L193-L204

allauth/socialaccount/providers/apple/views.py
```diff
@@ -0,0 +1,180 @@
...
+from datetime import timedelta
+
+from django.http import HttpResponseRedirect
+from django.urls import reverse
+from django.utils import timezone
+from django.utils.http import urlencode
+from django.views.decorators.csrf import csrf_exempt
+
+import jwt
+from requests import HTTPError
```
itaisteinherz, line 12: It seems like this is unused, and so this import can be removed.

bcarradini:
Can anyone provide an example of how to configure `SOCIALACCOUNT_PROVIDERS` to get up-and-running with this PR branch? The apply provider docs currently read: 

```In Social App:
    Client ID: Service ID identifier. In the case that you have an app too, this field should be a comma separated list, with the service identifier first, followed by any app identifiers.
    Secret Key: Key ID
    Key: Member ID/App ID Prefix (find it below your name at the top right corner of the
page, or it’s your App ID Prefix in your App ID)
    Cert: The contents of your private key, including "-----BEGIN/END PRIVATE KEY-----" and line breaks
```

I can't find any reference to `SOCIALACCOUNT_PROVIDERS` settings in the apple tests. Should I be directly seeding the `SocialApp` model in my database?

itaisteinherz:
@bcarradini I've yet to figure out how to make provider configuration using `SOCIALACCOUNT_PROVIDERS` work. What worked for me was to manually create a `SocialApp` object in the Django admin according to the documentation you included above, and manually apply the fix I suggested [here](https://github.com/pennersr/django-allauth/pull/2424#issuecomment-662370490).

allauth/socialaccount/providers/apple/views.py
```diff
@@ -0,0 +1,180 @@
...
+        add_apple_session(request)
+
+        # Exchange `code`
+        code = get_request_param(request, "code")
+        access_token_data = client.get_access_token(code)
+
+        return {
+            **access_token_data,
+            **self.get_user_scope_data(request),
+            "id_token":request.apple_login_session.get("id_token")
```
bcarradini, line 134: The contents of `access_token_data` should be the response from `POST https://appleid.apple.com/auth/token` (see `AppleOAuth2Client.get_access_token()`). According to [Apple's documentation](https://developer.apple.com/documentation/sign_in_with_apple/tokenresponse), the response should already contain the `id_token` attribute, and that is in fact what I see. So why override the `id_token` key with what is stored in `request.apple_login_session`?

This caused a problem from me when using ngrok to test apple login on localhost. Even though ngrok provides a public https URL, I was being directed back to the non-https version during the Apple login flow for reasons that I don't yet understand. Because of this, `request.apple_login_session` was being lost because `persist_apple_session()` stores the cookie with `secure=True`.

I can work around this issue on my local in two ways:
(1) comment out the above line of code.
(2) comment out `secure=True` in the `persist_apple_session()` function.

It may be worth considering whether the above line of code is necessary, given that `access_token_data` should already contain a value for `id_token`. If there _are_ scenarios in which `access_token_data` might be missing a value for `id_token`, this portion of code could be adjusted to first check if `id_token` is present in `access_token_data` before overriding it.

bcarradini:
For anyone interested, here's a run-down of how I was able to successfully use this PR for web and mobile login **and** how I was able to test locally, which took a bit of doing.

**My setup:**
- Django 2.2 web app that uses django-allauth to support web login
- [Expo-managed](https://expo.io/) React Native mobile app that uses an API published by the web app (django-rest-framework) to support mobile login

**Apple Developer setup (under `Certificates, Identifiers & Profiles`):**
- Configured the `App ID` for our mobile app to use `Sign In with Apple` and selected `Enable as a primary App ID`
- Created a `Sign In with Apple` key for the above `App ID` and downloaded the certificate
- Created a `Service ID` for our web app and enabled `Sign In with Apple` for it, associating it with the above `App ID`
  - Set domain as: `my.domain.com`
  - Set return URL with a trailing slash (**IMPORTANT**): `https://my.domain.come/apple/login/callback/`

**Adjustments for web login:**
- Added `'allauth.socialaccount.providers.apple'` to list of `ALLAUTH_APPS` in `project/settings.py`
- Manually created `SocialApp` object via Django admin console as follows:
  - Provider = `Apple`
  - Name = `Apple`
  - Client id = `<APPLE_SERVICE_ID>, <APPLE_APP_ID>, host.exp.Exponent`  _//host.exp.Exponent was required for local testing of Expo-managed mobile app_
  - Secret key = `<APPLE_KEY_ID>`
  - Key = `<APPLE_TEAM_ID>`
  - Certificate = `-----BEGIN PRIVATE KEY-----
...
-----END PRIVATE KEY-----` _//the certificate text exactly as downloaded, including line breaks_

**Adjustments for mobile login (front-end):**
- Followed [Expo documentation](https://docs.expo.io/versions/latest/sdk/apple-authentication/) to setup apple login
- (**IMPORANT**) The Expo apple authentication library provides a method to handle user sign in; when successful, the method returns an object that includes _both_ `authorizationCode` (which translates to `access_token` in this PR's current implementation) and `identityToken` (which translates to `id_token` in this PR's current implementation). I wanted to send both of these pieces of information to the django-rest-framework backend (as `access_token` and `id_token`, respectively), so I had to make some adjustments to the social login serializer...

**Adjustments for mobile login (back-end):**
- We already had a custom REST login serializer that extended `rest_auth.registration.serializers.SocialLoginSerializer` and was similar to the `SocialLoginSerializer` proposed by @babu-upcomer above.
- I made the following changes to that custom serializer:
  - Defined a new, optional field for the serializer: `id_token = serializers.CharField(required=False, allow_blank=True)`. This allowed me to pass both `access_token` and `id_token` from the mobile front-end.
  - Changed: ```        social_token = adapter.parse_token({'access_token': access_token})``` to: ```        social_token = adapter.parse_token({
            'access_token': access_token,
            'id_token': attrs.get('id_token') # For apple login
        })

**Localhost test setup:**
- I used ngrok to setup a public URL that routed to my localhost, because the Apple Service ID can only be configured with publicly-accessible return URLs. In order to get this working, I had to:
  - I added the ngrok domain and redirect URL to the Apple Service ID config each time I started ngrok (e.g., `eb033bf7311f.ngrok.io`, `https://eb033bf7311f.ngrok.io/apple/login/callback/`)
  - I commented out `secure=True` in `../django-allauth/allauth/socialaccount/providers/apple/apple_session.py`, because the Apple login flow would return me to the HTTP version of my ngrok site even if I started at the HTTPS version, and `secure=True` was causing the cookie to be lost
  - Added `host.exp.Exponent` to the comma-separated list in `SocialApp.client_id` in order to get mobile login working for our Expo-managed app running inside the iOS simulator; otherwise, the `jwt.decode()` inside `AppleOAuth2Adapter.get_verified_identity_data()` fails due to "invalid audience."

Hope that's helpful to someone!

cchacholiades:
@0leg5ergeev thank you for the implementation on this PR  & @bcarradini thank you for the notes, super helpful. I have now successfully implemented it in our Django 2.2 project. It works great! 

I was wondering during the first time a user signs up if they select to share their email, does that apply for their name as well? 

```
account = SocialAccount.objects.get(user=self.user, provider="apple")
print(account.extra_data)
```

The fields in account extra_data: 

```
access_token 
id_token 
iss
aud 
exp
iat
sub
c_hash
email
email_verified
auth_time
nonce_supported
```

Can I acquire the users name and/or surname along with their email? (if they choose to share, also these info are given only the first time a user signs up?)

lihhh:
> For anyone interested, here's a run-down of how I was able to successfully use this PR for web and mobile login **and** how I was able to test locally, which took a bit of doing.
> 
> **My setup:**
> 
> * Django 2.2 web app that uses django-allauth to support web login
> * [Expo-managed](https://expo.io/) React Native mobile app that uses an API published by the web app (django-rest-auth) to support mobile login
> 
> **Apple Developer setup (under `Certificates, Identifiers & Profiles`):**
> 
> * Configured the `App ID` for our mobile app to use `Sign In with Apple` and selected `Enable as a primary App ID`
> * Created a `Sign In with Apple` key for the above `App ID` and downloaded the certificate
> * Created a `Service ID` for our web app and enabled `Sign In with Apple` for it, associating it with the above `App ID`
>   
>   * Set domain as: `my.domain.com`
>   * Set return URL with a trailing slash (**IMPORTANT**): `https://my.domain.com/apple/login/callback/`
> 
> **Adjustments for web login:**
> 
> * Added `'allauth.socialaccount.providers.apple'` to list of `ALLAUTH_APPS` in `project/settings.py`
> * Manually created `SocialApp` object via Django admin console as follows:
>   
>   * Provider = `Apple`
>   * Name = `Apple`
>   * Client id = `<APPLE_SERVICE_ID>, <APPLE_APP_ID>, host.exp.Exponent`  _//host.exp.Exponent was required for local testing of Expo-managed mobile app_
>   * Secret key = `<APPLE_KEY_ID>`
>   * Key = `<APPLE_TEAM_ID>`
>   * Certificate = `-----BEGIN PRIVATE KEY----- ... -----END PRIVATE KEY-----` _//the certificate text exactly as downloaded, including line breaks_
> 
> **Adjustments for mobile login (front-end):**
> 
> * Followed [Expo documentation](https://docs.expo.io/versions/latest/sdk/apple-authentication/) to setup apple login
> * (**IMPORANT**) The Expo apple authentication library provides a method to handle user sign in; when successful, the method returns an object that includes _both_ `authorizationCode` (which translates to `access_token` in this PR's current implementation) and `identityToken` (which translates to `id_token` in this PR's current implementation). I wanted to send both of these pieces of information to the django-rest-framework backend (as `access_token` and `id_token`, respectively), so I had to make some adjustments to the social login serializer...
> 
> **Adjustments for mobile login (back-end):**
> 
> * We already had a custom `RestSocialLoginSerializer` that extended `rest_auth.registration.serializers.SocialLoginSerializer` and was similar to the `SocialLoginSerializer` proposed by @babu-upcomer above. I made the following changes to that custom serializer:
>   
>   * Defined a new, optional field for the serializer: `id_token = serializers.CharField(required=False, allow_blank=True)`. This allowed me to pass both `access_token` and `id_token` from the mobile front-end.
>   * Changed `social_token = adapter.parse_token({'access_token': access_token})` to:
> 
> ```
>         social_token = adapter.parse_token({
>             'access_token': access_token,
>             'id_token': attrs.get('id_token') # For apple login
>         })
> ```
> 
> * Created an apple login view that extended our existing custom `RestSocialLoginView`:
> 
> ```
> class RestAppleLogin(RestSocialLoginView):
>     """Extend RestSocialLoginView to specify Apple adapter."""
>     adapter_class = AppleOAuth2Adapter
> ```
> 
> * Added to our existing django-rest-auth setup to include a URL for apple login:
> 
> ```
>     url(r'^api/v1/auth/apple$', RestAppleLogin.as_view(), name='rest_apple_login'),
> ```
> 
> **Localhost test setup:**
> 
> * I used ngrok to setup a public URL that routed to my localhost, because the Apple Service ID can only be configured with publicly-accessible return URLs. In order to get this working, I had to:
>   
>   * I added the ngrok domain and redirect URL to the Apple Service ID config each time I started ngrok (e.g., `eb033bf7311f.ngrok.io`, `https://eb033bf7311f.ngrok.io/apple/login/callback/`)
>   * I commented out `secure=True` in `../django-allauth/allauth/socialaccount/providers/apple/apple_session.py`, because the Apple login flow would return me to the HTTP version of my ngrok site even if I started at the HTTPS version, and `secure=True` was causing the cookie to be lost
>   * Added `host.exp.Exponent` to the comma-separated list in `SocialApp.client_id` in order to get mobile login working for our Expo-managed app running inside the iOS simulator; otherwise, the `jwt.decode()` inside `AppleOAuth2Adapter.get_verified_identity_data()` fails due to "invalid audience."
> 
> Hope that's helpful to someone!

to get the SOCIALACCOUNT_PROVIDERS to work, I have modified the AppleOAuth2Adapter's get_client_id
```
def get_client_id(self, provider):        
        # app = SocialApp.objects.get(provider=provider.id)        
        applep = providers.registry.by_id(provider.id, self.request)
        app = applep.get_app(self.request)
        return [aud.strip() for aud in app.client_id.split(",")]
```

allauth/socialaccount/providers/apple/views.py
mrkovalchuk: I too have found trouble with this behaviour. Don't understand why we override `id_token`

pennersr:
There is already a way of relying on using secrets from the configuration instead of reading them from the database:

`
SOCIALACCOUNT_PROVIDERS = { 'apple': { 'APP': { 'client_id': '123...', ...} } } 
`

In commit 744d9cddb5b6d86de1eccecac74fb99fc889ee0c this mechanism is extended to be able to take an additional `certificate` parameter, which alleviates the need for having to add another field to the SocialApp model as is done in this PR.

cchacholiades:
@0leg5ergeev Thank you for this PR. @bcarradini I have followed the instructions you listed above, very helpful. Now, I am having trouble to acquire the full name for the user. On iOS we can see the full name being populated in the object `ASAuthorizationAppleIDCredential`. Although, our client implementation sends the `id_token` + `access_token` to the backend and the variables `first_name` + `last_name` are always empty. Any ideas?

mrkovalchuk:
@cchacholiades 

> Our client implementation sends the id_token + access_token to the backend although the variables first_name + last_name are always empty

Apple send requested user information (first_name, last_name, email) only in first "Sing In" request. 

you can find Names in response to this request:
`https://appleid.apple.com/auth/authorize?\
state=abvgd&\
response_type=code&\
client_id=ServiceID&\
scope=email+name&\
response_mode=form_post&\ redirect_uri=https://redirec.url` 

User email you can get from id_token from response to this request: 

`curl -X POST https://appleid.apple.com/auth/token -d '\
client_id=some_client_id&\
code=12345&\
client_secret=jwt_part1.jwt_part2.jwt_part3&\
grant_type=authorization_code'`

only in FIRST auth request.

If you need "refresh" your logic: https://support.apple.com/en-us/HT210426 - remove your app from the list. 

I hope, it's answer for your trouble.

cchacholiades:
@mrkovalchuk thank you for your response. 

For the implementation on iOS we followed the [instructions here](https://developer.apple.com/documentation/authenticationservices/implementing_user_authentication_with_sign_in_with_apple). 

For the backend: I did the following amends to our serialiser and the iOS app sends via POST the fields `id_token, access_token`. This works great to login users, however, I think the values of first_name + last_name are empty because at this point it is not the first auth request.

```
social_token = adapter.parse_token({
    'access_token': access_token,
    'id_token': attrs.get('id_token') # For apple login
})
```

yurabysaha:
This PR will be include in master and when?

cchacholiades:
@0leg5ergeev We have been using your branch for a while now, works great so far. Could you please pull allauth version 0.42.0 (it supports a newer Facebook API) in your add-sign-in-with-apple-method branch?

kelkarajay:
@0leg5ergeev I've opened a pull request to bring your branch up to date with upstream master. Can you please review that and maybe merge? https://github.com/0leg5ergeev/django-allauth/pull/13

@cchacholiades

pennersr:
Merged via 16e9c61102de46a52d6c51e7b63fed22865eadb7 -- thanks!

kelkarajay:
@pennersr Are there plans to roll out this with a release? Say 0.43.0?

lee-hodg:
Should the `CLIENT_ID` really be a comma-delimited string of both client IDs then? like `CLIENT_ID = f"{APPLE_SERVICE_ID}, {APPLE_APP_ID}"? (in the `SOCIALACCOUINT_PROVIDERS` setting I mean)

winnerNick:
I have an issue with apple provider described in #2759

Could anyone help? Thanx

encetamasb:
> I commented out `secure=True` in `../django-allauth/allauth/socialaccount/providers/apple/apple_session.py`, because the Apple login flow would return me to the HTTP version of my ngrok site even if I started at the HTTPS version, and `secure=True` was causing the cookie to be lost

FIY when you are testing with ngrok you might want to set `ACCOUNT_DEFAULT_HTTP_PROTOCOL = 'https'` to fix the protocol of the callback url.

allauth/socialaccount/providers/apple/views.py
SeongIkKim: If I had seen your reviews a few days ago...
"Social Network Login Failure" says almost nothing to me.
Tracing up the reason why allauth missed `id_token` despite of valid `id_token` in my chrome network tap, was so hard journey.

2021.01.17
`Django==2.2.9`
`PyJWT==1.7.1`
At `HTTP` production environment,
this review really works for me and I finally made apple sign in successfully.

Thank you!  ☺️

gorapidrakesh:
Got this error 

AttributeError at /users/rest-auth/apple/
'str' object has no attribute 'decode'

  File "/app/users/serializers.py", line 301, in validate
    token = client.get_access_token(code)
  File "/app/allauth/socialaccount/providers/apple/client.py", line 55, in get_access_token
    client_secret = self.generate_client_secret()
  File "/app/allauth/socialaccount/providers/apple/client.py", line 45, in generate_client_secret
    payload=claims, key=app.certificate, algorithm="ES256", headers=headers

This error comes from this method 
  client_secret = jwt.encode(payload=claims, key=app.certificate, algorithm="ES256", headers=headers).decode("utf-8")                                                                                      
  return client_secret 

I customized the allauth package by removing decode("utf-8") from jwt and it worked.
Can anyone suggest to me a better way to resolve this error

DataGreed:
So, is it implemented? I cannot find anything in the docs - https://django-rest-auth.readthedocs.io/en/latest/search.html?q=apple&check_keywords=yes&area=default

coratype:
Is all of this convoluted patching still necessary? I can't find a straightforward answer on implementing Apple Auth with this

pennersr:
@coratype  You may want to look into https://docs.allauth.org/en/latest/headless/openapi-specification/ -- allauth has native support for this, without requiring patches.

pennersr:
@coratype But you seem to be still using dj-rest-auth, although putting it on a URL path that is from `allauth.headless`. What I meant was, you might want to drop dj-rest-auth altogether and just switch to `allauth.headless`.

coratype:
Okay so I found the solution to the default serializer not working. This is how you should set up the Swift request. Although now I'm getting a weird thing where its saying when a new user signs up that a user with the email already exists, but it succeeds at creating a user in the auth_user table regardless and future requests are authenticated?

```
Backend returned an error: 400
Response body: {"non_field_errors":["User is already registered with this e-mail address."]}

```

```
            let backendURL = URL(string: "\(baseURL)/api/_allauth/app/v1/auth/provider/token")!
            var request = URLRequest(url: backendURL)
            
            request.httpMethod = "POST"
            request.setValue("application/json", forHTTPHeaderField: "Content-Type")

            let parameters: [String: Any] = [
                "provider": "apple",
                "process": "login",
                "id_token": idTokenString,
                "access_token": authCodeString,
            ]

            do {
                request.httpBody = try JSONSerialization.data(withJSONObject: parameters, options: [])
            } catch {
                print("Error encoding parameters: \(error.localizedDescription)")
                return
            }
            
```
         
 
Original Issue

pennersr:
Please read up here: https://docs.allauth.org/en/latest/headless/installation.html

There is no need to do any of this:

```
from allauth.socialaccount.providers.apple.client import AppleOAuth2Client
from allauth.socialaccount.providers.apple.views import AppleOAuth2Adapter


class AppleLogin:
    adapter_class = AppleOAuth2Adapter
    client_class = AppleOAuth2Client
```

coratype:
Oh okay I see now, thank you! Removes a lot of overhead

coratype:
I also have a profile/user table id like to populate after a user successfully authenticates, how would I do that without a class? 

For anyone stumbling upon this in the future, this is the basic setup:

Swift:

```
if let appleIDCredential = authorization.credential as? ASAuthorizationAppleIDCredential {
            guard let identityToken = appleIDCredential.identityToken,
                  let authorizationCode = appleIDCredential.authorizationCode,
                  let idTokenString = String(data: identityToken, encoding: .utf8),
                  let authCodeString = String(data: authorizationCode, encoding: .utf8) else {
                print("Failed to decode identityToken or authorizationCode")
                return
            }
            
            let simulatorBaseURL = "http://192.168.1.28:8000"
            let deviceBaseURL = "https://dev.voir.space"

            #if targetEnvironment(simulator)
            let baseURL = simulatorBaseURL
            #else
            let baseURL = deviceBaseURL
            #endif

            let backendURL = URL(string: "\(baseURL)/_allauth/app/v1/auth/provider/token")!
            var request = URLRequest(url: backendURL)
            
            request.httpMethod = "POST"
            request.setValue("application/json", forHTTPHeaderField: "Content-Type")

            let parameters: [String: Any] = [
                "provider": "apple",
                "process": "login",
                "token": [
                    "client_id": "space.voir.voir",
                    "id_token": idTokenString,
                    "code": authCodeString,
                ]
            ]

            do {
                request.httpBody = try JSONSerialization.data(withJSONObject: parameters, options: [])
            } catch {
                print("Error encoding parameters: \(error.localizedDescription)")
                return
            }
```


Settings.py

```
SOCIALACCOUNT_PROVIDERS = {
    'apple': {
        'APP': {
            'client_id': APPLE_CLIENT_ID,
            'secret': APPLE_SECRET,
            'key': APPLE_KEY,
            "settings": {
                "certificate_key": """-----BEGIN PRIVATE KEY-----
erhjkdfsnjkh
-----END PRIVATE KEY-----
"""
            }
        }
    }
}

```

pennersr:
If you need to ask additional questions during signup, you can use the regular allauth `ACCOUNT_SIGNUP_FORM_CLASS`. Be sure to use `SOCIALACCOUNT_AUTO_SIGNUP = False`. Then, after posting the token you will still need to complete the https://docs.allauth.org/en/dev/headless/openapi-specification/#tag/Authentication:-Providers/paths/~1_allauth~1%7Bclient%7D~1v1~1auth~1provider~1signup/post provider signup endpoint.

coratype:
If I gather this all correctly

User initiates Sign Up with Apple via the headless Social route

SOCIALACCOUNT_AUTO_SIGNUP = False makes it so that the Sign Up doesn’t go through fully

Prompt the user to fill out a regular ALL_AUTH form

Send that via the headless SignUp route

Use that data to create the profile, Allauth takes care of auth_user, and done?

pennersr:
> Prompt the user to fill out a regular allauth form

No, this can all be done using headless. You signup by posting the apple token. If there is missing information (e.g. no email), or if auto signup is turned off, an additional step to complete the signup is needed. This is signalled to the client by means of a 401 when the token is POST'ed. You then POST the missing data to `_allauth/app/v1/auth/provider/signup` -- so this is not a regular Django form POST, it's all headless.

coratype:
Okay I completely understand now, thank you!

coratype:
Last thing, how does the headless know to create a Profile model with a one to one link to the auth_user after posting to provider/signup?

pennersr:
> how does the headless know to create a Profile model

This is not specific to headless, but rather general allauth usage. Suppose you want to ask the user for his/her favorite color at signup and store that, you typically use something like this:

```
# settings.py
ACCOUNT_SIGNUP_FORM_CLASS = 'foo.forms.AdditionalSignupFieldsForm'

# foo/forms.py:
class AdditionalSignupFieldsForm(forms.Form):
    color = forms.CharField()

        # At this point allauth has signed up the user -- create your profile model here.
        Profile.objects.create(user=user, color=self.cleaned_data['color'])
```

coratype:
Got it, I’ll apply it and see how it goes

coratype:
Am I missing anything here? I’m getting a

```
packages/allauth/socialaccount/adapter.py", line 168, in get_signup_form_initial_data
    user = sociallogin.user
           ^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'user'
```

Settings.py

```
SOCIALACCOUNT_AUTO_SIGNUP = False
ACCOUNT_SIGNUP_FORM_CLASS = 'api.forms.form.AdditionalSignupFieldsForm'
```

Form.py

```

from django import forms
from ..models import User


class AdditionalSignupFieldsForm(forms.Form):
    username = forms.CharField()

        # Log the user object to debug potential issues
        apple_id = user.extra_data.get('sub')

        User.objects.create(
            apple_id=apple_id,
            auth_user=user,
            email=user.email,
            username=self.cleaned_data['username'],
            image='',
            status='active',
        )

```

pennersr:
I suspect that you are calling the provider signup endpoint while the `provider_signup` flow is not pending, and is hence not expected to be called. Of course, it shouldn't be crashing like this, I'll see to it that that gets fixed, but do make sure to only call it when pending.

pennersr:
The crash is fixed via 2a8ebc73

coratype:
When the client receives the 401 response, I immediately prompt the user on iOS to enter their username and they submit it to the headless sign-up route, are you saying I have to use the is_open_for_signup method?

```
                // 401 response = Apple Token authenticated, prompt for username
                if httpResponse.statusCode == 401 {
                    DispatchQueue.main.async {
                        print("Backend returned 401: \(httpResponse.statusCode)")
                        self.isProviderSignupPending = true
                        self.promptForUsername(baseURL: baseURL)
                    }
                    
                    private func submitUsername(baseURL: String, username: String) {
        guard isProviderSignupPending else {
            print("Provider signup flow is not pending. Aborting the request.")
            return
        }
        let signupURL = URL(string: "\(baseURL)/_allauth/app/v1/auth/provider/signup")!
        var request = URLRequest(url: signupURL)
        
        request.httpMethod = "POST"
        request.setValue("application/json", forHTTPHeaderField: "Content-Type")

        let parameters: [String: Any] = [
            "username": username
        ]

        do {
            request.httpBody = try JSONSerialization.data(withJSONObject: parameters, options: [])
        } catch {
            print("Error encoding parameters: \(error.localizedDescription)")
            return
        }
        
        let task = URLSession.shared.dataTask(with: request) { data, response, error in
            if let error = error {
                print("Error sending request to backend: \(error.localizedDescription)")
                return
            }
            
            guard let httpResponse = response as? HTTPURLResponse, (200...299).contains(httpResponse.statusCode) else {
                if let data = data, let responseString = String(data: data, encoding: .utf8) {
                    print("Response body: \(responseString)")
                }
                return
            }
            
```

pennersr:
@coratype In your code, I don't think you are passing along the session token as instructed here:

https://docs.allauth.org/en/latest/headless/openapi-specification/#section/App-Usage/Session-Tokens

If you receive the 401 with the pending provider signup flow, and then post to that signup endpoint without passing along the accompanying session, the session is lost and the flow would indeed not be considered pending.

coratype:
I'm in! Thanks so much for being so helpful and quick to respond

coratype:
Sorry last thing, is the auth_user thats created accessible here so I can create the link between the auth_user and profile?

```
from django import forms
from ..models import User


class AdditionalSignupFieldsForm(forms.Form):
    username = forms.CharField()

        apple_id = user.extra_data.get('sub')
        Profile.objects.create(
            apple_id=apple_id,
            auth_user=user,
            email=user.email,
            username=self.cleaned_data['username'],
            image='https://assets.voir.space/profile_image%2F720sse7txjp4wrb-4f85392d-a18d-44e7-a9d4-77aa4983e3ed.jpg',
            status='active',
            follow_notifications=1,
            reply_notifications=1,
            heart_notifications=1,
            updated_at=None,
            last_active=None,
            date_joined=user.date_joined
        )
```

pennersr:
This bit seems off:

```
        apple_id = user.extra_data.get('sub')
```

Assuming you did not override the user model, it does not have `extra_data`. So it is not to be mistaken with the allauth `SocialAccount`. The `user` parameter is just the regular Django user model. So, if you need the `SocialAccount`, you can need to query/filter on that.

coratype:
Oh I think I conflated the sociallaccount with authuser, I understand completely now thanks !

coratype:
a few questions, ive been using this and it makes two calls to the database when creating a user, is this unavoidable?

also, can signup return more than just Session Token: ---------, User Data: ["has_usable_password": 0, "id": 16, "username": hellooo, "email":@gmail.com, "display": hellooo] and how come the ID is just a regular incrementing number? what im used to usually id is a UUID or something 

from django import forms
from ..models import User
import uuid
from allauth.socialaccount.models import SocialAccount


class AdditionalSignupFieldsForm(forms.Form):
    username = forms.CharField()

        social_account = SocialAccount.objects.get(user=user)
        data = social_account.extra_data
        apple_id = social_account.uid

        new_user = User.objects.create(
            id=str(uuid.uuid4()),
            auth_user=user,
            email=user.email,
            username=self.cleaned_data['username'],
            image='',
            status='active',
            updated_at=None,
            last_active=None,
            date_joined=user.date_joined
        )

        # Print new user creation
        print(f'New User Created: {new_user}')

thanks again for your help from last time

coratype:
> Not sure what you mean by this. If you are noticing that the session token can change from one value to another, that is correct... that can happen when Django calls `session.cycle_key()`. But, getting multiple session tokens in one go should not be possible.

when signing a user up the response is this

```
Session Token: lu64ksn5buclsu7im84u5wmdc27leuc6, User Data: ["has_usable_password": 0, "id": 16, "username": hellooo, "email":@gmail.com, "display": hellooo]
```

can i change the response to also include data from the custom user model?

pennersr:
Currently, that user data payload is not configurable. Note that its purpose is to only convey the minimal amount of information needed during the authentication phase, it is not intended for more general purpose profile information, exposing e.g. a profile picture and your personal bio. For those kinds of use cases, you will have to create your own profile endpoint.

coratype:
Got it, I’ll find a workaround

coratype:
Can all auth also receive the device token for APNS?

pennersr:
@coratype The APNS token is not related to authentication et al. An iOS app can request the token without authenticating. All in all, requesting the token and handing it over to the backend is completely orthogonal to authentication, and hence not related to allauth.

PR #2434, opened 2020-01-09
Fix: use new function, fallback to depricated one

2tunnels:
Newer Django versions has `url_has_allowed_host_and_scheme` function as replacement for deprecated`is_safe_url`.
I've added import check to use new one, if it exists and fallback for old one if necessary.
This patch will fix this warning:

```
RemovedInDjango40Warning: django.utils.http.is_safe_url() is deprecated in favor of url_has_allowed_host_and_scheme()
```

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -138,6 +138,7 @@ Udi Oron
 Victor Semionov
 Volodymyr Yatsyk
 Vuong Nguyen
+Vlad Dmitrievich
 Wendy Edwards
 Will Gordon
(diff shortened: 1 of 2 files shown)


2tunnels:
Failed because `E501 line too long (88 > 79 characters)`.
Should I create new pull request with proper formatting?

0leg5ergeev:
> Should I create new pull request with proper formatting?
Just push a new commit with changes

PR #2437, opened 2020-01-12
fix: Store RealmID for Quickbooks provider

ayushsood:
RealmID: The ID that identifies the specific QuickBooks company
Sub: An identifier for the user

The current implementation only stores sub. RealmIDs are used for the majority of API calls on behalf of the user.


pennersr:
This change breaks the test suite:

image

pennersr:
Closed via ee63361c

Issue #2443, opened 2020-01-18
Google Provider Error


steverecio:
When I try social authentication with Google the provider returns a KeyError (see stacktrace below). I read [here](https://medium.com/@gonzafirewall/google-oauth2-and-django-rest-auth-92b0d8f70575) to modify the extract_uid method. How can I provide my adapter a modified GoogleProvider class? Or maybe the packaged provider should be fixed?

```
django_1        |   File "/usr/local/lib/python3.6/site-packages/rest_auth/views.py", line 104, in post
django_1        |     self.serializer.is_valid(raise_exception=True)
django_1        |   File "/usr/local/lib/python3.6/site-packages/rest_framework/serializers.py", line 236, in is_valid
django_1        |     self._validated_data = self.run_validation(self.initial_data)
django_1        |   File "/usr/local/lib/python3.6/site-packages/rest_framework/serializers.py", line 437, in run_validation
django_1        |     value = self.validate(value)
django_1        |   File "/usr/local/lib/python3.6/site-packages/rest_auth/registration/serializers.py", line 122, in validate
django_1        |     login = self.get_social_login(adapter, app, social_token, access_token)
django_1        |   File "/usr/local/lib/python3.6/site-packages/rest_auth/registration/serializers.py", line 57, in get_social_login
django_1        |     social_login = adapter.complete_login(request, app, token, response=response)
django_1        |   File "/app/project/users/adapters.py", line 47, in complete_login
django_1        |     login = self.get_provider().sociallogin_from_response(request, extra_data)
django_1        |   File "/usr/local/lib/python3.6/site-packages/allauth/socialaccount/providers/base.py", line 86, in sociallogin_from_response
django_1        |     uid = self.extract_uid(response)
django_1        |   File "/usr/local/lib/python3.6/site-packages/allauth/socialaccount/providers/google/provider.py", line 43, in extract_uid
django_1        |     return str(data['id'])
django_1        | KeyError: 'id'
```

steverecio:
Fix provided in #2444

matthijskooijman:
I just set up the Google provider here as well, which works as expected.

Tracing the allauth code, I see that the failing dict comes from `extra_data` here: https://github.com/pennersr/django-allauth/blob/5b9b85593e683da28897635e9179b41f7db95ee1/allauth/socialaccount/providers/google/views.py#L16-L26

[...]

PR #2444, opened 2020-01-18
Fix Google Provider

steverecio:
Fixes the GoogleProvider class as documented here: https://medium.com/@gonzafirewall/google-oauth2-and-django-rest-auth-92b0d8f70575


matthijskooijman:
I think this PR can be closed, see my comments in #2443.

Issue #2481, opened 2020-03-12
Untappd too many requests (but that's not the case)


adriaanh:
First, thanks for this amazing app. Really helps with implementing social logins into my app. 

I use Untappd for social login, however it seems that it is not working, no matter what I try, I get the 'Social login failure' screen. I did some debugging and with the `{{ auth_error.code }}` and `{{ auth_error.exception }}` template tags I was able to see the problem: `Code: unknown, Error: Error retrieving access token: b'Too Many Requests'`. 

I tried with two different apps from Untappd (I have two APIs), and both provide the same result. When I try the same credentials from the same machine in another app I developed myself in PHP, I don't get an error and the communication with Untappd is just fine. So it seems that there is no problem on the side of Untappd, and the callback URL is also okay (when I change the callback URL at Untappd with some random URL, I get the notification that the callback URL is not correct). 

As I think my settings are all correct, it seems that something with the request to Untappd is not correct, but I can't figure out what is wrong (looked at your code as well). What steps could I take to provide more insights in the specifics?

For reference my settings below:

Relevant parts of the settings.py:
```
INSTALLED_APPS = [
    'django.contrib.admin',
    'django.contrib.auth',
    'django.contrib.contenttypes',
    'django.contrib.sessions',
    'django.contrib.messages',
    'django.contrib.staticfiles',
    'django.contrib.sites',

    # Allauth apps
    'allauth',
    'allauth.account',
    'allauth.socialaccount',
    'allauth.socialaccount.providers.untappd',

    # own apps
    'core',
]

AUTHENTICATION_BACKENDS = (
    'django.contrib.auth.backends.ModelBackend',
    'allauth.account.auth_backends.AuthenticationBackend',
)

SITE_ID = 3

```
URLs:
```
urlpatterns = [
    path('', include('core.urls', namespace='core')),
    path('accounts/', include('allauth.urls')),
    path('admin/', admin.site.urls),
]
```
Site settings in DB (below is site 3):
```
Domain name: http://localhost
Display name: http://localhost
```
Setup of Social Application in DB:
```
Name: Untappd
Cliend id: <client id from untappd>
Secret key: <client secret from untappd>
Key: empty
Sites: http://localhost
```
Callback URL at Untappd:
```
http://localhost/accounts/untappd/login/callback/
```
Running the local server with the following command:
`python manage.py runserver localhost:80`

Edit: Added runserver commands

adriaanh:
[...]

PR #2486, opened 2020-03-14
Allow custom user agent for Untappd auth

adriaanh:
# Submitting Pull Requests

Fixes issue #2481. Updated documentation as well. 

## General

### DIFF
--- a/allauth/socialaccount/providers/untappd/client.py
+++ b/allauth/socialaccount/providers/untappd/client.py
@@ -32,7 +32,7 @@ class UntappdOAuth2Client(OAuth2Client):
         # Allow custom User Agent to comply with Untappd API
         settings = app_settings.PROVIDERS.get(UntappdProvider.id, {})
         headers = {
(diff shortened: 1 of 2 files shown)


allauth/socialaccount/providers/untappd/client.py
```diff
@@ -26,11 +29,16 @@ def get_access_token(self, code):
         if self.access_token_method == 'GET':
             params = data
             data = None
+        # Allow custom User Agent to comply with Untappd API
+        settings = app_settings.PROVIDERS.get(UntappdProvider.id, {})
+        headers = {
+            'User-Agent': settings.get('USER_AGENT', 'django-allauth-header')}
```
pennersr: Shouldn't the default agent be just "django-allauth" instead of "django-allauth-header" ?

adriaanh: Thanks for the feedback, seems that the default agent 'django-allauth' is indeed working just fine. Changed it in commit 9a22b4d

PR #2493, opened 2020-03-19
Fixed 'NoneType' object has no attribute 'strip'

LunatikCoder:
Hi, i found a bug when accessing the page {% url 'socialaccount_connections' %} (/social/connections/) it gave me a 'NoneType' object has no attribute 'strip' and the culprit is provider.py. The problem is that this file calls the strip() function on a null value and python being python doesn't like it ;) The solution i found is quite simple, in the provider.py file the MicrosoftGraphAccount class should be updated from this:

`class MicrosoftGraphAccount(ProviderAccount): def to_str(self): name = self.account.extra_data.get('displayName') if name.strip() != '': return name return super(MicrosoftGraphAccount, self).to_str()`

to this:

`class MicrosoftGraphAccount(ProviderAccount): def to_str(self): name = self.account.extra_data.get('displayName') if not name is None: if name.strip() != '': return name return super(MicrosoftGraphAccount, self).to_str()`

That's it !

### DIFF
--- a/allauth/socialaccount/providers/microsoft/provider.py
+++ b/allauth/socialaccount/providers/microsoft/provider.py
@@ -11,6 +11,7 @@ class MicrosoftGraphAccount(ProviderAccount):
         if not name is None:
             if name.strip() != '':
                 return name
+            return name
         return super(MicrosoftGraphAccount, self).to_str()
 
 
--- a/allauth/socialaccount/providers/microsoft/provider.py
+++ b/allauth/socialaccount/providers/microsoft/provider.py
@@ -11,6 +11,7 @@ class MicrosoftGraphAccount(ProviderAccount):
         if not name is None:
             if name.strip() != '':
                 return name
+        else:
             return name
         return super(MicrosoftGraphAccount, self).to_str()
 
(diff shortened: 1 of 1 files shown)


LunatikCoder:
I don't know why this is failling only for python 3.7.. the CI test should be updated, this change is working!

allauth/socialaccount/providers/microsoft/provider.py
```diff
@@ -8,7 +8,10 @@ class MicrosoftGraphAccount(ProviderAccount):
 
     def to_str(self):
         name = self.account.extra_data.get('displayName')
-        if name.strip() != '':
+        if not name is None:
+            if name.strip() != '':
+                return name
+        else:
             return name
         return super(MicrosoftGraphAccount, self).to_str()
```
pennersr: With these changes you never end up in the `return super()...` statement. So I think the only think that needs to be changed is the `if`. Change it into `if name and name.strip() != '':`

LunatikCoder: Thanks man, you were right ;) Just made the change and it passed test coverage !

PR #2499, opened 2020-03-28
fix: typo which prevents authentication with nextcloud

saemideluxe:
It seems this code has never been tested in production? Anyway, thanks for the great allauth package, it makes external authentication a joy!


saemideluxe:
Any chance this will get merge? I have another patch ready which fixes an exception if the nextcloud user has no email address set.

Mageti:
Awesome !
I can confirm this fixes the login with nextcloud.
I just lost days searching why I could not use my nextcloud to login to my django app.
When can we expect it to be released ?

PR #2501, opened 2020-03-29
Replace assert by RuntimeError

bittner:
Assert statements should not be used in production code.

- We should understand asserts as "comments that don't become outdated".
- Optimized Python code will have assert statements stripped out. (Try `python -O script.py`)

The intent of the original code was to ensure a value, hence `if`, and abort by raising an Exception otherwise.


bittner:
I'd like to extend [the tests](https://github.com/pennersr/django-allauth/blob/master/allauth/socialaccount/providers/facebook/tests.py), but I fail to understand which test code covers [the `if` and `else`](https://github.com/pennersr/django-allauth/blob/master/allauth/socialaccount/providers/facebook/provider.py#L63) branches. Can someone explain?

PR #2514, opened 2020-04-22
fix(instagram): Updated to use new API

zachmccormick:
This fixes the Instagram provider to work with the new Instagram API as
per the docs at https://developers.facebook.com/docs/instagram-basic-display-api

# Submitting Pull Requests

## General

 ## Provider Specifics
 
### DIFF
--- a/allauth/socialaccount/providers/instagram/provider.py
+++ b/allauth/socialaccount/providers/instagram/provider.py
@@ -9,9 +9,6 @@ class InstagramAccount(ProviderAccount):
     def get_profile_url(self):
         return self.PROFILE_URL + self.account.extra_data.get('username')
 
-    def get_avatar_url(self):
-        return self.account.extra_data.get('profile_picture')
-
     def to_str(self):
         dflt = super(InstagramAccount, self).to_str()
         return self.account.extra_data.get('username', dflt)
@@ -23,17 +20,16 @@ class InstagramProvider(OAuth2Provider):
     account_class = InstagramAccount
 
     def extract_extra_data(self, data):
-        return data.get('data', {})
+        return data
 
     def get_default_scope(self):
         return ['basic']
 
     def extract_uid(self, data):
-        return str(data['data']['id'])
+        return str(data['id'])
 
     def extract_common_fields(self, data):
-        return dict(username=data['data'].get('username'),
-                    name=data['data'].get('full_name'))
+        return dict(username=data.get('username'))
 
 
 provider_classes = [InstagramProvider]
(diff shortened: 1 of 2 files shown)


allauth/socialaccount/providers/instagram/provider.py
```diff
@@ -23,17 +20,16 @@ class InstagramProvider(OAuth2Provider):
     account_class = InstagramAccount
 
     def extract_extra_data(self, data):
-        return data.get('data', {})
```
zachmccormick, line 26: the shape no longer looks like

```json
{
  "data": {
    "id": "1234",
    "username": "foobar"
  }
}
```

allauth/socialaccount/providers/instagram/views.py
```diff
@@ -13,11 +13,12 @@ class InstagramOAuth2Adapter(OAuth2Adapter):
     provider_id = InstagramProvider.id
     access_token_url = 'https://api.instagram.com/oauth/access_token'
     authorize_url = 'https://api.instagram.com/oauth/authorize'
-    profile_url = 'https://api.instagram.com/v1/users/self'
+    profile_url = 'https://graph.instagram.com/me'
```
zachmccormick, line 16: https://developers.facebook.com/docs/instagram-basic-display-api/reference/me

allauth/socialaccount/providers/instagram/views.py
```diff
@@ -13,11 +13,12 @@ class InstagramOAuth2Adapter(OAuth2Adapter):
...
 
     def complete_login(self, request, app, token, **kwargs):
         resp = requests.get(self.profile_url,
-                            params={'access_token': token.token})
+                            params={'access_token': token.token,
+                                    'fields': ['id', 'username']})
```
zachmccormick, line 21: if you don't provide fields, you won't get the username back

zachmccormick:
The only addition here is that we might want to change the default scope. I don't think it's `basic` anymore

dorianfm:
@zachmccormick agree, the scope should be `user_profile` now, from what I can tell, instagram returns an error if `basic` scope is set.

marcosalcazar:
+1 at this pull request, plus the required change on the scope as both @zachmccormick and @dorianfm have mentioned already. Thanks for the effort Zach, I'm using your branch for now.

zachmccormick:
Ill try to update the scope and get this in a mergeable state this week unless someone else wants to make their own branch and get it mergeable

pennersr:
> The only addition here is that we might want to change the default scope. I don't think it's `basic` anymore

https://www.instagram.com/developer/authorization/

Here they still mention `basic` as the only scope:

image

dorianfm:
That’s the deprecated developer docs, you need to check out the ones on facebook.com for the current API

—

dorianfm:
see info on migration of API here https://www.instagram.com/developer/

—

mightym:
I just stumbled over this note while implementing the Instagram "Basic Display" in my django app:

> Note that Basic Display is not an authentication tool. Data returned by the API cannot be used to authenticate your app users or log them into your app. If your app uses API data to authenticate users, it will be rejected during App Review. If you need an authentication solution, use [Facebook Login](https://developers.facebook.com/docs/facebook-login/) instead.

So actually (as far as I understand this) Instagram isn't a way to authenticate a user anymore.

PR #2527, opened 2020-05-04
fix(twitch): Add missing client ID header

beheh:
This PR fixes the Twitch provider's `complete_login` method, which otherwise throws an exception.

Background: [Twitch is changing their API](https://discuss.dev.twitch.tv/t/requiring-oauth-for-helix-twitch-api-endpoints/23916?u=beheh) to require the `Client-ID` for all endpoints, including the profile URL endpoint. This change is going fully into effect on May 11th, and since May 1st Twitch is performing scheduled windows where this is in effect already.

Without this PR the OAuth login does not complete and instead throws the following error:

```
[twitch] Authentication error: 'unknown' (exception=OAuth2Error('Twitch API Error: Unauthorized (Client ID and OAuth token do not match)',))
```

I've tested the PR in my production application and this seems to fix it.

# Submitting Pull Requests

## General

 - [ ] ~~JavaScript code should adhere to [StandardJS](https://standardjs.com).~~
 - [ ] ~~If your changes are significant, please update `ChangeLog.rst`.~~
 - [ ] ~~Feel free to add yourself to `AUTHORS`.~~

### DIFF
--- a/allauth/socialaccount/providers/twitch/tests.py
+++ b/allauth/socialaccount/providers/twitch/tests.py
@@ -1,3 +1,6 @@
+from django.test.client import RequestFactory
+from django.urls import reverse
+
 from allauth.socialaccount.models import SocialToken
 from allauth.socialaccount.providers.oauth2.client import OAuth2Error
 from allauth.socialaccount.tests import OAuth2TestsMixin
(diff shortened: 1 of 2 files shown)


beheh:
I've also slightly cleaned up the tests to use a real Django request and the real app in favor of the mocks that were used previously, so that the tests pass.

beheh:
I believe the remaining test failure is due to master failing right now, not this PR.

PR #2568, opened 2020-06-16
feat(providers): Retrieve providers enabled on the site

BradLucky:
# Details of this change

The list of providers does not account for the fact that some sites may have an individual provider turned off. Instead, every site gets the full list of providers installed rather than just the ones that are enabled.

This adds a new method for getting only providers that are specifically enabled for the site.

It seems to me that this would have been the intended behavior all along, so it might be preferable to make this the default behavior. That might also inspire changes to `socialaccount.providers.ProviderRegistry.load()` to not load _everything_ from `INSTALLED_APPS`.

# Submitting Pull Requests

## General

 ## Provider Specifics
 
### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -28,6 +28,7 @@ Benjamin Jorand
 Biel Massot
 Björn Andersson
 Bojan Mihelac
+Brad Luczywo
 Bruno Alla
 Chris Beaven
 Chris Davis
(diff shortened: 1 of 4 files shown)


pennersr:
I agree with you that it would be better to change the default behavior of the `get_providers` tag, let's opt for that.

pennersr:
Fixed via cc5279bb61dba9cf0fafb10f4ae175c018749f1f

PR #2623, opened 2020-08-04
fix(forms): signup.custom_signup should use self instead of super due to MRO

### LINKED ISSUE #2435: CustomSignupForm method signup() doesn't get called
Thank you for django-allauth :-)

I tried creating a custom signup form like this
```
class CustomSignupForm(SignupForm):
    first_name = forms.CharField(max_length=30, label="First Name")
    last_name = forms.CharField(max_length=150, label="Last Name")
    company = forms.CharField(max_length=50, label="Company Name")

        print("hello "*10)
        user.first_name = self.cleaned_data["first_name"]
        user.last_name = self.cleaned_data["last_name"]

        user.profile.company = self.cleaned_data['company']
        user.save()
```
and I linked it in `settings.py` with
```
ACCOUNT_FORMS = {"signup": "users.forms.CustomSignupForm"}
```
I noticed that the form contains the additional fields but the `signup()` method doesn't get called. There ist no debug output and the `company` is not saved. `first_name` and `last_name` are saved, giving the impression that `signup()` is called when in fact they are saved by the adapter of the model. Took me a while to figure that out :-/. 

Having dug a bit deeper, I think the problem is due to the `custom_signup()` method of the `BaseSignupForm()` class especially this line
https://github.com/pennersr/django-allauth/blob/ef0774318993c2f7757df6ca2b962bc0df1d5674/allauth/account/forms.py#L350
Since `custom_signup()` is a method of `BaseSignupForm()` and not `SignupForm()` or even `CustomSignupForm()` the `super()` call always finds the `signup()` method of ` _DummyCustomSignupForm()`, which is empty.

Overwriting `custom_signup()` in `CustomSignupForm()` fixes the problem.
```
class CustomSignupForm(SignupForm):
    first_name = forms.CharField(max_length=30, label=_("First Name"))
    last_name = forms.CharField(max_length=150, label=_("Last Name"))
    company = forms.CharField(max_length=50, label=_("Company Name"))

    def custom_signup(self, request, user):
        custom_form = self #use local signup()
        if hasattr(custom_form, 'signup') and callable(custom_form.signup):
            custom_form.signup(request, user)

        user.first_name = self.cleaned_data["first_name"]
        user.last_name = self.cleaned_data["last_name"]

        user.profile.company = self.cleaned_data['company']
        user.save()
```

My question: Is this even the correct approach? [StackOverflow](https://stackoverflow.com/questions/36015534/how-django-allauth-signal-adapter-flow-works/36021818#36021818) tells me I should inherit from `forms.Form` and have a `signup()` method. [The documentation](https://django-allauth.readthedocs.io/en/latest/forms.html#signup-allauth-account-forms-signupform) tells me I should subclass `SignupForm` but have a `safe()` method. `custom_signup()` says `save()` is deprecated.

Im using Django 3, Python 3.7 and django-allauth 0.41.0

iarp:
Due to python MRO, a custom form inheriting from ``account.forms.SignupForm`` and using ``ACCOUNT_FORMS['signup']`` currently does not allow calling ``def signup`` due to the use of super which is calling the very base class which is ``_DummyCustomSignupForm`` by default instead of the custom form.

The check and use of signup should be based on self.

Related to #2039 because overriding the base class using ``ACCOUNT_SIGNUP_FORM_CLASS`` does call ``signup`` properly whereas ``ACCOUNT_FORMS['signup']`` does not due to MRO.

Fixes #2435 

# Submitting Pull Requests

## General

### DIFF
--- a/allauth/account/forms.py
+++ b/allauth/account/forms.py
@@ -347,16 +347,15 @@ class BaseSignupForm(_base_signup_form_class()):
         return cleaned_data
 
     def custom_signup(self, request, user):
-        custom_form = super(BaseSignupForm, self)
-        if hasattr(custom_form, 'signup') and callable(custom_form.signup):
-            custom_form.signup(request, user)
+        if hasattr(self, 'signup') and callable(self.signup):
+            self.signup(request, user)
         else:
             warnings.warn("The custom signup form must offer"
                           " a `def signup(self, request, user)` method",
                           DeprecationWarning)
             # Historically, it was called .save, but this is confusing
             # in case of ModelForm
-            custom_form.save(user)
+            self.save(user)
 
 
 class SignupForm(BaseSignupForm):
--- a/allauth/account/forms.py
+++ b/allauth/account/forms.py
@@ -355,7 +355,7 @@ class BaseSignupForm(_base_signup_form_class()):
                           DeprecationWarning)
             # Historically, it was called .save, but this is confusing
             # in case of ModelForm
-            self.save(user)
+            super(BaseSignupForm, self).save(user)
 
 
 class SignupForm(BaseSignupForm):


iarp:
Coveralls fails because I cannot figure out how to test the custom_signup class when the current base signup form contains def signup with pass. Python initializes the classes before ``@override_settings(ACCOUNT_SIGNUP_FORM_CLASS='...')`` can do its thing so the base class is always _DummyCustomSignupForm which just passes on the function instead of something like ``raise NotImplementedError``.

pennersr:
You're right -- closed via b8bf1c90

PR #2630, opened 2020-08-07
feat(email): add option to limit number of EmailAddress on an account

### LINKED ISSUE #2621: Feature request: Allow choice between using multiple email addresses and a single email address
Not all websites **need** multiple email addresses and this might prove to be overwhelming for some coupled with the user experience in having multiple email addresses. 
I think a setting should exists such as `ACCOUNT_USE_MULTIPLE_EMAILS` with a default value of true which allows websites choose between working with one email address and working with the default multiple email addresses that this package presently provides. 
I have a piece of hacky, untested code that provides this feature together with a patch for an [issue](https://github.com/pennersr/django-allauth/issues/2620) I raised that could solve this problem. 
As stated in the issue, the code is still in its hacky stages and would be improved upon by the community.

iarp:
Fixes #2621

General


9mido:
You are on a roll @iarp! Keep it up! You can make 2 other quick PRs (hopefully they are quick) if you fix the 2 email spamming issues I commented a solution to which are also pretty bad and need fixes.

stfenjobs:
please merge this pull  @pennersr 

thank you very much

ChangeLog.rst
```diff
@@ -7,6 +7,9 @@ Note worthy changes
 - If ``ACCOUNT_LOGIN_ATTEMPTS_LIMIT`` is set and the user successfully
   resets their password, the timeout is cleared to allow immediate login.
 
+- New setting ``ACCOUNT_EMAIL_LIMIT_ON_ACCOUNT`` lets you decide how many
```
pennersr, line 10: How about `ACCOUNT_MAX_EMAIL_ADDRESSES` ?

allauth/account/forms.py
```diff
@@ -432,6 +432,8 @@ def clean_email(self):
                               " with this account."),
             "different_account": _("This e-mail address is already associated"
                                    " with another account."),
+            'over_account_limit': _("You cannot add more email addresses as "
+                                    "you are at maximum allowance.")
```
pennersr, line 436: Would be nice if this message simply said: "You cannot setup more than %d e-mail addresses.".

allauth/account/forms.py
```diff
@@ -441,8 +443,15 @@ def clean_email(self):
             raise forms.ValidationError(errors["this_account"])
         if on_diff_account and app_settings.UNIQUE_EMAIL:
             raise forms.ValidationError(errors["different_account"])
+        if self.is_over_account_email_limit():
+            raise forms.ValidationError(errors['over_account_limit'])
         return value
 
+    def is_over_account_email_limit(self):
+        email_limit = app_settings.EMAIL_LIMIT_ON_ACCOUNT
+        users_emails = self.user.emailaddress_set.count()
```
pennersr, line 452: You could skip doing this DB query if there is no limit configured.

allauth/templates/account/email.html
```diff
@@ -44,14 +44,19 @@ <h1>{% trans "E-mail Addresses" %}</h1>
 
 {% endif %}
 
-
+  {% if can_add_emails %}
     <h2>{% trans "Add E-mail Address" %}</h2>
 
+    {% if account_email_limit and account_email_limit > 1 %}
```
pennersr, line 50: From a less-is-more perspective, I think it is more user friendly to not constantly present this limit to the user. Likely, under normal circumstances, users should not be hitting the limit anyway... so I would drop this.

docs/configuration.rst
```diff
@@ -89,6 +89,16 @@ ACCOUNT_EMAIL_MAX_LENGTH(=254)
   3-byte and 4-byte Unicode character sets
   <https://dev.mysql.com/doc/refman/5.5/en/charset-unicode-conversion.html>`_.
 
+ACCOUNT_EMAIL_LIMIT_ON_ACCOUNT(=None)
+  Limit how many EmailAddress records a user can add to their account.
+
+  Will not affect users that already have more than the limit after changing
+  this value later on in your projects lifespan.
+
+  Warning: An account must have at least One email address. A user cannot
```
pennersr, line 98: "One" caps.

docs/configuration.rst
```diff
@@ -89,6 +89,16 @@ ACCOUNT_EMAIL_MAX_LENGTH(=254)
...
+  remove the only email address on their account. If you set this option to 1
+  they will be unable to ever add or change their current email address.
```
pennersr, line 100: This is true -- so shouldn't we just raise ImproperlyConfigured when the limit is set to 1 ?

pennersr (changes requested):

docs/configuration.rst
iarp: I'm not sure where it would be best to raise the error, in AppSettings init?

pennersr: You are already asserting that it is > 0 there -- why not make that > 1 ?

9mido:
I would like to add that personally I would want MAX_EMAIL_ADDRESSES=1 to prevent a user from adding an unverified email address that is not their own then spamming that other person's inbox by spam clicking the "re-send verification" button. This helps with CAN-SPAM and prevents the spam victim from saying "why did I receive this email from this website without opting in"? As for others who want more than 1 email per account plus all the other possible actions, the other timeout setting would be needed anyways just as an abuse prevention mechanism.

pennersr:
Closed via 14c52ed1 -- thanks!

9mido:
Yay! Thank you :)

9mido:
@iarp I just tried testing this new setting and I have the following in my settings.py file:

ACCOUNT_MAX_EMAIL_ADDRESSES=1

In requirements.txt I have: 

git+https://github.com/pennersr/django-allauth

But on /accounts/email/ I can still add as many unverified emails as I want as if the setting does not work. 

Am I doing something wrong here?

9mido:
@pennersr 

I see in that PR you have - "You can now limit the amount of email addresses a user can associate to his
  account by setting ``ACCOUNT_MAX_EMAIL_ADDRESSES``.

But when I look at @iarp's code I see MAX_EMAIL_ADDRESSES.

When I do:

pip3 install git+https://github.com/pennersr/django-allauth

and in my settings file:

MAX_EMAIL_ADDRESSES=1

or 

ACCOUNT_MAX_EMAIL_ADDRESSES=1

on the /accounts/email/ page, I am able to add as many unverified emails as I want. It does not seem to be working for me. Am I doing something wrong here? 

Please let me know.

iarp:
Working fine for me. Confirmed to be `ACCOUNT_MAX_EMAIL_ADDRESSES`.

9mido:
@iarp thank you for the confirmation. 

I did a fresh restart of my live linux USB and installed all my libraries from scratch from my wiped system. 

Not sure if the issue was upgrading from the version on PyPi to the new one using pip3 install git+https://github.com/pennersr/django-allauth command.

It does indeed work after I did installed from scratch on a new system. 

As always, and once again, thank you!

PR #2631, opened 2020-08-07
feat(timeouts): add timeout system to prevent spamable actions

### LINKED ISSUE #2167: Password Reset Emails Not Subject to Cooldown
There appears to be no rate-limiting for password resets.

I just got hit by a bot who sent themselves 5,000 confirmation emails and got my email service to temporarily suspend me, due to #1921.  Now that I got that worked around by disabling email HMAC as described in that issue, I was testing things and noticed that even though those confirm emails are now rate limited -- the password reset emails do NOT appear to be rate-limited.  This is a vulnerability which we should consider protecting in similar manner. However this feature would probably require another DB table.

Using allauth 0.38.

### LINKED ISSUE #2595: Spam Clicking Email Re-send Verification Button
Update: A solution for this issue is below and there is a pull request but a merge is needed. 

Clicking really fast on form buttons such as the /accounts/email/ form re-send verification button for example will allow someone to spam a user's inbox over and over and fill it up quickly. 

If using a free email provider like gmail, there is a limit of how many emails you could send per day. But if this bug is abused, a business might get emails temporarily suspended and not be able to send any emails once that limit is reached which could be disastrous. 

There should be a way to disable/hide this button after submitting and after the form's action page refresh. Or set a timer on how long each user has to click that button again. Might need to use localStorage if using JavaScript. Otherwise something on the django side might need to be done which would be better than JavaScript.

Examples of fixing this would be like Reddit and Stackoverflow posts where users are not able to post again until X minutes after the original post. 

https://i.stack.imgur.com/4J3jr.png

If an email address is already confirmed, the re-send verification button should be disabled unless the user adds another email address and selects the radio button to modify the new email address. There is no need for an existing email address to keep getting verification emails if it has already been verified.

iarp:
This is what I would consider for an email timeout system to help prevent someone from sending too many emails.

Currently the ResetPasswordForm will send a potentially unlimited number of emails every time someone fills out the form. This PR adds an option to limit how often someone can request the reset thereby limiting the number of emails they could potentially send.

At this time its only available for use on ResetPasswordForm, and ResendEmailVerification as I'm not overly sure of its usefulness anywhere else. There is a note in configuration docs stating if someone has overridden this form in particular they may need to apply the necessary code themselves.

There is a new model called EmailTimeout in the account app. If someone was malicious enough to use a bot they could very well avoid the session system thereby avoiding any sort of session storage for the limiting of emails. The model has been built under the premise of being able to be used in other locations based on the action field.

Resolves #2595 resolves #2167


9mido:
@iarp is a hero! 

So will this account for both the password reset form and the re-send verification button when spam clicked for emails on the /accounts/email/ page? 

Session variables should not be used for something like this since they're only saved on response. If you modify a session's contents, it isn't visible to other requests until the view returns its response so if someone starts 5 requests to send the email, they could all see the same session state.

pennersr:
I have some doubts whether introducing a new model for this is the best way forward. It would just as well be possible to simply keep track of this in a (Django) cache. Keeping it in the database may require a janitor process as well. All in all, I think it would be best to wrap up all timeout related logic in an adapter method, which by default is using the simplest implementation (cache). Then, should users feel the need for more, they can always override the adapter and hookup to their own model.

iarp:
I've updated the system to use djangos cache and moved everything into the adapter to be easily overridable.

At the same time I renamed the methods to be more agnostic (removing `_EMAIL_` from everything) since technically the system could easily be adapted to other actions in the future, all it requires is the action and a value to set a timeout on.

9mido:
@pennersr 

iarp kindly made the changes you requested. Can this now be merged?

docs/configuration.rst
```diff
@@ -89,6 +89,33 @@ ACCOUNT_EMAIL_MAX_LENGTH(=254)
...
+  someone fills it out. Setting the timeout will prevent someone requesting
+  a password reset without waiting.
+
+  In this example, restrict the ResetPasswordForm to only allow sending an
+  email every 5 minutes::
+
+    ACCOUNT_TIMEOUTS = {
+        'ResetPasswordForm': 5*60,
+        'ResendEmailVerification': 5*60,
+        'ChangePasswordForm': 10,
```
pennersr, line 107: I think that the use case for these actions is solely limited to the sending of emails, right? The email is used as cache key after all. Though, the naming of the setting suggests that this is a more general concept ("certain action a user can do"). So, I am wondering if this setting should not be named ACCOUNT_EMAIL_TIMEOUTS.

At the same time, I am wondering how one would configure these values. If the goal is to prevent people from triggering too many mails, why do we need multiple individually configurable values instead of having just one general email cooldown value.

We now have a EMAIL_CONFIRMATION_COOLDOWN setting -- wouldn't renaming this into a general EMAIL_COOLDOWN setting and using it all over where emails are sent suffice?

9mido: @pennersr 

Yes the sending of emails since all of those actions (reset password, resend email verification, change password) involve sending an email. 

That is not a bad idea to rename the ACCOUNT_TIMEOUTS setting to EMAIL_COOLDOWN where the one assigned value to that setting controls all 3 actions instead of keeping them separated. As long as this setting is separate from the EMAIL_CONFIRMATION_COOLDOWN setting that seems reasonable.

From a UX standpoint though, does it make sense to only provide 1 value to the EMAIL_COOLDOWN setting to control all 3 email actions? Would that one value for one email action make sense everywhere else for the other email actions? If not, then keep in mind an allauth programmer would need to change the one value for all 3 email action places which will then get affected all the same. The end user (site visitor) would need to wait that single allotted time for all 3 actions. 

I imagine an end user scenario like this - "I tried resending verification too many times and now I can't change my password I need to wait the same amount of time to do anything else". Versus "I tried resending verification too many times and now at least I can change my password but need to wait less time to do it". 

Not sure if the PR here works like that scenario above though.

Should one of the following things happen?
1) Site visitor can perform action A x number of times per minute. Then action A only gets timed out. Actions B and C do not get timed out until the user performs them. 
2) Site visitor can perform action A x number of times per minute. Then actions A, B, and C all get timed out for the same amount of time.
3) Something else? 

I am fine with either combining the actions or separating the actions into that EMAIL_COOLDOWN setting as 1 value or multiple values though. I leave that to you all to decide. 

@iarp 

Curious to hear your opinions on how this should be handled.

iarp: At first it was written to be an email cooldown type system but as I wrote it up I realized it could be applied to various other number of things not just email, for instance limiting how fast someone can do a certain action, such as changing a password.

9mido: @iarp Right, so are you leaning towards penners idea or keeping each action configured separately? From a UX standpoint, does it make sense to time out everything for the same time period as long as one action gets abused? I imagine users might not like that if that is the case because they have to wait to do the other actions but not 100% sure if that is what penners means. How does your code currently work? Does it timeout an action as long as that one action gets abused? No other action gets timed out if one is abused? Or does your code do something else?

iarp: I'm starting to think this is growing beyond allauth's scope.

The adapter methods are written in a way that means they could be used for any sort of timeout purpose. Which is why I included the ChangePasswordForm as an example, I personally cannot find any reason within me that I would ever time that out but its an example of a non-email based thing happening with timeout restrictions.

It's setup to allow an admin to set custom timeouts because I think certain actions should be allowed faster than others. Resetting a users password via email requires time, the user needs to allow enough time for smtp connection, propagation, checking spam folders.. etc. But then if I use the changing password example, why restrict that to 5 minutes? 30 seconds should be long enough. The real question for someone hitting the change password timeout is why are they changing a password so fast like that?

9mido:
@iarp 

Not sure if this is included but the user could continuously log in and log out with an unverified email address and cause a verification email to be sent each time the user logs in. There should be a limit on the number of verification emails sent each day per user for the log in process. 

I really hope this is not beyond scope and I really hope this gets merged. Sorry for the delay of this comment and for anything that I may have said above to make it seem like it would be beyond scope.

9mido:
@iarp I was able to solve the issue for continuously logging in with an unverified email when ACCOUNT_EMAIL_VERIFICATION = "mandatory". But it is not as fancy as something you would be able put together into an additional update to this PR. As for an anonymous user continuously signing up, not so sure how to do that. 

https://github.com/pennersr/django-allauth/blob/master/allauth/account/utils.py#L143-L200

In the function above, add the following code:

```
    #from django.http import HttpResponseServerError
    action = 'LoginForm'
    if adapter.timeout_status(user.email, action):
        messages.error(request, "Please wait before trying again.")
        return HttpResponseRedirect("/accounts/login/")
        #return HttpResponseServerError("Please wait before trying again.")
    else:
        adapter.timeout_apply(user.email, action)
```

Settings.py:

```
ACCOUNT_TIMEOUTS = {'
    'LoginForm': 5*60, 
}
```

isms:
The exact issue (#2595) that this PR solves just caused our domain to be temporarily blocked by our email sender, triggering an outage. We use reCAPTCHA, but we couldn't prevent thousands and thousands of "Please Confirm Your E-mail Address" emails being sent.

image

We banned the particular offender, of course, but please consider this a respectful endorsement for a speedy if not completely perfect solution from an otherwise extremely happy allauth user of ~6 years!

ogurec-ogurec:
@9mido Hi, thanks for a great job - that's awesome! I have a small question, I set the higher-level commits, everything seems to work, but the ACCOUNT_MAX_EMAIL_ADDRESSES = 1 parameter does not react. I change its value, but there is no reaction. That is, the form remains, the form for adding an email and an email is added. What to do / see? I will be glad for any help

9mido:
@ogurec-ogurec Thanks but I don't deserve all of the credit @iarp is the brains behind this email timeouts coding operation and the ACCOUNT_MAX_EMAIL_ADDRESSES setting. You should be thanking him too. I simply found this issue and helped him think of stuff he might not have thought of and he did everything else. I would have never been able to code something advanced like this.

I think the email spam is the worst problem allauth has right now according to the graph above and this timeout mechanism should be a top priority to fix that problem. It isn't even worth using allauth if an email sender could block your domain and prevent any more users from signing up. 

@pennersr Any chance @iarp's timeouts PR could get merged yet or perhaps my other PRs could be merged? My code above is an addon code for this timeout PR for timing out logins and signups and should also be included as well with this timeout PR because the spam could come from any action that has to do with emails or other actions not even having to do with email which could be added to future PRs. @iarp has not been around lately to update his branch with these changes I wrote above so not sure what to do.

ogurec-ogurec:
> @iarp I was able to solve the issue for continuously logging in with an unverified email when ACCOUNT_EMAIL_VERIFICATION = "mandatory". To reproduce, sign up with an unverified email and continuously try to login without verifying. This code below will also solve when a user continuously signs up from their session since they are Anonymous user. But it is not as fancy as something you or someone else would be able put together into an additional update to this PR. Feel free to make changes. Showing the users how much time is remaining would be a good change too (but that can be a later on thing).
> 
> https://github.com/pennersr/django-allauth/blob/master/allauth/account/utils.py#L143-L200
> 
> In the perform_login function from the first link above, add the following code:
> 
> For continuous logins:
> 
> ```
>     #from django.http import HttpResponseServerError
>     action = 'LoginForm'
>     if adapter.timeout_status(user.email, action):
>         messages.error(request, "Please wait before trying again.")
>         return HttpResponseRedirect("/accounts/login/")
>         #return HttpResponseServerError("Please wait before trying again.")
>     else:
>         adapter.timeout_apply(user.email, action)
> ```
> 
> Settings.py:
> 
> ```
> ACCOUNT_TIMEOUTS = {'
>     'LoginForm': 5*60, 
> }
> ```
> 
> For continuous signups:
> 
> https://github.com/pennersr/django-allauth/blob/master/allauth/account/utils.py#L203-L216
> 
> In the complete_signup function from the second link above, add the following code:
> 
> ```
>     adapter = get_adapter(request)
>     #from django.http import HttpResponseServerError
>     action = 'SignupForm'
>     if adapter.timeout_status(request.session.session_key, action):
>         messages.error(request, "Please wait before trying again.")
>         return HttpResponseRedirect("/accounts/signup/")
>         #return HttpResponseServerError("Please wait before trying again.")
>     else:
>         adapter.timeout_apply(request.session.session_key, action)
> ```
> 
> Settings.py:
> 
> ```
> ACCOUNT_TIMEOUTS = {'
>     'SignupForm': 5*60, 
> }
> ```

I have another question for you, we have a setting: `ACCOUNT_LOGIN_ATTEMPTS_LIMIT`, it limits the number of login attempts. BUT, the limitation is only for a specific user. If I try to log in after the limit for another user,the system will skip it. What do you think can be done that in principle limit the number of login attempts, and not for a specific user? @9mido @pennersr

9mido:
@ogurec-ogurec Yes ACCOUNT_LOGIN_ATTEMPTS_LIMIT is for a specific user of any type if they fail to login after a certain amount of times. With ACCOUNT_TIMEOUTS = {''LoginForm': 5*60, } for example it should only occur for users who have unverified emails (if not _has_verified_for_login) to prevent unverified users trying to login from sending lots of spam emails if they try to brute force your form. 

If you wanted to, you could apply the timeout for every user who tries to login (verified, unverified, active, inactive, superuser, etc) but we already have ACCOUNT_LOGIN_ATTEMPTS_LIMIT for that and would need more changes to this PR. All the ACCOUNT_TIMEOUTS setting does is it makes it more strict where the user only has 1 try to do something before getting timed out for a set amount of time. 

The ACCOUNT_LOGIN_ATTEMPTS_LIMIT should still apply for those unverified users plus other user classifications separately. The unverified users trying to login though with both settings will have it more strict though because they will be hit with a timeout from ACCOUNT_TIMEOUTS after every single login attempt rather than a set amount of attempts from ACCOUNT_LOGIN_ATTEMPTS_LIMIT. 

It sounds like you want to prevent logins for all users? That sounds really strict and would not be a good user experience if everybody cannot login if one user ruins the party for everyone else. You might want to use something like the django-lockdown library for that.

ogurec-ogurec:
@9mido thanks for the detailed answer. By the script deny login attempts for all users, I meant the following script:

We have the following settings:
```
ACCOUNT_LOGIN_ATTEMPTS_LIMIT = 3
ACCOUNT_LOGIN_ATTEMPTS_TIMEOUT = 60
```

1.We have a list of users, for example (John, Williams, Admin)
2. I'm trying to log in (bruteforce password) as user John.
3. Enter password incorrectly 3 times for user John.
4. I get the limitation ACCOUNT_LOGIN_ATTEMPTS_LIMIT = 3.
5. I get a timeout of 60 seconds (ACCOUNT_LOGIN_ATTEMPTS_TIMEOUT = 60)
6. While the timeout is passing, I can do the same (point 2,3) for the next user, for example Williams.

Thus, the timeout is valid for a specific user. It seems to me that it would be a good idea to limit the number of attempts in general (from a specific ip address or maybe by session), and not for a specific user (John).

What do you think about this?

9mido:
@ogurec-ogurec Now I understand what you are saying. I think that what you are recommending is a good idea. Preventing brute force based on IP address / session so that it completely stops the same attacker from logging in by trying multiple different accounts and their passwords. This can be applied in other places besides logins too like passwords resets, signups, etc. 

It seems like other django libraries like django-axes and django-defender could do what you are saying for logins. I say try them out and report back. You are also encouraged to make another django-allauth issue to have something like this built in though without needing to rely on these other libraries if the others say it is out of scope for this PR. It could always be a future PR perhaps. 

In the meantime though, I leave it up to @iarp and @pennersr if they want to timeout with ACCOUNT_TIMEOUTS based on IP address / session. It should probably be another setting to work in conjunction with the timeouts setting of this PR to catch the spammer to completely time them out no matter how many accounts they try it with. Otherwise maybe timing out by IP address / session could be built in by default and not need another setting.

docs/configuration.rst
9mido: I agree with this. Sorry for making things more confusing.

9mido:
Here is a method of how to do it with session variables:

```
from django.http import HttpResponseServerError
import datetime

    current_time = datetime.datetime.now()
    try:
        last_send_time = request.session.get('last_email_time',None)
    except:
        pass
    else:
        last_send_time = request.session.setdefault('last_email_time',None)
    print(current_time)
    print(request.session['last_email_time'])
    print(last_send_time)
    if last_send_time is not None:
        diff = current_time.timestamp() - last_send_time
        print(diff)
        minutes_diff = divmod(diff, 60)
        if minutes_diff[0] > 1:
            print('form can be submitted again')
        else:
            return HttpResponseServerError("Please wait")

            request.session['last_email_time'] = current_time.timestamp() #put at end of form when submitted
```

9mido:
For IP addresses, using ipware is probably the best bet. 

I recommend everybody reading this entire page even the comments:

vabene1111:
So if anyone stumbles across this and needs an easy and flexible solution right now and not when this is properly merged you can do the following: 

override the default account adapter and enable any of the django cache backends (alternatively create yourself a DB model)
```
ACCOUNT_ADAPTER = 'CustomAdapter'

CACHES = {
    'default': {
        'BACKEND': 'django.core.cache.backends.locmem.LocMemCache',
        'LOCATION': 'default',
    }
}
```

In the custom adapter override the `send_email` function like this 
```
    def send_mail(self, template_prefix, email, context):
        if settings.EMAIL_HOST != '': # only send email if host is configured
            default = datetime.datetime.now()
            c = caches['default'].get_or_set(email, default, timeout=360) # set a cache entry on the email address (360 is the timeout in seconds)
            if c == default:
                super(AllAuthCustomAdapter, self).send_mail(template_prefix, email, context)
            else:
                messages.add_message(self.request, messages.ERROR, _('In order to prevent spam, the requested email was not send. Please wait a few minutes and try again.')) # add a message because the success page is displayed and a user needs to know the email was not send
        else:
            pass
```

This solution is faaaar away from being perfect but it can easily be build upon. You could for example not show the real success page but implement a redirect higher up in the adapter. You can also use the ip address of the sender as the cache key instead of the email if you want to protect against that.

I appreciate all the work the authors here have put into allauth and love the project, i am sure a proper solution will one day make this hack obsolete :)

adsee42:
@vabene1111 
Thanks for the idea.
Using cache seems a great solution for now.

Instead record the timestamp only when sending email, I extended this to password reset email, password reset itself, and of course  verification email.

9mido:
If using the cache, how do you handle cache invalidation?

vabene1111:
as long as the cach is not invalidated every few seconds/minutes its fine, i actually use the default timeout function of the cache to set the application timeouts

9mido:
According to pennersr in one of my now closed PRs, he mentions exponential backoff as another solution we could use for this. I am not sure if we should implement it in this PR or a separate PR to show the differences in both. 

Maybe we could use ->  https://pypi.org/project/backoff/

For others who don't know what exponential backoff is, it means:

The rate of attempts decreases exponentially with the increasing number of failures. 

Or in our case, as an increasing number of spammable emails go through, the less often a user could do it.

Protection against spamming would allow some rate of sending emails up to the cutoff where abuse might be suspected (probably some number of emails sent in a fixed window), and from there the allowed rate of sending would be reduced.

We would probably need new allauth settings for allowing customization of a timeout, cutoff, exponent, etc. if we were to do something with exponential backoff.

pennersr:
The concept of timeouts is a bit limiting. For certain actions, such as sending confirmation emails, it makes perfect sense to only send one confirmation email every 5 minutes or so. But, I don't think that concept is usable everywhere else. For other situations, such as requesting password resets, signing up or logging in, it is perfectly fine to fine to do this a few times in a row, but just not too often. For those situations, we need to resort to true rate limiting, and also taking the remote IP address into consideration.

So, in order to move this idea of timeouts forward I've turned it into a rate limiter, see: 116b7a89. In that commit, only password reset is covered, but using the decorator approach it can (and will) be easily applied elsewhere.

PR #2635, opened 2020-08-10
test(test_settings): decrease password iterations to increase test speed

iarp:
I'm finding tests currently take roughly 200s to complete. That can be reduced greatly by decreasing the PBKDF2 iterations using a custom hasher. This is only applied to test_settings.py and should never be used in production.

Standard Iterations: Ran 525 tests in 227.867s
Decrease Iterations: Ran 525 tests in 17.950s

## General

### DIFF
--- a/test_settings.py
+++ b/test_settings.py
@@ -161,7 +161,9 @@ from django.contrib.auth.hashers import PBKDF2PasswordHasher
 
 class MyPBKDF2PasswordHasher(PBKDF2PasswordHasher):
     """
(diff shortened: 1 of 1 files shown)


9mido:
@iarp Your PRs are shaping up to be quite possibly one of allauth's biggest updates of all time :1st_place_medal: 

What do you think about issues 2360, 2525, and 615? Hopefully they can also be quick to help you break some more coding records. Otherwise, looking forward to other ideas you have in your head like this one you just made :+1:

PR #2636, opened 2020-08-10
fix(password/confirm): logout user A if user b clicks password or confirmation link

### LINKED ISSUE #615: Email verification when a different user is already logged in
when a user is already logged in and some other user user on the same browser verifies  his email address (by clicking the verification link in his email). 

it creates problem when
ACCOUNT_LOGIN_ON_EMAIL_CONFIRMATION = True

it does not Sign out the logged in user ( the logged in user is not changed )

is this the expected behavior.

### LINKED ISSUE #2302: How can I receive user data upon successful login?
Currently I can pass it in the logged_in.txt and into a messages cookie, I would like to return the user data as a json response.like to return the user data as a json response.

### LINKED ISSUE #2303: Password reset when another user is logged-in
If user-A is logged-in and another user-B attempts to reset their password, it succeeds while user-A stays logged-in. It is a bit confusing, although I'm not sure if there's a security issue here. It would be better to invalidate the current active session in such case. 

Somewhat like this issue : https://github.com/pennersr/django-allauth/issues/615

iarp:
Resolves #615
Resolves #2303

If User B requested an email confirmation or a password reset and then later on that day User A logs in (same machine same browser). User B comes along and clicks the links while User A is still logged in on the same browser. Currently the system will correctly reset the password or confirm the email for the correct account however User A is left still logged in.

This PR checks that if the currently user is not the same user that requested the action, logout the current user.

## General


allauth/account/views.py
```diff
@@ -279,6 +290,15 @@ def get(self, *args, **kwargs):
     def post(self, *args, **kwargs):
         self.object = confirmation = self.get_object()
         confirmation.confirm(self.request)
+
+        # In the event someone clicks on an email confirmation link
+        # for one account while logged into another account,
+        # logout of the currently logged in account.
+        cur_user = self.request.user
+        if cur_user.is_authenticated and \
+                cur_user != confirmation.email_address.user:
```
pennersr, line 299: Can you turn this into `if cur_user.pk != confirmation.email_address.user_id` -- that avoids potentially dereferencing a foreign key and avoids the check on `is_authenticated.

allauth/account/views.py
```diff
@@ -690,6 +712,15 @@ def dispatch(self, request, uidb36, key, **kwargs):
                 data={'uidb36': uidb36, 'key': self.key})
             if token_form.is_valid():
                 self.reset_user = token_form.reset_user
+
+                # In the event someone clicks on a password reset link
+                # for one account while logged into another account,
+                # logout of the currently logged in account.
+                cur_user = self.request.user
+                if cur_user.is_authenticated and cur_user != self.reset_user:
```
pennersr, line 720: Here you can use `pk` based comparison as well.

pennersr (changes requested):

allauth/account/views.py
iarp: The logout should only be called if a user is already logged in. Using pk != user_id matches on other occasions. Trying to change this causes account.tests.ConfirmationViewTests to fail due to session variables being deleted. I need to at least check that cur_user.pk is true.

9mido: @pennersr Your thoughts based on what iarp said?

pennersr: I am not sure I understand the response. My main point is that doing `confirmation.email_address.user` needlessly invokes a DB query to fetch the user -- comparison on `email_address.user_id` will avoid that.

iarp: Ok I understand what you're talking about now. Your original comment removed the is_authenticated check which breaks logic used in views tested via ConfirmationViewTests. The is_authenticated check is necessary as this logic should only apply when someone is already logged in.

pennersr:
Can you have a look at the conflicts?

pennersr:
Closed via 063f1be7 -- thanks

PR #2650, opened 2020-08-25
feat(login ajax view): add available social providers to ajax response data

iarp:
Current ajax based view lists the form and the required fields but does not list the currently enabled providers. Useful when using JS to build the page with its field and you want to add the providers.

## General

### DIFF
--- a/docs/forms.rst
+++ b/docs/forms.rst
@@ -90,11 +90,11 @@ Used on `account_email <views.html#e-mails-management-account-email>`__ view.
     from allauth.account.forms import AddEmailForm
     class MyCustomAddEmailForm(AddEmailForm):
 
-        def save(self):
+        def save(self, request):
 
             # Ensure you call the parent class's save.
             # .save() returns an allauth.account.models.EmailAddress object.
-            email_address_obj = super(MyCustomAddEmailForm, self).save()
+            email_address_obj = super(MyCustomAddEmailForm, self).save(request)
 
             # Add your own processing here.
 


9mido:
I love how you are fixing all the core issues @iarp - it is so important (more important than social aspects of the app IMO).

What do you think of issues 1872, 1307, 252, 1522, 1921, 2360? I am trying to find other core ones that could help make allauth more bulletproof.

allauth/account/views.py
```diff
@@ -13,7 +13,9 @@
 from django.views.generic.base import TemplateResponseMixin, TemplateView, View
 from django.views.generic.edit import FormView
 
+from ..app_settings import SOCIALACCOUNT_ENABLED
 from ..exceptions import ImmediateHttpResponse
+from ..socialaccount import providers
```
pennersr, line 18: It's best not to introduce a dependency from `account` to `socialaccount` as that will result in both modules to have a cyclic dependency upon each other.

allauth/account/views.py
```diff
@@ -174,6 +176,39 @@ def get_context_data(self, **kwargs):
                     "redirect_field_value": redirect_field_value})
         return ret
 
+    def get_ajax_data(self):
+
+        enabled_providers = {}
+
+        if SOCIALACCOUNT_ENABLED:
+
+            for provider in providers.registry.get_list(self.request):
```
pennersr, line 185: In the current main branch, this should become `adapter.list_providers(request)`.

pennersr (changes requested):

PR #2666, opened 2020-09-23
fix(docs): updated installation, overview, and test_settings.py with missing providers

iarp:
Various documentation files are missing providers as requested in pull request templates.


pennersr:
Merged via 05c011a2, thanks.

PR #2701, opened 2020-10-30
feat(adapter): SocialAccount model str returns value from adapter

iarp:
Adding method get_socialaccount_model_str to DefaultSocialAccountAdapter so that SocialAccount's __str__ method can be customized. As it currently stands it returns self.user which by default only returns the username.

## General


iarp:
@pennersr can we get this in or is there something wrong?

PR #2705, opened 2020-11-05
fix(apple): documentation error on certificate key id

dany-nonstop:
if use the ```certificate_key``` as the key, sign in with Apple ID would fail with error message ImproperlyConfigured Apple 'certificate' missing". From the code it seems to be ```certificate```, located at line 34 of socialaccount/apple/client.py.

### DIFF
--- a/docs/providers.rst
+++ b/docs/providers.rst
@@ -153,7 +153,7 @@ Add the following configuration to your settings:
                 "key": "MEMAPPIDPREFIX",
 
                 # The certificate you downloaded when generating the key.
-                "certificate_key": """-----BEGIN PRIVATE KEY-----
+                "certificate": """-----BEGIN PRIVATE KEY-----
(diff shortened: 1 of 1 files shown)


docs/providers.rst
```diff
@@ -153,7 +153,7 @@ Add the following configuration to your settings:
                 "key": "MEMAPPIDPREFIX",
 
                 # The certificate you downloaded when generating the key.
-                "certificate_key": """-----BEGIN PRIVATE KEY-----
+                "certificate": """-----BEGIN PRIVATE KEY-----
```
kelkarajay, line 156: This is intentionally changed to `certificate_key` from the looks of it > https://github.com/pennersr/django-allauth/commit/e43bfd96039eea009925888ab8e4efb8428a69e1#

kelkarajay (changes requested):

pennersr:
You're likely reading/applying master docs while running 0.43.0.

Issue #2027, opened 2018-05-21
'OneToOneRel' object has no attribute 'to' while using password reset (custom template)


ghost:
`/home/patryk/workspace/.virtualenvs/project/lib/python3.6/site-packages/allauth/account/utils.py in url_str_to_user_pk, line 421`

Python 3.6.5, 3.5.5 doesn't work too.

zvolsky:
Same with Email Confirmation after SignUp.
In current allauth version 0.43.0 the code is on the line 470.
This change would help:
if getattr(User._meta.pk, "remote_field", None) and getattr(User._meta.pk.remote_field, "to", None):

pennersr:
Closed via f4618614

pennersr:
https://codeberg.org/allauth/django-allauth/issues/2027

PR #2719, opened 2020-11-22
hotfix issues/2027: avoid failure with missing .to attribute

zvolsky:
The old code fails with missing .to attribute.
This can be fixed with just testing if .to is present (this PR),
however I am not sure for the values leading to the failure if the code from 'else:' branch (ie. pk_field = User._meta.pk) is ok.

Please see at least the issue
https://github.com/pennersr/django-allauth/issues/2027
where example of values leading to the crash can be found.


pennersr:
Closed via f4618614

Issue #754, opened 2014-11-05
Verification link to a verified email


mazing:
In the demo I can send a verification mail even though the email address is verified. Shouldn't it checked if email address is already verified before sending another email?

pawciobiel:
Also it would be better if EmailAddress.send didn't use request object at all - simple query against DB would be just fine.

9mido:
To the _action_send function inside the EmailView class add this code:

Views.py:

```
from django.http import HttpResponseServerError
from allauth.account.models import EmailAddress
from django.contrib import messages

    def _action_send(self, request, *args, **kwargs):
        email = request.POST["email"]
        if EmailAddress.objects.filter(email=email, verified=True).exists():
            messages.error(self.request, "Email has already been verified")
            return HttpResponseRedirect(self.request.path_info)
            #return HttpResponseServerError("Email address has already been verified.")
```

pennersr:
Handled in 880d963d

pennersr:
https://codeberg.org/allauth/django-allauth/issues/754

PR #2770, opened 2021-01-27
fix(accounts) Only re-send email to unverified users

9mido:
Attempts to solve #754

A user is able to click on the re-send verification in /accounts/email/ which will re-send another verification email again to the already verified email address. This will check to see if that email address has already been verified and prevent the re-send verification email from sending out again. 

I decided to use "A link to activate your account has been emailed to the address provided" as the error message as opposed to "This e-mail has already been verified" to stay within OWASP guidelines - see the email privacy leak issue (if we decide on using this OWASP error messaging throughout allauth). 

But I suppose changing the error message to something else would be okay. 

I do not see a need for a setting because there is no reason for verified emails to receive email confirmations again. 

Update: Added a setting to satisfy the possibility of people wanting to still re-send to verified emails. 

### DIFF
--- a/allauth/account/app_settings.py
+++ b/allauth/account/app_settings.py
@@ -339,6 +339,9 @@ class AppSettings(object):
                 ret = []
         return ret
 
+    @property
+    def EMAIL_RESEND_VERIFICATION_UNVERIFIED(self):
+        return self._setting("EMAIL_RESEND_VERIFICATION_UNVERIFIED", True)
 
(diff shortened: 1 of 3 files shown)


arnav13081994:
I think there are mulitple use-cases to re-verify already verified emails. Please check this out https://github.com/pennersr/django-allauth/issues/2857

pennersr:
I believe this is handled in 880d963d

PR #2627, opened 2020-08-07
feat(login): deny unverified EmailAddress from login

iarp:
Allow admin to set a config option that disables a users ability to login via an unverified email address.

Only applies when there are multiple EmailAddress entries, single entries are handled by login system checking for verified emails when ``ACCOUNT_EMAIL_VERIFICATION = 'mandatory'``.

In relation to #650

## General

 - [x] Make sure you use [semantic commit messages](https://seesparkbox.com/foundry/semantic_commit_messages).
       Examples: `"fix(google): Fixed foobar bug"`, `"feat(accounts): Added foobar feature"`.
 - [x] All Python code must be 100% pep8 and isort clean.
 - [ ] JavaScript code should adhere to [StandardJS](https://standardjs.com).
 - [x] If your changes are significant, please update `ChangeLog.rst`.
 - [ ] Feel free to add yourself to `AUTHORS`.

Issue #2360, opened 2019-09-19
Send password reset to verified emails


vsemionov:
Hi @pennersr  and other authors,

Thank you very much for this wonderful project!

I see that `ResetPasswordForm` may send password reset links to any registered emails, not just verified ones. A user could enter an incorrect email address during registration. If `ACCOUNT_EMAIL_VERIFICATION` is not `mandatory`, the user may keep using their account for a while. Later, someone could enter the same same email in the password reset form. It could either be the registered user, whose browser autocompleted the stored incorrect email. But it could also be a malicious person who owns the entered email. In both cases, the owner of the email will now have access to another person's account.

What is your opinion about this scenario? Is it a missed case, or is the behavior of `allauth` intentional?

I also think that even if password reset emails are sent to verified emails, that would still not be completely secure. Currently, anyone with the email confirmation link can verify the email and send the password reset link to themself (if they open the link before it expires). To prevent this, the email verification should be made more "real" (by guaranteeing the person opening the link is the account owner). For example, `ConfirmEmailView` could require that the owner of the email is logged in before confirming the email.

What do you think?

Best regards,
Victor

pennersr:
The email address is the key to the account. If a user adds somebody else's email to his account, he is indeed handing over his account key to that user. I do not consider that a security issue, but rather an error on the side of the user.

Your second part I do not fully comprehend. You mention "anyone with the email confirmation link". That assumes that malicious users have access to the user's mailbox. If that is the case, whatever you would change in the confirm email view is not going to help, as the malicious user can always kick off the password reset flow, intercept the mail, and change the password.

vsemionov:
Thank you for responding.

Indeed, adding someone else's email address to your account is an error (unless done deliberately). However, I am trying to protect users from as many mistakes as possible.

By "anyone with the email confirmation link", I meant someone whose address was added (by mistake). Let me explain my concern with an example. A new user (person A) creates an account and makes a typo in their email address. By coincidence, the mistyped address actually exists, and belongs to person B (whose name is similar). Person B could verify the email address by following the link in the email. Person A does not notice that the email is now verified, despite that they didn't receive a confirmation email). All parts of the website, that require a verified email, are now accessible, so person A engages further and enters sensitive information. Person B now resets the password (by requesting a password reset email to their own address) and locks person A out. He gains full control of the account with all its data.

[...]

Issue #2103, opened 2018-09-04
User not able to sign up if someone else tried to add their email to an account


treyhunner:
On my website, someone attempted to add one of my email addresses to their account. I don't know if this was by mistake or if they were seeing whether they could somehow gain access to my account this way.

This was an email address that I had not yet associated with my account, but that I'd like to associate with my account.

Unless an admin manually logs in and deletes the unverified email account from the account that attempted to add it, that email address cannot be added to another account and cannot be used to register.

I realize that this bug may be seen as something to resolve outside the scope of django-allauth. I do think the documentation should probably address the recommended path for resolving this situation gracefully though.

Thoughts on whether this could be fixed within allauth itself and what I should do in the meantime?

codescribblr:
On a related note, if the user that added your email address to their account has another verified email address, they can actually login using _your_ email address (unverified in their account) and their password. It still logs them into their own account, but this seems problematic at best.

9mido:
Linking #2699 and #2360 and the solutions to this issue contained in both issue comment sections. The solutions in the comments needs PRs with configurable settings. 

Related issue: #2627

pennersr:
> Unless an admin manually logs in and deletes the unverified email account from the account that attempted to add it, that email address cannot be added to another account and cannot be used to register.

This is not true. If someone adds your email to their account, by doing so, that person gives you full access to their account. Meaning, you as a user can use the password reset functionality to take over their account and remove the email as you see fit. It does not at all require an admin to fix things. There is really not much you can do here, the email address is the key to the account, and if somebody adds your key to their account you will have access. In this scenario, a user is either sloppy with security, or perhaps intentionally wants to give you ownership of an account, but all in all this works as intended.

pennersr:
https://codeberg.org/allauth/django-allauth/issues/2103

Issue #1317, opened 2016-03-05
Reset password allowed for inactive users


JulieGoldberg:
A user whose account has been deactivated can use the reset password form to log in.  They have to update their password every time, but they can get in that way.

Presumably filter_users_by_email() should exclude inactive users.  Or it might need to be an additional filter in ResetPasswordForm.clean_email() depending on how widespread filter_users_by_email() is used.  I'm on an older version, but the current version doesn't fix it.

akarambir:
+1 on issue. We had to patch an api where we were using all-auth.

peckjon:
+1 as well... this is a not-uncommon bug in auth systems, and one that definitely needs fixing!

JulieGoldberg:
I don't think the perform login always checks if the user is active.  Maybe
it depends on the django version?  I can and do impersonate inactive users.

However that works, the check comes too late.

I don't believe that an inactive user should be able to request the
password reset link in the first place.

Sent from my iPhone

On Mar 10, 2016, at 1:50 PM, Raymond Penners <notifications@github.com>
wrote:

Could you elaborate a bit. If I am not mistaken, the password reset comes
in 2 flavors. One, where after completing the password reset the user has
to follow the regular login procedure. In that case, the perform_login
method will check user.is_active and not allow the user to log in. The
other case is where login-on-password-reset is turned on, but even in that
case the code path follows perform_login which checks user.is_active. Could
you let me know what scenario you are following to reproduce the issue?

—
Reply to this email directly or view it on GitHub
<https://github.com/pennersr/django-allauth/issues/1317#issuecomment-195066189>
.

pennersr:
allauth always routes through its own `perform_login`:

https://github.com/pennersr/django-allauth/blob/master/allauth/account/utils.py#L130

If you are impersonating users.. you are probably doing some other user session magic beyond allauth scope?

JulieGoldberg:
We use django-impersonate package for impersonating users.  It was just an example.

My point still stands that inactive users shouldn't even be able to get a lost password link in the first place.

pennersr:
That is just polish, and even debatable as it leaks even more information about a user (see #1307) ...  but I was a bit surprised to see the +1s when there is no issue in allauth:

- "...can use the reset password form to log in. They have to update their password every time, but they can get in that way." -- not true, at least, not true as long as allauth is in control. From a django-impersonate point of view, you can even argue that it is a feature to be able to impersonate inactive users.

[...]

PR #2771, opened 2021-01-27
feat(accounts) Password Reset Verified and/or Active Emails Only

9mido:
Attempts to solve #2360

Related: #2103 and #2627 and #1267

After extensive digging I found #1317 which was hard to find since it was closed. You can use the same methodology for preventing inactive accounts from receiving password reset emails. 

These raised form validation errors should also follow OWASP guidelines for password resets:


pennersr:
> Attempts to solve https://github.com/pennersr/django-allauth/issues/2360

#2360 was closed -- see https://github.com/pennersr/django-allauth/issues/2360#issuecomment-997276955

PR #2773, opened 2021-01-27
feat(account) Prevent Privacy Leaking

9mido:
Attempts to solve #1307

Changed multiple error messages for password reset, signup, and adding email pages to make it harder for a user to see proof that another user's credentials exist. 

Follows OWASP recommendations:

### DIFF
--- a/allauth/account/forms.py
+++ b/allauth/account/forms.py
@@ -458,7 +458,7 @@ class AddEmailForm(UserForm):
                 "This e-mail address is already associated" " with this account."
             ),
             "different_account": _(
-                "This e-mail address is already associated" " with another account."
+                "A link to activate your account has been emailed to the address provided."
             ),
             "max_email_addresses": _("You cannot add more than %d e-mail addresses."),
         }
--- a/allauth/account/forms.py
+++ b/allauth/account/forms.py
@@ -19,7 +19,8 @@ from . import app_settings
 from .adapter import get_adapter
 from .app_settings import AuthenticationMethod
 from .models import EmailAddress
-from django.contrib.auth.models import User
+from django.contrib.auth import get_user_model
+User = get_user_model()
 from .utils import (
     filter_users_by_email,
     get_user_model,
--- a/allauth/account/forms.py
+++ b/allauth/account/forms.py
@@ -19,6 +19,7 @@ from . import app_settings
 from .adapter import get_adapter
 from .app_settings import AuthenticationMethod
 from .models import EmailAddress
+from django.contrib.auth.models import User
 from .utils import (
     filter_users_by_email,
     get_user_model,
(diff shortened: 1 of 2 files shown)


allauth/account/forms.py
```diff
@@ -335,13 +336,23 @@ def __init__(self, *args, **kwargs):
         )
 
     def clean_username(self):
-        value = self.cleaned_data["username"]
-        value = get_adapter().clean_username(value)
+        if self.cleaned_data["username"]:
+            value = self.cleaned_data["username"]
+        else:
+            value = get_adapter().clean_username(value)
```
pennersr, line 342: This looks wrong -- `value` is passed to `clean_username()` without being assigned a value first.

allauth/account/forms.py
```diff
@@ -335,13 +336,23 @@ def __init__(self, *args, **kwargs):
...
+        users = User.objects.filter(username__iexact=value)
+        if users:
+            raise forms.ValidationError("A link to activate your account has been emailed to the address provided")
```
pennersr, line 345: Currently, the user would see an error "username taken". So, that indeed leaks information. Now, you would see a different error message if the username is taken -- how does this leak less information? One can come to the same conclusion that a username is taken. 

Also, the error message is simply not stating what truly happens. Suppose I signed up with a username before, and forgot about it. Then, 1 month later, I signup again using the same username. I am presented that error message, and will spend looking at my mailbox for that mail to come in, but no mail will arrive as this is only changing the message itself?

allauth/account/forms.py
9mido: That makes sense. Except I am not really sure what to do here. 

If left the way it was originally like this:
```
        if self.cleaned_data["username"]:
            value = self.cleaned_data["username"]
            value = get_adapter().clean_username(value)
```
the "A link to activate your account has been emailed to the address provided" would not execute and the default "User with this username already exists." will be used instead. 

That is because I am not using an adapter. The value variable gets passed into the query `User.objects.filter(username__iexact=value)` so there needs to be a way to determine if an adapter is being used versus if an adapter is not being used.

allauth/account/forms.py
9mido: Good point an error message about an email link being sent does not make sense for a username. Especially if the email verification setting is None. Or email address is not used for registration at all. 

The point of these error messages is to make it less obvious if the username/email was taken or not. In the case of emails not being leaked it does not explicitly say that the email is taken and makes someone think even if the email address does or does not exist, an email will "appear" to be sent for activation.

Maybe we could just change the duplicate error message for usernames to be "username is invalid" even if there is a duplicate username already being used. It will force the user to use a different username and will not explicitly tell if the username typed in already exists to help prevent brute force attacks. 

Your scenario makes sense. I feel as though that would be a rare occurrence though. This whole thing falls into harvesting and enumeration categories of security. We really need a middle-ground between good error messaging and security without there hopefully being a complete trade off for one or the other. 

It would be cool to be able to allow users to customize these forms.py validation errors without them being hard coded. But not sure how to do that myself.  

You could also change the wording for emails perhaps to be "A link to activate your account may or may not be emailed to the address provided" so that it satisfies your scenario for emails perhaps?

pennersr:
Closing for above mentioned reasons. The way to move forward to prevent leaking is to always send an email for signup. Meaning, it can only work if signup is mandatory. Then, when signing up with an already taken email you can email that fact to the user -- "you're receiving an email to signup but you already did sign up before. Use password reset to recover your account" or something like this. As for username, it's debatable whether usernames are an issue,  any user concerned about privacy is likely not using "firstname.lastname" like usernames anyway.

Issue #1958, opened 2018-02-22
Change email doesn't give feedback when trying to make unverified email primary


invertium:
https://github.com/pennersr/django-allauth/blob/master/allauth/templates/account/email.html

When trying to make an unverified email address primary, the page reloads. There should be a way to tell the user he has to verify it first.

derek-adair:
I'm a bit confused here.  This is potentially a valid feature but it's not clear what the user expects to happen vs what is actually happening.

The user needs to provide;

1. Their allauth version / relevant configuration.  Specifically if verification is required.
2. Their expected behavior.

I _THINK_ they are saying that when VERIFICATION is required and the user hasn't verified there is no error message.  I will look into this one when I have a moment next week sometime.

derek-adair:
One of the things you should do to help coach users to submit valid issues is to set up some [issue templates](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository)

pennersr:
Closing, dated. You do get a proper "Your` primary email address must be verified." message in this case.

pennersr:
https://codeberg.org/allauth/django-allauth/issues/1958

Issue #290, opened 2013-06-09
Can't delete primary email


pjreddie:
Even if an email is not required, you can't delete a primary email account and you can't make the email not primary unless you have one. So if you happen to click make primary once, you are forever stuck with some email (even if it isn't real) instead of having the option to delete all email addresses. It just seems like a strange design choice.

numpde:
I second the "strange design choice" opinion.  The allauth framework is imposing too many constraints here.  I want my visitors to be able to remove all e-mail if they want.  With notifications and social login it is not necessary to have one and feels authoritarian.  Thank you.

9mido:
Views.py EmailView class _action_remove function:

```
            unverified_primary_email = EmailAddress.objects.filter(user=request.user,email=email, primary=True, verified=False).values_list('email', flat=True)
            if email_address.primary and email not in unverified_primary_email:
                get_adapter(request).add_message(
                    request,
                    messages.ERROR,
                    'account/messages/'
                    'cannot_delete_primary_email.txt',
                    {"email": email})
            elif email in unverified_primary_email:
                email_address.delete()
            else:
                email_address.delete()
```

derek-adair:
Close as per discussion in https://github.com/pennersr/django-allauth/pull/2774 (i think??).  It seems like this is a corner case that can be solved w/ a custom adapter.

pennersr:
No, this is a valid issue. If `ACCOUNT_EMAIL_REQUIRED=False` and `ACCOUNT_AUTHENTICATION_METHOD="username"` then indeed you should be able to delete all email addresses from your account.

derek-adair:
Right but is that not solved by a custom adapter? Or did i misread your comment to mean it's worth modifying the core adapter to support this?

pennersr:
It is an issue that can be worked around with by customizing things yourself, but this should indeed just work out of the box.

pennersr:
Closed via a0ff1aa2

pennersr:
https://codeberg.org/allauth/django-allauth/issues/290

PR #2774, opened 2021-01-27
fix(accounts) Control Removal Of Emails

9mido:
Attempts to solve #290 and #1958

I made it so that users could control whether they want to delete emails based on if they are verified, primary, or any combination of those.

### DIFF
--- a/allauth/account/app_settings.py
+++ b/allauth/account/app_settings.py
@@ -340,8 +340,20 @@ class AppSettings(object):
         return ret
 
     @property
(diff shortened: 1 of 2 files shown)


pennersr:
We really need to be careful with introducing new settings. When introducing a new setting, there should be a really good use case for flipping it one way or the other. What is the purpose of e.g. `DELETE_UNVERIFIED_UNPRIMARY_EMAIL` -- when would you want to disallow this? If we need such fine grained control, I think it would make more sense to put a `can_delete_email` method on the `adapter`. Therefore, I am closing this.

PR #2778, opened 2021-02-01
feat(accounts) Encrypt Email Address

9mido:
https://gdpr.eu/email-encryption/

This depends on the pgcrypto extension to be installed in Postgres and it only works if Postgres is the database.

```
psql -d database_name -c 'CREATE EXTENSION pgcrypto'
CREATE EXTENSION pgcrypto;
```

It uses a third party library to do the encryption/decryption with a symmetric key users set in settings.py:

https://github.com/incuna/django-pgcrypto-fields

Settings.py:

```
PGCRYPTO_KEY = 'KEY_VALUE'
INSTALLED_APPS = [ 'pgcrypto',]
```

If users wanted to encrypt the username, Allauth has a user = models.ForeignKey in models.py so you would probably need to encrypt usernames using the custom user model. What this PR does is it encrypts the account_emailaddress email column so that whenever Allauth updates in the future, the django developer does not need to override Allauth each time just to add encryption to the email address inside the account_emailaddress table. 

Warning - a user would need to do a data migration if they decide to switch from decrypted email to encrypted email or vice versa. It is better in the beginning of a new project to choose whether you want the email to be encrypted/decrypted and never change it later on.

https://docs.djangoproject.com/en/3.2/howto/writing-migrations/

### DIFF
--- a/allauth/account/app_settings.py
+++ b/allauth/account/app_settings.py
@@ -339,6 +339,9 @@ class AppSettings(object):
                 ret = []
         return ret
 
+    @property
+    def EMAIL_ENCRYPTED(self):
+        return  self._setting('EMAIL_ENCRYPTED', True)
 
 # Ugly? Guido recommends this himself ...
 # http://mail.python.org/pipermail/python-ideas/2012-May/014969.html
--- a/allauth/account/models.py
+++ b/allauth/account/models.py
@@ -14,18 +14,22 @@ from .adapter import get_adapter
 from .managers import EmailAddressManager, EmailConfirmationManager
 from .utils import user_email
 
+from django.conf import settings
+if settings.PGCRYPTO_KEY:
+   from pgcrypto import fields
+
 
 class EmailAddress(models.Model):
+
+    if app_settings.EMAIL_ENCRYPTED is True:
+        email = fields.EmailPGPSymmetricKeyField(unique=app_settings.UNIQUE_EMAIL,max_length=app_settings.EMAIL_MAX_LENGTH,verbose_name=_('e-mail address'))
+    else:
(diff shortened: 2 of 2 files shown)


pennersr:
That GDPR link is about email encryption, not encryption of the email address. Note that even if you encrypt the email address, your organisation still has access to the email address as it can be freely decrypted, so GDPR wise encrypting the email address does not solve anything.  Additionally, I don't want to pick up `django-pgcrypto-fields` as a dependency just yet.  If anybody wants to do stuff like this, I think a feature of making the allauth models swappable makes more sense, then people are free to do whathever they want. All in all, I prefer to close this one...

9mido:
@pennersr Yes that is GDPR but other links that I did not include are PII, CCPA, LGPD, HIPPA. 

This whole topic is debatable. All companies I have worked for seem to encrypt their PII data just to be compliant / lawful even though to many people it seems like a waste of time. If anything, it will make things a bit harder for an attacker and auditors won't keep bothering you to be compliant. If there is a data breach, the attacker cannot do anything with the database or see the plaintext emails if they do not have the secret key from your django project. Only if both your django project and database are compromised then you are screwed. There are other ways besides this to be secure like encryption in transit or full disk encryption of your server, linux permissions, etc. 

I guess for now I won't be using allauth and I will need to use the built in django authentication framework to do this until swappable models become a feature of allauth. 

We don't have to use django-pgcrypto-fields, we could use something like django-cryptography where only django querysets can decrypt the data, and SQL queries cannot and appear encrypted always. But what would be even better is to have something like homomorphic encryption to prepare us for quantum computing. 

https://www.youtube.com/watch?v=iKL1pGy-r_M

https://security.stackexchange.com/questions/224663/storing-email-addresses-in-plain-text-is-okay

https://stackoverflow.com/questions/767276/what-is-the-best-and-safest-way-to-store-user-email-addresses-in-the-database

PR #2779, opened 2021-02-03
fix(accounts) 404 /accounts/email/ setting

9mido:
There is no reason to have /accounts/email/ accessible when ACCOUNT_MAX_EMAIL_ADDRESSES == 1. If ACCOUNT_MAX_EMAIL_ADDRESSES > 1, then provide access to /accounts/email/.

Update: Changed this to add a new setting that allows the user to either enable or disable this without worrying about ACCOUNT_MAX_EMAIL_ADDRESSES.

### DIFF
--- a/allauth/account/app_settings.py
+++ b/allauth/account/app_settings.py
@@ -316,6 +316,10 @@ class AppSettings(object):
         return self._setting("PRESERVE_USERNAME_CASING", True)
 
     @property
+    def EMAIL_PAGE_404(self):
+        return self._setting("EMAIL_PAGE_404", False)
+    
+    @property
     def USERNAME_VALIDATORS(self):
         from django.core.exceptions import ImproperlyConfigured
 
--- a/allauth/account/urls.py
+++ b/allauth/account/urls.py
@@ -2,6 +2,8 @@ from django.urls import path, re_path
 
 from . import views
 
+from django.conf import settings
+from django.views.defaults import page_not_found
 
 urlpatterns = [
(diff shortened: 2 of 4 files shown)


ogurec-ogurec:
> There is no reason to have /accounts/email/ accessible when ACCOUNT_MAX_EMAIL_ADDRESSES = 1. If ACCOUNT_MAX_EMAIL_ADDRESSES > 1, then provide access to /accounts/email/

If we do not have a verified email, how can we call a repeated email if the path to the page where the "Re-send Verification" button is located is not available?)

9mido:
@ogurec-ogurec this sounds similar to the other question you asked in the timeouts PR. When an unverified user tries to login, a verification email will be sent to them for every login attempt until they verify. But this depends on verification setting to be mandatory. When the verification setting is None, no email verifications will be sent at all. For optional verifications, verification emails will only be sent on signups. I guess to satisfy that, we could have the urls.py code in this PR to be -> if settings.ACCOUNT_MAX_EMAIL_ADDRESSES == 1 and settings.ACCOUNT_EMAIL_VERIFICATION = 'mandatory'. That way, optional settings can still get email verifications sent and not have /accounts/email/ blocked.

Or you could have a different setting like ACCOUNT_EMAIL_PAGE_404 = True|False and use that in the urls.py also instead of just doing it on ACCOUNT_MAX_EMAIL_ADDRESSES == 1.

ogurec-ogurec:
@9mido Thanks!

pennersr:
Closing -- we do need it, see #3346

Issue #2726, opened 2020-11-26
Does Django-allauth have a built in way of deleting accounts?


nh916:
As I have been using django-allauth in more projects I have noticed over and over that deleting accounts is missing and I thought it would complete the solution. I have used Django allauth for years and love it and would love to see it included

9mido:
https://github.com/pennersr/django-allauth/issues/344

nh916:
Oh yeah I saw that one but I thought it would be way better if it was built in. Allauth does a great job with every part of the accounts process so I thought this is the last thing that could really complete their whole framework from registration to deletion

9mido:
I agree to some extent since GDPR basically has "the right to erasure" built into it. 

Although I see pennersr's point about different implementations of deletion depending on the type of website. If this were something like Facebook picture uploads I would also want to delete the pictures along with the account with models.CASCADE and not have the people who shared the pictures access them anymore also and have them 404 or hidden or something. Maybe the comments and URL exist but the picture / comments will not. Like if this were a child's provocative pictures those need to be deleted also along with everything else the child posted.  

But for other websites without media files associated with the account or even just saved profile text based information for that user, it should be part of allauth since deleting everything is now part of many apps nowadays compared to 7 years ago when the issue was first created.

pennersr:
See: https://github.com/pennersr/django-allauth/pull/2786#issuecomment-977960052

pennersr:
https://codeberg.org/allauth/django-allauth/issues/2726

Issue #344, opened 2013-07-30
Delete Account


solidstate-tech:
How would I go about creating a view where the user can delete their account ?

9mido:
If you want to delete an account do this:

View:

```
from django.contrib.auth import logout
from django.contrib.auth.models import User
from .forms import DeleteUser

def delete_user_view(request):
    if request.user.is_authenticated:
        if request.method == 'POST':
            form = DeleteUser(request.POST)

            if form.is_valid():
                if request.POST["delete_checkbox"]:
                    rem = User.objects.get(username=request.user)
                    if rem is not None:
                        rem.delete()
                        logout(request)
                        messages.info(request, "Your account has been deleted.")
                        return redirect(reverse("app:url_name"))
                    else:
                        messages.info(request, "There was an error.")
        else:
            form = DeleteUser()
        context = {'form': form}
        return render(request, 'file.html', context)
        if request.user.is_anonymous:
            return HttpResponse(render(request, "401.html"), status=401)
```

Form:

```
class DeleteUser(forms.Form):
    delete_checkbox = forms.BooleanField(label=mark_safe('Are you sure you want to delete your account?'), required=True)
    def __init__(self, *args, **kwargs):
        super(DeleteUser, self).__init__(*args, **kwargs)

```

Urls:

```
from .views import(delete_user_view,)
path('delete/', delete_user_view, name='delete_user_view'),
```

HTML:

```
<form id='delete' action="{% url 'app:delete_user_view' %}" method="post">
	    {% csrf_token %}
	    {{ form.as_p }}
    <button id="delete-button" type="submit">Submit Deletion</button>
</form>

```

pennersr:
https://codeberg.org/allauth/django-allauth/issues/344

PR #2785, opened 2021-02-16
feat(accounts) - Delete account

9mido:
Attempts to solve #2726, #344

Not sure how to do the adapter part where the django developer could override the forms and views for this. But otherwise, this is a minimal viable product & built in allauth way to allow users to delete their account. This code does delete the user and the developer could specify where to redirect after the account deletion with ACCOUNT_DELETE_REDIRECT_URL and whether they want to enable/disable account deletion with ACCOUNT_DELETE. 

The django developer should specify in their models with on_delete whether they want to delete things like media files from each account. The django developer should be able to customize the form and views from this new code so that they could do things like surveys or store feedback in the database to satisfy project specific needs. Otherwise something default like a checkbox for consent should be fine.

### DIFF
--- a/allauth/account/app_settings.py
+++ b/allauth/account/app_settings.py
@@ -318,6 +318,10 @@ class AppSettings(object):
     @property
     def DELETE_REDIRECT_URL(self):
         return  self._setting('DELETE_REDIRECT_URL', '/accounts/login/')
+    
+    @property
+    def DELETE(self):
+        return  self._setting('DELETE', False)
 
     @property
     def USERNAME_VALIDATORS(self):
--- a/allauth/account/app_settings.py
+++ b/allauth/account/app_settings.py
@@ -316,6 +316,10 @@ class AppSettings(object):
         return self._setting("PRESERVE_USERNAME_CASING", True)
 
     @property
+    def DELETE_REDIRECT_URL(self):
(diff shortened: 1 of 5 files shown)


ogurec-ogurec:
@9mido thanks, great job! Please tell me how you can issue a warning - that the page will be deleted permanently, for example (like when logging out)?

9mido:
@ogurec-ogurec Thanks. There are a couple of ways. You can override the new allauth delete.html template and add a <p> tag somewhere with your warning. The other way is to override the forms.py BooleanField label. And the page is not necessarily deleted permanently it is the entire account and possibly other data associated with that account will be deleted (but this depends on your app's models on_delete values if you want to delete everything or not). When the form submits after the user consents to having their account deleted, the user will be logged out automatically and their account will be deleted with or without extra data associated with it. Once deleted, the user that was deleted could always come back and re-create their account in the future.

pennersr:
Closing for reasons mentioned here: https://github.com/pennersr/django-allauth/pull/2786#issuecomment-977960052

aehlke:
I disagree with the logic, if I may: yes deletion requests can be improved, however it's now a regulatory requirement across the world enforced by various agencies as well as by Facebook themselves and other vendors, I think it's more important to satisfy the compliance requirement as a baseline and allow devs to extend or replace the behavior's business logic

PR #2786, opened 2021-02-16
feat(accounts) - Deactivate account

9mido:
There does not seem to be an already created issue for this. But lots of apps nowadays allow users to deactivate their accounts themselves and then possibly come back later on. This will allow users to deactivate their accounts but currently this code does not offer a way to allow users to come back / reactivate afterwards unless they contact the website to request it. There should probably be a way to differentiate users who choose to deactivate themselves versus bad acting deactivated users who deserve to be deactivated permanently. 

Not sure how to do the adapter part where the django developer could override the forms and views for this. But otherwise, this is a minimal viable product & built in allauth way to allow users to deactivate their account.

### DIFF
--- a/allauth/account/app_settings.py
+++ b/allauth/account/app_settings.py
@@ -339,6 +339,13 @@ class AppSettings(object):
                 ret = []
         return ret
 
+    @property
+    def DEACTIVATE_REDIRECT_URL(self):
+        return  self._setting('DEACTIVATE_REDIRECT_URL', '/accounts/login/')
+
+    @property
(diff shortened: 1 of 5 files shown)


pennersr:
I don't think it is wise to include deactivation in the scope of allauth. For example, you've opted for immediately deactivating the account, whereas a more elaborate implementation would set a future date on which the account is deactivated, and allow the user to change his mind before user data is permanently destroyed. All in all, these scenarios are best handled elsewhere.

PR #2796, opened 2021-02-27
feat(accounts) - Primary Email Change Enable/Disable Setting

### LINKED ISSUE #2620: Feature request: Restriction of freedom to change primary email addresses
I noticed that changing primary email requires no verification except that the new email has to be verified. This open up a path to vulnerability issues. 
Take a scenario where a user's password has been compromised. A hacker could just login to the user's account, add their email, verify it and then change the primary email to their(the hacker)  own. 
This means in the database of users falls into the hands of hackers, the whole accounts are gone. 

**What should happen instead**
In the event primary email is going to be changed, a confirmation email should be sent to the present primary email for the user to approve this change. 

I was working on a project that needed this package and I noticed this vulnerability. I put together a hack that works for just this purpose. If the maintainers are interested, I could make a PR on a separate branch of this and the code would be improved upon to be stable and thoroughly tested before being made available to users. This PR would include another feature, one which I'll create another issue on later. 

Thanks for the opportunity

9mido:
Partially attempts to solve #2620. This PR will allow django developers to use a setting to allow/deny users to change their primary email address. If True, allow users to change their primary email address to another email address. If False, prevent users from changing their primary email address at all. 

This PR does not include what #2620 suggests as a solution. That feature could always come later on in a future PR.

### DIFF
--- a/allauth/account/app_settings.py
+++ b/allauth/account/app_settings.py
@@ -316,6 +316,10 @@ class AppSettings(object):
         return self._setting("PRESERVE_USERNAME_CASING", True)
 
     @property
+    def PRIMARY_EMAIL_CHANGE(self):
+        return self._setting("PRIMARY_EMAIL_CHANGE", True)
+
+    @property
     def USERNAME_VALIDATORS(self):
         from django.core.exceptions import ImproperlyConfigured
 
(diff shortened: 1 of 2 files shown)


pennersr:
Just closing the door on changing the email address without leaving the user with any other option to proceed as suggested in #2620 does not really result in a valid feature/scenario. Therefore, I am closing this.

Issue #2826, opened 2021-04-06
Django 3.2: multiple warnings


danjac:
Django 3.2 (released today) generates a number of deprecation and other warnings, for example:

`HINT: Configure the DEFAULT_AUTO_FIELD setting or the AccountConfig.default_auto_field attribute to point to a subclass of AutoField, e.g. 'django.db.models.BigAutoField'.
`

`
RemovedInDjango41Warning: 'allauth.account' defines default_app_config = 'allauth.account.apps.AccountConfig'. Django now detects this configuration automatically. You can remove default_app_config.`

djpeacher:
Looks like adding `DEFAULT_AUTO_FIELD = 'django.db.models.BigAutoField'` to `settings.py` is a temp solution.

https://docs.djangoproject.com/en/3.2/releases/3.2/#customizing-type-of-auto-created-primary-keys

Flimm:
The trouble with setting `DEFAULT_AUTO_FIELD` to `'django.db.models.BigAutoField'` in `settings.py` is this: when you run `python manage.py makemigrations`, a new migrations file will be created for allauth. If you're using a virtualenv that is not checked into version control, this is not ideal.

Instead, it is better to override `apps.py` in your application code, as explained in this answer: https://stackoverflow.com/a/67007098/247696

I recommend that `django-allauth` modify `apps.py` to avoid this warning. I'll be submitting a pull request soon.

9mido:
Great job @Flimm that was a perfect explanation and thanks for the PR.

pennersr:
https://codeberg.org/allauth/django-allauth/issues/2826

PR #2829, opened 2021-04-08
Set default_auto_field to 'AutoField'

Flimm:
Django 3.2 introduces DEFAULT_AUTO_FIELD . If you set this to something other than `'django.db.models.AutoField'`, migrations will need to be generated for allauth's models. This pull request sets `default_auto_field` just for these apps so that no migrations need to be generated.

It also hides warnings on Django 3.2.

Fixes: https://github.com/pennersr/django-allauth/issues/2826

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -16,6 +16,11 @@ Backwards incompatible changes
   been renamed to ``certificate_key``. This is done to prevent the key from being displayed
   without being masked in Django debug pages.
 
+0.44.0
+******
+
+- Better compatibility with Django 3.2
+
 
 0.43.0 (2020-10-15)
 *******************
--- a/allauth/account/apps.py
+++ b/allauth/account/apps.py
@@ -5,3 +5,4 @@ from django.utils.translation import gettext_lazy as _
 class AccountConfig(AppConfig):
     name = "allauth.account"
     verbose_name = _("Accounts")
+    default_auto_field = 'django.db.models.AutoField'
--- a/allauth/socialaccount/apps.py
+++ b/allauth/socialaccount/apps.py
@@ -5,3 +5,4 @@ from django.utils.translation import gettext_lazy as _
 class SocialAccountConfig(AppConfig):
     name = "allauth.socialaccount"
     verbose_name = _("Social Accounts")
+    default_auto_field = 'django.db.models.AutoField'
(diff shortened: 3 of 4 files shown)


ObserverOfTime:
When will this be merged?

czue:
What's the status of this issue? Any plans for a merge/release anytime soon?

adamchainz:
It would be better to migrate all the models to `BigAutoField` with explicit declarations. This would prevent them from exhausting ID's, which is the reason Django added `DEFAULT_AUTO_FIELD` in the first place. But that might be a major version bump.

czue:
Yeah agreed that makes sense in the long term. In the short term the creation of these migrations can be problematic for new projects, which can easily introduce migrations that depend on files that only exist in a virtualenv on a single dev's machine, and then lead to confusing "missing migration" error when trying to deploy on another environment. 

If the release process isn't too painful, do you think it might make sense to roll the fixes out in two stages to avoid people getting tripped up in the interim?

adamchainz:
Two fixes would be a good idea, yes. (I am not a maintainer, just a user.)

simkimsia:
While waiting for this to be merged, what's a regular user to do to at least temporary suppress the warning without triggering a migration file at least?

Flimm:
> While waiting for this to be merged, what's a regular user to do to at least temporary suppress the warning without triggering a migration file at least?

[My post on Stack Overflow](https://stackoverflow.com/a/67007098/247696) explains a workaround.

simkimsia:
@Flimm 

Thank you. HOwever, when I followed your workaround including the one for bug #2853, i get the following

```
ERRORS:
django_1    | account.EmailAddress.user: (fields.E301) Field defines a relation with the model 'auth.User', which has been swapped out.
django_1    | 	HINT: Update the relation to point at 'settings.AUTH_USER_MODEL'.
django_1    | socialaccount.SocialAccount.user: (fields.E301) Field defines a relation with the model 'auth.User', which has been swapped out.
django_1    | 	HINT: Update the relation to point at 'settings.AUTH_USER_MODEL'
```

Suggestions?

brianhelba:
@pennersr This is an important compatibility fix for Django 3.2, and it's generated significant user attention. Could you please review this?

After this is merged, could we also get a PyPI release?

Flimm:
Yay! Thank you for merging.

brianhelba:
@pennersr Thanks! Could you issue a new PyPI release to include this change?

Reston:
I had this problem!

Waiting for this to be released :smile: 

@pennersr any update?

Flimm:
The latest release on PyPI does contain the fix, now.

Issue #2862, opened 2021-05-20
Bug or feature?: Auto login after registration does not while using a different browser


sowinski:
Hi,

if I do the registration process with the same browser, everything works fine:
1) sign up
2) click on link in email
3) click on accept email in the website
4) => user is logged in#

but if I click on the link on the email and the page is opened with a different browser then I see the login page after step 3).

Not sure if this is a bug or security related feature?

If this is not a bug. How can I change this behavior?

sowinski:
I have set:
ACCOUNT_LOGIN_ON_EMAIL_CONFIRMATION = True

pennersr:
This is by design, not a bug. Explained here: https://github.com/pennersr/django-allauth/blob/master/allauth/account/views.py#L337

pennersr:
https://codeberg.org/allauth/django-allauth/issues/2862

Issue #1522, opened 2016-10-25
Email Verification Link Invalidation


last-partizan:
If user uses same link again within `ACCOUNT_EMAIL_CONFIRMATION_EXPIRE_DAYS` it should not be working.

Possible fix for `EmailConfirmationHMAC`: use not only EmailAddress.pk, but pk and verified flag. So key is invalidated if email confirmed.

9mido:
Similar to #2731

gabn88:
Subscribing to this topic. 
In my flow, blacklisting email verification links is mandatory. This is because admin users in an organisation can create new users for their organisation (with an emailaddress). But they can also change the email-address afterwards if the user hasn't logged in yet (for example if they had a typo in the mailaddress). But let's say the emailaddress with the typo really exists, that user can use the email confirmation link to create a password for the user account and change the emailaddress to something they want.

9mido:
For the signup confirmation links to 'appear' be expired immediately after the user confirms their email address, you can do this:

https://github.com/pennersr/django-allauth/blob/master/allauth/account/views.py:

```
class ConfirmEmailView(TemplateResponseMixin, LogoutFunctionalityMixin, View):

    template_name = "account/email_confirm." + app_settings.TEMPLATE_EXTENSION

    def get(self, *args, **kwargs):
        try:
            self.object = self.get_object()
            if app_settings.CONFIRM_EMAIL_ON_GET:
                return self.post(*args, **kwargs)
            email = self.object.email_address.email
            if EmailAddress.objects.filter(email=email, verified=True).exists():
                raise Http404
        except Http404:
            self.object = None
        ctx = self.get_context_data()
        return self.render_to_response(ctx)
```

I say 'appear' because they are not truly expired until the ACCOUNT_EMAIL_CONFIRMATION_EXPIRE_DAYS runs out. It would be better to just run a piece of code that forces the expiration of the links as if ACCOUNT_EMAIL_CONFIRMATION_EXPIRE_DAYS runs out. In other words, figure out how to make a link that expires immediately after clicking it inside of the email provider inbox. If anyone here can figure that out, please share. Maybe that can be a separate allauth setting. But then again, if a user accidentally closes the one time link, they cannot access that link again which may be a bad idea. So I guess this approach is a happy middle ground. 

[...]

PR #2865, opened 2021-05-24
feat(accounts) Only show /accounts/confirm-email/ link confirmation process to unverified email addresses

9mido:
Attempts to solve: 

https://github.com/pennersr/django-allauth/issues/1522

https://github.com/pennersr/django-allauth/issues/2862

Currently when a user signs up, they get an email with a link where they are redirected to /accounts/confirm-email/ to confirm their account is legitimate. After clicking their /accounts/confirm-email/ link confirm button, they can re-use that same /accounts/confirm-email/ link again in the same browser or a different browser and go through the process of confirming again. 

With this new setting turned set to True, a django developer could decide to not allow verified email addresses to confirm themselves again with the same link after already verifying themselves the first time. It will instead show the message "This e-mail confirmation link expired or is invalid" the second time around. Otherwise, with this setting set to False, the email address could be verified again with the same link after they have already verified themselves and need to wait until ACCOUNT_EMAIL_CONFIRMATION_EXPIRE_DAYS runs out to no longer confirm themselves which is the current allauth behavior.

### DIFF
--- a/allauth/account/app_settings.py
+++ b/allauth/account/app_settings.py
@@ -314,10 +314,6 @@ class AppSettings(object):
     @property
     def PRESERVE_USERNAME_CASING(self):
         return self._setting("PRESERVE_USERNAME_CASING", True)
-
-    @property
-    def EMAIL_CONFIRMATION_UNVERIFIED_ONLY(self):
-        return self._setting("EMAIL_CONFIRMATION_UNVERIFIED_ONLY", True)
     
     @property
     def USERNAME_VALIDATORS(self):
--- a/allauth/account/app_settings.py
+++ b/allauth/account/app_settings.py
@@ -316,6 +316,10 @@ class AppSettings(object):
         return self._setting("PRESERVE_USERNAME_CASING", True)
 
     @property
+    def EMAIL_CONFIRMATION_UNVERIFIED_ONLY(self):
+        return self._setting("EMAIL_CONFIRMATION_UNVERIFIED_ONLY", True)
+    
+    @property
     def USERNAME_VALIDATORS(self):
         from django.core.exceptions import ImproperlyConfigured
 
--- a/allauth/account/app_settings.py
+++ b/allauth/account/app_settings.py
@@ -314,7 +314,7 @@ class AppSettings(object):
     @property
     def PRESERVE_USERNAME_CASING(self):
         return self._setting("PRESERVE_USERNAME_CASING", True)
-    
+
     @property
     def USERNAME_VALIDATORS(self):
         from django.core.exceptions import ImproperlyConfigured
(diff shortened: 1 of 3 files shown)


allauth/account/views.py
```diff
@@ -293,6 +293,9 @@ def get(self, *args, **kwargs):
             self.object = self.get_object()
             if app_settings.CONFIRM_EMAIL_ON_GET:
                 return self.post(*args, **kwargs)
+            email = self.object.email_address.email
+            if EmailAddress.objects.filter(email=email, verified=True).exists() and app_settings.EMAIL_CONFIRMATION_UNVERIFIED_ONLY is True:
```
pennersr, line 297: Doing this check here results in this not working when `CONFIRM_EMAIL_ON_GET` is `True`.

Wouldn't `EmailConfirmationHMAC::from_key()` be a better place? The `EmailAddress` lookup is already there.

allauth/account/app_settings.py
```diff
@@ -315,6 +315,10 @@ def SALT(self):
     def PRESERVE_USERNAME_CASING(self):
         return self._setting("PRESERVE_USERNAME_CASING", True)
 
+    @property
+    def EMAIL_CONFIRMATION_UNVERIFIED_ONLY(self):
+        return self._setting("EMAIL_CONFIRMATION_UNVERIFIED_ONLY", True)
```
pennersr, line 320: Wondering why would this need to be a setting, as having a setting means that there are good reasons to turn this switch either way.  Is there a good reason why `True` should not be the only/default behaviour ?

pennersr (changes requested):

allauth/account/app_settings.py
9mido: You're right. I don't think there needs to be a setting here and it should be True all the time. This setting can be removed then.

allauth/account/views.py
9mido: Ok good catch I am fine with that.

allauth/account/views.py
```diff
@@ -293,6 +293,7 @@ def get(self, *args, **kwargs):
             self.object = self.get_object()
             if app_settings.CONFIRM_EMAIL_ON_GET:
                 return self.post(*args, **kwargs)
+            email = self.object.email_address.email
```
pennersr, line 296: This is not used I think?

allauth/account/app_settings.py
```diff
@@ -314,7 +314,7 @@ def SALT(self):
     @property
     def PRESERVE_USERNAME_CASING(self):
         return self._setting("PRESERVE_USERNAME_CASING", True)
-
+
```
pennersr, line 317: Wrongly formatted...

PR #2872, opened 2021-05-28
feat(accounts) Login with verified emails only

9mido:
Adds a setting to only allow verified emails to login. Unverified emails will get a form validation error.

This is useful because it prevents unverified emails trying to login from sending spam confirmation emails over and over. And it ensures only legitimate people are able to login to your website. 

If this setting is true, the django developer could always run a cronjob to delete unverified emails if the newly signed up user loses their original confirmation email or it expires. Then the user could signup again with the same email address.

### DIFF
--- a/allauth/account/forms.py
+++ b/allauth/account/forms.py
@@ -175,6 +175,9 @@ class LoginForm(forms.Form):
             return
         credentials = self.user_credentials()
         user = get_adapter(self.request).authenticate(self.request, **credentials)
+        email = user.email
+        if EmailAddress.objects.filter(email=email, verified=False).exists() is True and app_settings.LOGIN_VERIFIED_EMAIL_ONLY is True:
+            raise forms.ValidationError('Please check ' + email + ' to verify your email address.')
         if user:
             self.user = user
         else:
--- a/allauth/account/forms.py
+++ b/allauth/account/forms.py
@@ -175,17 +175,20 @@ class LoginForm(forms.Form):
             return
         credentials = self.user_credentials()
         user = get_adapter(self.request).authenticate(self.request, **credentials)
-        email = user.email
-        if EmailAddress.objects.filter(email=email, verified=False).exists() is True and app_settings.LOGIN_VERIFIED_EMAIL_ONLY is True:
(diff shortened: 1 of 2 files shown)


allauth/account/forms.py
```diff
@@ -183,6 +183,9 @@ def clean(self):
                 login = self.cleaned_data["login"]
                 if self._is_login_email(login):
                     auth_method = app_settings.AuthenticationMethod.EMAIL
+                    email = user.email
+                    if EmailAddress.objects.filter(email=email, verified=False).exists() is True and app_settings.LOGIN_VERIFIED_EMAIL_ONLY is True:
```
matmair, line 187: @9mido wouldn't it be more efficient to put the settings check first? Database-access-wise I mean

9mido: You mean change the order of the if statement to be:

`if app_settings.LOGIN_VERIFIED_EMAIL_ONLY is True and EmailAddress.objects.filter(email=email, verified=False).exists() is True:` ? I don't think it really matters much.

matmair: Access to settings does not require a db load so first checking that would reduce dB actions. But it is marginal

9mido: Right. Can you show me the exact code correction you would like to see and how I would change it?

matmair: if app_settings.LOGIN_VERIFIED_EMAIL_ONLY is True and EmailAddress.objects.filter(email=email, verified=False).exists() is True:

pennersr:
You mention two improvements:

> 1) This is useful because it prevents unverified emails trying to login from sending spam confirmation emails over and over. 

But, it essentially blocks people from recovering their account in case they lost the first verification email. Also, I can create signups for a lot of email addresses that I don't own, and all those email addresses would be blocked from signing up.

That is quite a price to pay for addressing the issue of sending too many confirmation emails. 


> And it ensures only legitimate people are able to login to your website.

I don't see why this is listed as an improvement point, as this is already the case (when email verification is mandatory).

If sending too many confirmation emails is the problem, I would rather tackle just that. For example, we could just as well implement an exponential back off for sending those mails and we solve the issue without blocking users or having to resort to cron jobs.

PR #2881, opened 2021-06-07
fix(docs): fixed reference in ACCOUNT_LOGOUT_REDIRECT_URL docs

ankurahuja-tech:
Fixed the reference in `ACCOUNT_LOGOUT_REDIRECT_URL` docs in the `configuration.rst` file which as I believe should refer to default Django `LOGOUT_REDIRECT_URL` instead of `LOGIN_REDIRECT_URL` as its counterpart.


pennersr:
Closed via 341c460d

Issue #2876, opened 2021-05-30
Default AddEmailForm ignores ACCOUNT_EMAIL_VERIFICATION setting and attempts to send email, resulting in OSError at /accounts/email/


lorey:
So, this might be an issue/misunderstanding on my side, but maybe it's not:

I'm running a basic Django setup without much custom configuration/settings, esp. no email set-up so far. I'm using allauth with the default configuration. Since I haven't configured email yet, I can't send verification emails and have thus set `ACCOUNT_EMAIL_VERIFICATION = "none"`. It seems that allauth tries to send an email anyhow.

## Expected behavior
With `ACCOUNT_EMAIL_VERIFICATION = "none"` I would expect no verification attempt as the documentation states:
> When set to “mandatory” the user is blocked from logging in until the email address is verified. Choose “optional” or “none” to allow logins with an unverified e-mail address. In case of “optional”, the e-mail verification mail is still sent, whereas in case of “none” no e-mail verification mails are sent.

## Observed behaviour
When adding an email to my account with the default view, I get the following exception:
```
Environment:


Request Method: POST
Request URL: http://localhost:8008/accounts/email/

Django Version: 3.2.3
Python Version: 3.8.10
Installed Applications:
['django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'django.contrib.sites',
 'crispy_forms',
 'networking_public',
 'networking_base',
 'networking_web',
 'allauth',
 'allauth.account',
 'allauth.socialaccount',
 'allauth.socialaccount.providers.google']
Installed Middleware:
['django.middleware.security.SecurityMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware']



Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.8/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
[...]

PR #2896, opened 2021-07-06
fix(accounts): Do not ignore `ACCOUNT_EMAIL_VERIFICATION` settings on `EmailView`

saadmk11:
# Submitting Pull Requests

## General

Closes https://github.com/pennersr/django-allauth/issues/2876

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -101,6 +101,7 @@ Luiz Guilherme Pais dos Santos
 Luke Burden
 Luke Crouch
 Maksim Rukomoynikov
+Maksudul Haque
 Marcin Skarbek
 Marcin Spoczynski
 Marco Fucci
--- a/allauth/account/forms.py
+++ b/allauth/account/forms.py
@@ -466,8 +466,11 @@ class AddEmailForm(UserForm):
         return value
 
     def save(self, request):
+        confirm = (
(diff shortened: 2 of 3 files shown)


pennersr:
`ACCOUNT_EMAIL_VERIFICATION` is documented as:

>  Determines the e-mail verification method during **signup**

What use case is there for not verifying addtional emails after signup? And if there is a valid use case, this would need a different setting.

lorey:
You're right, the docs state signup. I would assume though that most people that don't want verification during signup also don't want it when adding additional emails either. Why would you want an unverified primary email but verified secondary emails? At least that was the case for me in #2876 as well and what @saadmk11 implemented, too. 

I think the issue is that the name of the setting implies it being a general setting turning all verification off whereas (only) the docs state otherwise. Easiest fix imho would be to make it a general one and let users change the details by overwriting the specific forms if they want to.

TheTerrasque:
> `ACCOUNT_EMAIL_VERIFICATION` is documented as:
> 
> > Determines the e-mail verification method during **signup**
> 
> What use case is there for not verifying addtional emails after signup? And if there is a valid use case, this would need a different setting.

In my case, I don't have email server configured, I don't need nor want confirmation email, the only users are manually added users, and the only reason to add an email is to match users to our authentik server, where all users are added manually and with verified data.

PR #2979, opened 2021-11-17
Redirect to LOGIN_REDIRECT_URL after setting password

confuzeus:
# Set PasswordSetView to redirect to LOGIN_REDIRECT_URL

1. A user who signed up using social account visits the `password_set` view.
2. He enters a desired password and submits the form.
3. The password is set and he's redirected to `password_change` view.

Instead of asking the user to **change his password after he just set it**, we need to redirect to a sensible place — like `LOGIN_REDIRECT_URL`.

### DIFF
--- a/allauth/account/views.py
+++ b/allauth/account/views.py
@@ -13,8 +13,6 @@ from django.views.decorators.debug import sensitive_post_parameters
 from django.views.generic.base import TemplateResponseMixin, TemplateView, View
 from django.views.generic.edit import FormView
 
-from ..exceptions import ImmediateHttpResponse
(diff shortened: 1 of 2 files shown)


pennersr:
Whether or not login redirect URL is a more sensible place is debatable. I think it makes more sense to somehow make the success URLs easily configurable, instead of replacing one (debatable) hardcoded URL with another.

confuzeus:
You're right and that's what I was going to do originally.

Maybe it makes more sense to have a PASSWORD_SET_REDIRECT_URL variable.

PR #2987, opened 2021-12-06
refactor(facebook): complete login

### LINKED ISSUE #2986: Update Facebook complete login
Users may want to create custom provider which extends FacebookProvider and set custom provider_id.

For example
```
from allauth.socialaccount.providers.facebook.provider import FacebookProvider


class FacebookGroupProvider(FacebookProvider):
    id = "facebook_group"
    name = "Facebook Group"


provider_classes = [FacebookGroupProvider]
```

The problem is in this case social account will be created with FacebookProvider.id instead of custom provider_id.

https://github.com/pennersr/django-allauth/blob/master/allauth/socialaccount/providers/facebook/views.py#L66
https://github.com/pennersr/django-allauth/blob/master/allauth/socialaccount/providers/facebook/views.py#L38
https://github.com/pennersr/django-allauth/blob/master/allauth/socialaccount/providers/facebook/views.py#L49
https://github.com/pennersr/django-allauth/blob/master/allauth/socialaccount/providers/base.py#L63
https://github.com/pennersr/django-allauth/blob/master/allauth/socialaccount/providers/base.py#L86

I think we don't need to update [login_by_token](https://github.com/pennersr/django-allauth/blob/master/allauth/socialaccount/providers/facebook/views.py#L73) function for custom provider.
Because we use this function for Facebook login purpose.

heindrickdumdum0217:
https://github.com/pennersr/django-allauth/issues/2986

Closes #2986

### DIFF
--- a/allauth/socialaccount/providers/facebook/views.py
+++ b/allauth/socialaccount/providers/facebook/views.py
@@ -35,7 +35,7 @@ def compute_appsecret_proof(app, token):
 
 
 def fb_complete_login(request, app, token):
-    provider = providers.registry.by_id(FacebookProvider.id, request)
+    provider = providers.registry.by_id(app.provider, request)
     resp = requests.get(
         GRAPH_API_URL + "/me",
         params={


pennersr:
Is it really necessary to pass a new parameter to the `fb_complete_login()` method? It already takes `app` as the parameter, and app has a `provider` property, so I would expect that this would do the job:

`provider = providers.registry.by_id(app.provider, request)`

heindrickdumdum0217:
You are right, let me update the code.

heindrickdumdum0217:
Updated, can you check please?

Issue #2717, opened 2020-11-17
Messages are being incorrectly escaped


frasern:
## Steps to reproduce
1. Create a user with an apostrophe in their name, for example **Alice O'Reilly**
2. Sign in as this user
3. A message will be displayed with incorrect HTML escaping:
    ```
    Successfully signed in as Alice O&#x27;Reilly
    ```

## Expected behaviour
The message should not be double-escaped:
```
Successfully signed in as Alice O'Reilly
```

## Cause
The problem lies here:

https://github.com/pennersr/django-allauth/blob/701d8ff905535e4cf79b6769ce9f60c7bed978ba/allauth/account/adapter.py#L326-L332

The call to `render_to_string()` returns a Django [SafeString](https://docs.djangoproject.com/en/dev/ref/utils/#django.utils.safestring.SafeString), but then calling `.strip()` converts it to a regular `str`, forcing escaping.

## Proposed solution

The call to `.strip()` is still necessary in order to allow blank messages to be ignored. By moving `.strip()` to the if-condition, we can pass the unaltered SafeString `message` through to `add_message()` to get the expected behaviour:

```python
 message = render_to_string(
     message_template,
     message_context,
     self.request,
 )
 if message.strip():
     messages.add_message(request, level, message, extra_tags=extra_tags)
```

9mido:
This worked for me:

```
import html
            try:
                if message_context is None:
                    message_context = {}
                message = html.unescape(render_to_string(
                    message_template,
                    message_context,
                    self.request,
                )).strip()
                if message:
                    messages.add_message(request, level, message, extra_tags=extra_tags)
            except TemplateDoesNotExist:
                pass
```

pennersr:
Solved via ce5e54c1

pennersr:
https://codeberg.org/allauth/django-allauth/issues/2717

PR #2993, opened 2021-12-13
fix(accounts) - Escape Messages Properly

9mido:
Solves https://github.com/pennersr/django-allauth/issues/2717


pennersr:
thanks, integrated via ce5e54c1

PR #2996, opened 2021-12-17
feat(account): Add unique constraint for a user's primary email

### LINKED ISSUE #225: Data model allows for having multiple primary e-mail addresses
As discussed in #210, having at most one primary e-mail address is currently not enforced at the database level. To be determined: should we fix this, e.g. by making user and primary "unique together" and using primary=NULL instead of False.

jnns:
Since Django 2.2 the UniqueConstraint supports conditions. This allows enforcing only one primary email per user on the database level.

The docs mention that [Oracle databases don't support partial indexes](https://docs.djangoproject.com/en/2.2/ref/models/indexes/#django.db.models.Index.condition). I'm not sure if that means that the index will silently be ignored (just as MariaDB and MySQL do).

Closes #225.

# Submitting Pull Requests

## General

### DIFF
--- a/allauth/account/adapter.py
+++ b/allauth/account/adapter.py
@@ -16,6 +16,7 @@ from django.contrib.auth.models import AbstractUser
 from django.contrib.auth.password_validation import validate_password
 from django.contrib.sites.shortcuts import get_current_site
 from django.core.mail import EmailMessage, EmailMultiAlternatives
+from django.db.models import CharField, EmailField
 from django.http import HttpResponse, HttpResponseRedirect
 from django.shortcuts import resolve_url
 from django.template import TemplateDoesNotExist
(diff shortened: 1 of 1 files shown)


allauth/account/adapter.py
```diff
@@ -483,15 +484,11 @@ def set_password(self, user, password):
...
-                app_settings.USER_MODEL_USERNAME_FIELD,
-                "first_name",
-                "last_name",
-                "email",
-            ],
-        )
+        return [
+            f.name
+            for f in user._meta.fields
+            if type(f) in [CharField, EmailField] and f.name != "password"
```
pennersr, line 490: The changes made here don't seem to be related to the PR title:

>  feat(account): Add unique constraint for a user's primary email #2996

derek-adair: Pretty nonsensical PR - close this.  Looks like they crossed feature branches or something...

PR #3003, opened 2022-01-02
feat(accounts): Added prefilling of forms with GET query strings

### LINKED ISSUE #2861: Feature Request: Prefill option for E-Mail on signup form
Hi,

would be nice if I could send a link to a customer with a prefilled version of the registration form.

For example:
https://www.exmaple.com/accounts/signup/?email=cumter@email.de

gmso:
Closes #2861 

## Summary
Added support for prefilling form fields with the query strings submitted on GET requests. This functionality has been added to the `login` and `signup` views. Examples:
 - https://example.com/accounts/login/?login=myuser&remember=true
 - https://example.com/accounts/signup/?username=myuser&email=user@mail.com

This feature was requested on issue #2861

# Submitting Pull Requests

## General

 - [X] If your change is substantial, feel free to add yourself to `AUTHORS`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,8 @@ pep8.txt
 *.db
 *.tmp
 virtualenv
+.venv
+.vscode
 .DS_Store
 
 *.prefs
--- a/AUTHORS
+++ b/AUTHORS
@@ -57,6 +57,7 @@ Francis Brito
 Frantisek Malina
 Fred Palmer
 Fábio Santos
+Germán Mené Santa Olaya
 George Whewell
 Griffith Rees
 Guignard Javier
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -27,6 +27,10 @@ Note worthy changes
   installed. Therefore, enabling or disabling ``sites`` is not something you can
   do on the fly.
 
+- The ``login`` and ``signup`` pages support now the prefilling of form fields with the 
+  GET query strings. Examples: 
+  ``https://example.com/accounts/login/?login=myuser&remember=true``,
+  ``https://example.com/accounts/signup/?username=myuser&email=user@mail.com``
 
 
 Backwards incompatible changes
--- a/allauth/account/tests.py
+++ b/allauth/account/tests.py
@@ -1051,29 +1051,29 @@ class AccountTests(TestCase):
         actual_message = msgs._queued_messages[0].message
         assert user.username in actual_message, actual_message
 
-
     def test_get_login_view_with_query_strings_prefills_form(self):
         from .views import LoginView
+
         form_fields = LoginView.form_class().fields
         self.assertNotEqual(form_fields, {})
 
-        query_string_dict = {k: "query" for k,_ in form_fields.items()}
+        query_string_dict = {k: "query" for k, _ in form_fields.items()}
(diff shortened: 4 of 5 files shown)


varunsaral:
Should i close this @pennersr ?

pennersr:
@varunsaral Not sure. The feature request seems valid, though the implementation seems a bit too general to me. E.g. it allows for `?email=foo@bar.com&password=secret&tos=true` -- as in, it allows for prepopulating any field including more sensitive fields such as the password or agreement with a ToS or privacy policy. Prepopulating email could make sense, the other fields not so.

varunsaral:
agreed @pennersr, i think username and email will be enough for prefilling. Or maybe we can add a blacklist for sensitive information and prefill the others.

varunsaral:
should i start with the implementation? @pennersr . Just prefiling the email/username field.

pennersr:
The only case where a feature like this could make sense is for the signup page, e.g. if you are sending out newsletters and want to make it easy for users to signup from a link in that mail. And, I have no idea how one can perform a prefill of the username for signup, what value would one choose?

- So this PR targets both login and signup, I would like to know the use case for targetting the login
- It targets both `username` and `email` -- what is the use case for `username`?

Without new insights in the above, I can imagine a simple prefill of `email` just for the signup page... though having a mixin for that is a bit overkill (`PrefillFormWithQueryStringMixin`).

varunsaral:
Alright @pennersr agreed. Login can make sense when user want to redirect the user from signup to login page with prefilled details of signup. Otherwise I don't see any use of it. Please tell if I should take up the implementation of it or drop it. I'm just trying to close all the existing pr and move to issues. Goal is to learn and fix

pennersr:
Login prefilling these days is more of a browser's job, IMO. In any case, it definitely helps to get rid of these old PRs/issues, so if you want to prep a PR that only prefills email on the signup page in a simple manner (e.g. without `PrefillFormWithQueryStringMixin`) then definitely, that will be accepted and we can close this ticket.

derek-adair:
I'd vote to close this one; very little gain for such a big change.  This is something I personally have no use for and would want it turned off.  It's probably not a *huge* deal but the idea of making it easier to exploit/script signups does not warrant saving a few key strokes for a user.

varunsaral:
@derek-adair, The change isn't big in itself(since it won't have a mixin so just a overriding a method will wokr) and i don't think it can be exploited(possibly thorough clickjacking) since it's just signup. Although i agree it don't have much gain but make easier to send prefilled forms to users for signup.

derek-adair:
Sure, just my 2 cents.  I'm a minimalist... if this type of specific functionality is desirable for other users it seems pretty straight forward to implement.

**if** it is deemed desirable I would suggest adding a configuration variable that defaults to off.

pennersr:
We already have a lot of settings, and we should only introduce new one with good reasons... but being a minimalist is not a good reason. Also, I would like to learn how this could be exploited?

derek-adair:
My thoughts were anything to make it easier to bot/brute force is bad.  However omitting the password field nullifies this advantage.  I am dispassionate about all this given that revelation.

pennersr:
Closing in favor of #3410

PR #3008, opened 2022-01-10
feat(accounts): add flag to allow login only with primary email

Marmelatze:
# Submitting Pull Requests

## General

## Description

Adds a Flag `EMAIL_ONLY_PRIMARY_LOGIN` to allow login with email addresses only for the primary address. Background for this is that I had a problem with user accounts beeing reset by outdated secondary email addresses the users did not remove.

### DIFF
--- a/allauth/account/app_settings.py
+++ b/allauth/account/app_settings.py
@@ -132,6 +132,13 @@ class AppSettings(object):
         return self._setting("EMAIL_MAX_LENGTH", 254)
 
     @property
+    def EMAIL_ONLY_PRIMARY_LOGIN(self):
+        """
+        Only allow login/password reset for primary email
+        """
+        return self._setting("EMAIL_ONLY_PRIMARY_LOGIN", False)
+
+    @property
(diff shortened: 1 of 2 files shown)


pennersr:
If `EMAIL_ONLY_PRIMARY_LOGIN` is turned on, then what use case remains for having secondary email addresses? Because if there is no use case, then you might just as well set `ACCOUNT_MAX_EMAIL_ADDRESSES` to `1` -- which would imply this new setting is not really needed?

Marmelatze:
Valid point, but the accounts in my project have multiple addresses for notifications. E.g. billing mails to address a, newsletter to b etc.

pennersr:
If I understand you correctly, you have a user with email user@domain. Additionally, the user also adds invoice@domain and newsletter@domain. But you actually don't want any of the allauth functionality to kick in. Only addressing login is insufficient, you would have to look into password reset as well -- even with your change in, the user could still request a password reset at invoice@domain. All in all, wouldn't it be better to avoid all of this complexity and create your own table of additional email addresses?

PR #3010, opened 2022-01-17
fix(account/forms): Use user instance for password validation

jnns:
The `SignupForm` is supposed to use a temporary dummy user, that the form data is written to, to enable password validation.
As it is right now, `dummy_user` is not actually a blank `User` instance but the model itself. That means that the `User` model always contains the data of the user that signed up most recently.

```python
def clean(self):
    super(SignupForm, self).clean()

    # `password` cannot be of type `SetPasswordField`, as we don't
    # have a `User` yet. So, let's populate a dummy user to be used
    # for password validaton.
    dummy_user = get_user_model()
    user_username(dummy_user, self.cleaned_data.get("username"))
    user_email(dummy_user, self.cleaned_data.get("email"))
    password = self.cleaned_data.get("password1")
```

I suppose the fix is simply to use 

```python
dummy_user = get_user_model()()
# Or better yet:
User = get_user_model()
dummy_user = User()
```

What makes this bug so nasty is that it can lead to leaked private data: For example, deferred user instances will infer data from the `User` class. Accessing `user.email` on a deferred instance will display the email address of the user that most-recently signed up.

---

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -27,6 +27,8 @@ Note worthy changes
   installed. Therefore, enabling or disabling ``sites`` is not something you can
   do on the fly.
 
+- Fixed ``SignupForm`` setting username and email attributes on the ``User`` class
+  instead of a dummy user instance.
 
 
 Backwards incompatible changes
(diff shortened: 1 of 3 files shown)


pennersr:
Thanks, merged via 16a8b7c7

PR #3049, opened 2022-03-09
fix(microsoft): Replace "TENANT" with "tenant"

adrianocorrea:
During the last update, the variable casing changed. It is causing incompatibilities with the documentation and previous versions. Reverted it to lowercase.


pennersr:
Closed via e2e30dd3

PR #3053, opened 2022-03-10
feat(Google): update URLs of endpoints

illia-v:
I updated URLs of Google endpoints to ones recommended in https://accounts.google.com/.well-known/openid-configuration.

Two `userinfo` fields were renamed:
* `id` to `sub` (also, its type changed from `int` to `str`);
* `verified_email` to `email_verified`.

Also, this update allowed skipping the `userinfo` call since an ID token received after [exchanging a code](https://developers.google.com/identity/protocols/oauth2/openid-connect?hl=en#exchangecode) contains all needed user information.

Closes #1470

---

# Submitting Pull Requests

## General

 - [x] Make sure you use [semantic commit messages](https://seesparkbox.com/foundry/semantic_commit_messages).
       Examples: `"fix(google): Fixed foobar bug"`, `"feat(accounts): Added foobar feature"`.
 - [x] All Python code must formatted using Black, and clean from pep8 and isort issues.
 - [x] JavaScript code should adhere to [StandardJS](https://standardjs.com).
 - [x] If your changes are significant, please update `ChangeLog.rst`.
 - [x] If your change is substantial, feel free to add yourself to `AUTHORS`.

 ## Provider Specifics

 In case you add a new provider:

- [ ] Make sure unit tests are available.
- [ ] Add an entry of your provider in `test_settings.py::INSTALLED_APPS` and `docs/installation.rst::INSTALLED_APPS`.
- [ ] Add documentation to `docs/providers.rst`.
- [ ] Add an entry to the list of supported providers over at `docs/overview.rst`.

PR #3069, opened 2022-04-05
Google Provider: Check for 'sub' if 'id' is not present in 'data'

Vipulkushwaha:
When ID Token is used, in place of an access token, (as in the case of an android app), a 'sub' key is present instead of 'id' in 'data'. This change considers both keys.

### DIFF
--- a/allauth/socialaccount/providers/google/provider.py
+++ b/allauth/socialaccount/providers/google/provider.py
@@ -39,7 +39,10 @@ class GoogleProvider(OAuth2Provider):
         return ret
 
     def extract_uid(self, data):
-        return str(data["id"])
+        try:
+            return str(data["id"])
+        except:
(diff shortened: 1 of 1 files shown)


Vipulkushwaha:
It seems to be. For 'id', I'm not sure why the same thing has different names.
I checked your #3053 too. :)
Are the present urls being deprecated on Google's end?

illia-v:
> It seems to be. For 'id', I'm not sure why the same thing has different names.

The newer names are compliant with [the OpenID Connect Core 1.0 specification](https://openid.net/specs/openid-connect-core-1_0.html).

> Are the present urls being deprecated on Google's end?

I am not sure. But the new ones are recommended. And using them eliminates a need for the `userinfo` request.

pennersr:
Closing in favor of #3053

PR #3080, opened 2022-04-20
feat(providers): add Django OAuth Toolkit provider

### LINKED ISSUE #3073: Idea: django-oauth-toolkit provider
I have the need to implement a custom OAuth2 provider for a service that uses [django-oauth-toolkit](https://django-oauth-toolkit.readthedocs.io/en/latest/). Instead of having only my custom provider (that is already implemented here [techmatters/terraso-allauth](https://github.com/techmatters/terraso-allauth/)), I thought that it could be interesting for the `django-allauth` project to have a generic provider that work well with any Django service using `django-oauth-toolkit`.

It seems this approach is a bit different from the current way of writing providers (to a specific library instead of to a service). So, it would be good to know if it looks like something acceptable for the project. If it's something acceptable, I can work on this front and submit a PR.

Of course, if you guys have experience doing this integration and have better approaches in mind, it would be great to hear more about :wink:

cacarrara:
This change adds an OAuth2 provider to work with Django OAuth Toolkit.
It connects with Django projects that use the toolkit library to implement the 
OAuth authorization server.

Resolves #3073


paulschreiber:
@pennersr Can you look this over?

PR #3108, opened 2022-06-03
feat(accounts): add key to password reset template

c-w:
Currently, the password reset email template gets passed the full URL to
the password reset endpoint. However, in some scenarios, this endpoint
isn't the location where the user will actually perform the password
reset data entry, for example if Django only hosts an API backend and
the user-facing part instead is implemented via a SPA.

To make it easier to integrate such setups with allauth, this change
additionally passes the uid and key to the password reset email
template. This provides full control to the template author to generate
their own password reset links and support a split frontend/backend.

### DIFF
--- a/allauth/account/forms.py
+++ b/allauth/account/forms.py
@@ -563,9 +563,10 @@ class ResetPasswordForm(forms.Form):
             # password_reset.save()
 
             # send the password reset email
+            uid = user_pk_to_url_str(user)
             path = reverse(
                 "account_reset_password_from_key",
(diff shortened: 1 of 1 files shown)


c-w:
Note the similar use-case in https://github.com/pennersr/django-allauth/pull/3081; I believe the implementation provided here is more generic as the path variable can now be reconstructed in the template via `reverse` FYI @jhhayashi

c-w:
@pennersr Any chance we could get this PR reviewed? Thanks in advance!

c-w:
TYSM for the merge!

PR #3115, opened 2022-06-10
Correcting login template file name - fixing #3114

### LINKED ISSUE #3114: Login template file is hardcoded to `login.html` but it should be respecting `ACCOUNT_TEMPLATE_EXTENSION`
The template name is hardcoded to use the extension `.html` when it shouldn't be:

https://github.com/pennersr/django-allauth/blob/4f9efa9e1d63e61a4f6cd067ef6c34b5343ef4fa/allauth/socialaccount/providers/base/mixins.py#L12

It should be working more like:

https://github.com/pennersr/django-allauth/blob/556bd220060697325d24ffce597db86a16c1e0c8/allauth/socialaccount/views.py#L30

Thank you. This is a great project.

violuke:
This should resolve #3114. Thank you.

### DIFF
--- a/allauth/socialaccount/providers/base/mixins.py
+++ b/allauth/socialaccount/providers/base/mixins.py
@@ -1,4 +1,5 @@
 from django.shortcuts import render
+from ..account import app_settings as account_settings
 
 from allauth.socialaccount import app_settings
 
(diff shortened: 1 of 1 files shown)


violuke:
Already fixed in 03abcab8adfa823f229e1222c4405dea4f54cbeb thanks 👍

Issue #3062, opened 2022-03-23
[Feature Request] Simplify email change form when ACCOUNT_MAX_EMAIL_ADDRESSES is set to 1


pktiuk:
Description from [configuration docs](https://django-allauth.readthedocs.io/en/latest/configuration.html)
>ACCOUNT_MAX_EMAIL_ADDRESSES(=None)
>   The maximum amount of email addresses a user can associate to his account. It is safe to change this setting for an already running project – it will not negatively affect users that already exceed the allowed amount. Note that if you set the maximum to 1, users will not be able to change their email address as they are unable to add the new address, followed by removing the old address.

I think in many cases, app developers would prefer setting value of `ACCOUNT_MAX_EMAIL_ADDRESSES` to 1.  
It would be convenient if form used for changing email addresses  would be simplified in such a case.

pennersr:
See #3346

pennersr:
https://codeberg.org/allauth/django-allauth/issues/3062

PR #3118, opened 2022-06-24
Allow simple e-mail replacement when MAX_EMAIL_ADDRESSES=1

pktiuk:
Closes: https://github.com/pennersr/django-allauth/issues/3062

This is just a draft to show general idea.   
But I am not whether is there any chance of getting a review and merging it, because there are a lot of PR-s without any review or comment from maintainers. Is this repo actively maintained? (If yes, I will finish this code and change from draft to proper PR)

I think this code would be very useful in cases when we expect from users having one email.  
In general this case complicates process of changing email, because you can't add a new one and remove old one, nor remove old one to add a new one (when users have to have emails linked with accounts).

## General

 ## Other things to do

- [ ] Fix tests
- [ ] Add migration
- [ ] Test it manually with my app

### DIFF
--- a/allauth/account/forms.py
+++ b/allauth/account/forms.py
@@ -634,3 +634,40 @@ class UserTokenForm(forms.Form):
             raise forms.ValidationError(self.error_messages["token_invalid"])
 
         return cleaned_data
+
+
+class EmailChangeForm(forms.Form):
+    email = forms.EmailField(label=_("New e-mail"), max_length=100, required=True)
+
+    email_rewritten = forms.EmailField(
+        label=_("Rewrite new e-mail"), max_length=100, required=True
+    )
+
+    def __init__(self, *args, user=None, **kwargs):
+        self.user = user
+        super(EmailChangeForm, self).__init__(*args, **kwargs)
+
+    def clean(self):
+        cleaned_data = super().clean()
+        if cleaned_data["email"] != cleaned_data["email_rewritten"]:
+            raise forms.ValidationError(_("E-mail addresses do not match."))
+
+    def clean_email(self):
+        email = self.cleaned_data["email"]
+        errors = {
+            "this_account": _(
+                "This e-mail address is already associated with this account."
+            ),
+            "different_account": _(
+                "This e-mail address is already associated with another account."
+            ),
+        }
+        users = filter_users_by_email(email)
+        on_this_account = [u for u in users if u.pk == self.user.pk]
+        on_diff_account = [u for u in users if u.pk != self.user.pk]
+
+        if on_this_account:
+            raise forms.ValidationError(errors["this_account"])
+        if on_diff_account:
+            raise forms.ValidationError(errors["different_account"])
+        return email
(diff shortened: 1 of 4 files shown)


e2489:
Why is there not more attention being paid to this issue? Being able to have one email address that users can change seems like extremely basic functionality. Almost every site on the internet does it this way. The way it's handled now is very outdated and overly complicated.

pktiuk:
> Is this repo actively maintained?   

It seems to be either not actively maintained, or just not open for contributors from outside.   
That's a shame.

pennersr:
This is a draft PR which is a bit short on details. It suggests a new "replace" email operation but does not explain how that would work. Particularly, in order to allow users to change their email you need to let them input a new email address, verify that one, and only once verified you can drop the old one. If you do not follow these steps users can and will lock themselves out of their account by inputting an email address with a typ-o. How would this new "replace" operation tackle this scenario?

pktiuk:
I was just unsure whether anyone would look at this PR and whether there is any need for this feature, and I wanted to check it with this basic PR. (I did not get any feedback in the linked issue, and I was not sure about that)  

> How would this new "replace" operation tackle this scenario?  

I think I will just create a simple form class validating everything.  

I will soon update this PR with a more detailed implementation.

pennersr:
But this goes beyond form validation, as you would need to store the new address, wait for it to be validated and only then you could swap. Perhaps, before starting to implement it, you could sketch how you are going to tackle it? It's difficult to judge whether a feature can be merged if it is not clear how the scenario is going to look like...

pktiuk:
> If you do not follow these steps users can and will lock themselves out of their account by inputting an email address with a typ-o.

I think the form should contain two fields with email fields to avoid simple mistakes like typos.

> But this goes beyond form validation, as you would need to store the new address

This is the case only when email verification is mandatory. In the case of optional email validation, we can just swap emails.

For the case of mandatory validation I think the best solution would be just adding the second, unconfirmed email assigned to user (I think model `EmailAddress` should get a flag called `replacement`) and sending a confirmation email.  

Only after clicking confirmation link we could delete the old email and mark the new one as validated.

e2489:
I'm new to django-allauth, so forgive me if this makes no sense or isn't plausible, but why not just add an actual 'change email' feature, separate from adding/removing addresses?

Something like an ACCOUNT_CHANGE_EMAIL configuration option, which requires ACCOUNT_MAX_EMAIL_ADDRESSES=1. If this is enabled, add/remove email becomes 'change email'. Store the pending email address, and once verified, replace the old one. If there's a typo or they try to change it again before verifying (for whatever reason), just invalidate the initial request and start over.

pktiuk:
@pennersr   
What do you think?  
I had some time and wrote some code showing what I want to do.

@e2489  
I think it is a question to repo maintainer.   
I will comply with any decision they will make.

pktiuk:
@pennersr

e2489:
Yeah, really hoping we can get this incorporated. It would be very useful.

pennersr:
Wondering, do we really need a `replacement` flag? Can't this information be derived from the fact that the max. number of email addresses is 1, yet, a second one is added and verified, implying the first one should get replaced?

pktiuk:
@pennersr   
You are right.  
When I started coding this I was not sure.

I removed `replacement` flag.

pktiuk:
@pennersr

pennersr:
Closing in favor of #3346

PR #3125, opened 2022-07-13
fix(locale): Fixed some translations to Spanish

sernamar:
`Signup` and `Sign Up` should be translated as `Registrarse` (without tilde) instead of `Regístrarse`, as that word is stressed on the penultimate syllable, not on the third-to-last syllable.

Reference (in Spanish): [Diccionario panhispánico de dudas](https://www.rae.es/dpd/tilde)


sernamar:
Commit [701e8da](https://github.com/pennersr/django-allauth/pull/3125/commits/701e8da8eb5882763483c690386277ccc7ac1313) changes `desde` to `de parte de` as `desde` is a very literal translation from English but very unnatural (at least for Spain Spanish) in this context.

PR #3131, opened 2022-07-27
save only updated fields

Akay7:
# Submitting Pull Requests

## General

 ## Provider Specifics
There is possible to get raise conditions. I.e.

1. allauth update email field of the user in the memory
2. another process update some other user fields
3. allauth save user into DB and this will override all changes on step 2

My point that is better to specify fields to update, in order to keep changes(https://docs.djangoproject.com/en/4.0/ref/models/instances/#specifying-which-fields-to-save).
But cons of that solution that could be some fields that should be updated on the user save, ie `updated_at` and because it's not specified they wont updated during save.

### DIFF
--- a/allauth/account/models.py
+++ b/allauth/account/models.py
@@ -48,7 +48,8 @@ class EmailAddress(models.Model):
         self.primary = True
         self.save()
         user_email(self.user, self.email)
-        self.user.save()
+        if app_settings.USER_MODEL_EMAIL_FIELD:
+            self.user.save(update_fields=[app_settings.USER_MODEL_EMAIL_FIELD])
         return True
 
     def send_confirmation(self, request=None, signup=False):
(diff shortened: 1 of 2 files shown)


allauth/account/models.py
```diff
@@ -48,7 +48,7 @@ def set_as_primary(self, conditional=False):
         self.primary = True
         self.save()
         user_email(self.user, self.email)
-        self.user.save()
+        self.user.save(update_fields=[app_settings.USER_MODEL_EMAIL_FIELD])
```
pennersr, line 51: **Issue**: `USER_MODEL_EMAIL_FIELD` can be `None` in case the model has no such field.

Akay7: Thanks @pennersr !
But overall PR makes sense or you don't see sense in minor changes like that?

Akay7: Great, I will update this PR then.

Akay7: I wrote tests and fix condition when `USER_MODEL_EMAIL_FIELD` is `None`

PR #3135, opened 2022-08-03
fix(template tags): Update templates.rst

leonard-colin:
Since I disabled the username field for my project, I struggled to understand how to display user email in template tags.
Until I simply used {{ user }} tag, after trying to override the view and the form to custom the get_context_data to retrieve the email.
I thought it could be useful to add it clearly in your documentation.
Your lib and documentation is great by the way, thanks for your work!

### DIFF
--- a/docs/templates.rst
+++ b/docs/templates.rst
@@ -43,6 +43,10 @@ Or, if you need to use in a ``{% blocktrans %}``::
 Then, override the ``ACCOUNT_USER_DISPLAY`` setting with your project
 specific user display callable.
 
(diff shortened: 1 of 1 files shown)


pennersr:
Note that this is actually standard Django behavior -- not something allauth takes care of:

https://github.com/django/django/blob/main/django/contrib/auth/base_user.py#L72

leonardtkt:
Indeed, thanks. But I thought it may not be obvious to everyone :)

leonard-colin:
Indeed, thanks. But I thought it may not be obvious to everyone :)

PR #3156, opened 2022-09-08
feat: Support Django 4.1

adamchainz:
# Submitting Pull Requests

## General

### DIFF
--- a/allauth/account/tests.py
+++ b/allauth/account/tests.py
@@ -4,6 +4,7 @@ import json
 import uuid
 from datetime import timedelta
 
+import django
 from django import forms
 from django.conf import settings
 from django.contrib import messages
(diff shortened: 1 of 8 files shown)


allauth/account/tests.py
```diff
@@ -158,12 +159,19 @@ def test_signup_password_twice_form_error(self):
...
+                "password2",
+                "You must type the same password each time.",
+            )
+        else:
+            self.assertFormError(
+                resp,
+                "form",
+                "password2",
+                "You must type the same password each time.",
+            )
```
adamchainz, line 174: This and similar changes are all to deal with [the test change](https://docs.djangoproject.com/en/4.1/releases/4.1/#tests) to `assertFormError()`. Without it, the test run records warnings from Django like:

```
/.../lib/python3.10/site-packages/django/test/testcases.py:279: RemovedInDjango50Warning: Passing response to assertFormError() is deprecated. Use the form object directly: assertFormError(response.context['form'], None, ...)
```

Using a branch on `django.VERSION` seems to be the most elegant way to handle the two different calling formats. If you use django-upgrade in the future, when dropping <4.1 support, it can rewrite the code to leave only the "new way of doing thigns" with its [versioned blocks fixer](https://github.com/adamchainz/django-upgrade/#versioned-blocks).

Issue #3072, opened 2022-04-14
Idea: Option to configure providers in Django admin/database


bufke:
Hello, this is a request for comment on an idea that could be implemented in django-allauth itself or as another package. Consider these two use cases:

- Open source project foo wants to support Login With GitLab. However foo project cannot know which providers settings can apply to users ahead of time. Today, this can be solved by instructing users to set environment variables that would ultimately be applied in SOCIALACCOUNT_PROVIDERS. However it's burdensome for the developer to create many environment setting mappings. And it may be burdensome to end users who are more comfortable configuring oauth in a web interface.
- SaaS project bar is multi tenanted and wants to enable organization admins to configure their own oauth. This cannot be done without greatly extending django-allauth as there is no way to set values such as GITLAB_URL which may be unique for every GitLab instance.

Proposed solution

Add a configuration JSONField to SocialApp. Let it be configured in django admin. When set, prefer this field for settings. When null, use SOCIALACCOUNT_PROVIDERS. This behavior would not introduce any breaking change nor make users adjust to a new system who don't wish to. Having two systems to do the same thing, however, would increase code maintenance burden.

pennersr:
I am not sure I fully understand, but you can already fully manage social apps without having provider settings in your settings.py. You can even have multiple providers of the same type, and hook them up to a site (as in, django.contrib.sites.models.Site). So, what kind of data do you expect to store in the proposed `JSONField` ?

bufke:
The current implementation, in my understanding, allows editing only name/client/secret keys. This is not always sufficient, as in the case of the GITLAB_URL which may be gitlab.com or gitlab.myinstance.com. Site A may want gitlab.com while Site B wants their own self hosted GitLab. It would be hard to predict which provider specific settings need tweaked per social app. The JSONField would allow anything configurable in SOCIALACCOUNT_PROVIDERS to be set in the database. URL is the most obvious thing to store, but I could imagine other settings as well. One GitLab instance may be configured differently from another. One set of users may want to sync a lot of data while another as little as possible and thus need to change scopes.

bufke:
Here's two examples of what I want to solve:

- It would remove the need (or greatly simplify) for merge requests like [this one](https://gitlab.com/glitchtip/glitchtip-backend/-/merge_requests/146). Any open source project using django-allauth would end up needing a bunch of duplicated work around this mapping of settings to allauth providers.
[...]

PR #3165, opened 2022-10-03
feat(providers): Add OpenID Connect dynamic provider

smkent:
# Submitting Pull Requests

## General

 ## Provider Specifics


smkent:
To see a working concept of this change in action, [see my PR to add OpenID Connect support to paperless-ngx](https://github.com/paperless-ngx/paperless-ngx/pull/1746).

bufke:
This looks amazing. Not my call, but I think the need to add a list in settings.py is cumbersome and would be better solved as a database value. This was my suggestion in #3072. Since you're putting the client id and secret in settings, it makes the "social application" model entirely redundant. I see why you did it this way as it's not possible to configure multiple of the same provider.

IMO if it was moved to a modified version of the social application model, it would be easier to configure and allow a developer to enable end user uploaded oidc themselves which would be good for a SaaS app.

smkent:
@bufke I'm glad you like my change!

django-allauth actually supports both database-based configuration in the social applications DB model or client IDs/keys in provider `APP` values for existing providers. This is mentioned at the [bottom of the long example block in the Installation -> Django section of the docs](https://django-allauth.readthedocs.io/en/latest/installation.html#django).

Similarly, this change also supports configuration via the database as an option instead of the `APP` config var method. For example, config like this:

```python
SOCIALACCOUNT_PROVIDERS = {
    "openid-connect": {
        "SERVERS": [
            dict(
                id="example_oidc",
                name="Example OIDC",
                server_url="https://oidc.example.com",
            ),
        ],
    },
}
```

adds a new provider to social application providers list:

scr

bufke:
Interesting! That's from _provider_factory, yes? And not typical django-allauth behavior. That mitigates 3072 - though IMO it's still not preferable for SaaS. Secret and client ID are editable in the DB, but not other configuration values. Such as a provider tenant name or in this case, server url.

I'll give your code a test on some services this week or next. It solves a few problems for an open source project I'm affiliated with - [KoboToolbox](kobotoolbox.org/). We're migrating to allauth and need to support potentially many thousands of organizations. That's why reading them in settings.py for each one isn't as nice as doing it Django Admin.

smkent:
> That's from _provider_factory, yes?

Right, `_provider_factory` dynamically creates a new provider instance for each configured OpenID Connect server.

> And not typical django-allauth behavior.

Logically, OpenID Connect is a protocol (similar to, and in fact built on top of OAuth2). django-allauth exposes implementations of protocols as providers, such as Google (which uses OAuth2) or Keycloak (which uses OpenID Connect in the existing codebase). 

I wanted to keep this change consistent with the existing design patterns. To support arbitrary OpenID Connect providers and keep them as logical top-level objects like the other providers, I'm using a factory function to create independent classes for each configured OpenID Connect provider.

> That mitigates 3072 - though IMO it's still not preferable for SaaS. Secret and client ID are editable in the DB, but not other configuration values. Such as a provider tenant name or in this case, server url.

Your use case in #3072 is definitely reasonable, and django-allauth would likely benefit from more configuration flexibility. IMO a change on that front should be separate from a change like this, which simply adds new provider(s) matching the existing design pattern.

> I'll give your code a test on some services this week or next. It solves a few problems for an open source project I'm affiliated with - [KoboToolbox](kobotoolbox.org/). We're migrating to allauth and need to support potentially many thousands of organizations. That's why reading them in settings.py for each one isn't as nice as doing it Django Admin.

Thanks for giving my change a try and I hope it works well!

bufke:
I tested with Google Apps and Azure and it worked great! I can add multiple Azure tenants without resorting to any hacks like creating duplicate provider classes! Azure's server url has the tenant ID in it already.

The instructions made sense to me and explained the unique oidc/ID callback url. I agree you shouldn't blow up the scope of this PR.

bufke:
Tested with OneLogin but couldn't get it working. I'll try to debug. They seem to support OIDC. Their server URL is https://<tenant name>.onelogin.com/oidc/2

allauth/socialaccount/providers/openid_connect/views.py
```diff
@@ -0,0 +1,40 @@
...
+    @property
+    def authorize_url(self):
+        return self.openid_config["authorization_endpoint"]
+
+    @property
+    def profile_url(self):
+        return self.openid_config["userinfo_endpoint"]
+
+    def complete_login(self, request, app, token, response):
+        response = requests.post(
```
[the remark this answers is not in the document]
smkent: Great catch, this should definitely be a `GET` request. [The existing provider code for Keycloak makes a `POST` request for user info](https://github.com/pennersr/django-allauth/blob/69e571f89e5b98cb2eb308b8b788532d6746f0b6/allauth/socialaccount/providers/keycloak/views.py#L33), which is how this ended up in my change. I've updated this PR to use a `GET` request instead.

smkent:
Hi @pennersr -- just a friendly ping! It would be great to get this merged. Generalized OpenID Connect support would make django-allauth useful for a number of applications such as paperless-ngx and Tandoor Recipes. I'm more than happy to consider any of your feedback. Thanks!

allauth/socialaccount/providers/openid_connect/views.py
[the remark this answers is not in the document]
smkent: The `master` branch now performs a `GET` request on the userinfo endpoint in the Keycloak provider as of 39f8cafbf58c98f08193eb353e117405330abb14. Using `GET` here is no longer new functionality in this PR.

smkent:
I've resolved the merge conflict on `ChangeLog.rst`.

kurokay:
Nice, i'm waiting this feature to be merged. Will be awesome !

muhmuhhum:
I have a small little problem when i want to set the the OAUTH_PKCE_ENABLED setting for a openId_connect social account.
`
  "openid-connect'" {
              "SERVERS": [
                  {
                      "id": "testAuth",
                      "name": "Test Auth",
                      "server_url": SOCIAL_OIDC_URL,
                      "APP": {
                          'client_id': SOCIAL_OIDC_CLIENT_ID,
                          'secret': SOCIAL_OIDC_SECRET
                      }
                  }
              ]
          }
`

For this i have to make a workaround
`
'openid-connect_testAuth':{
            'OAUTH_PKCE_ENABLED': True
        }
`

smkent:
> I have a small little problem when i want to set the the OAUTH_PKCE_ENABLED setting for a openId_connect social account. `"openid-connect'" { "SERVERS": [ { "id": "testAuth", "name": "Test Auth", "server_url": SOCIAL_OIDC_URL, "APP": { 'client_id': SOCIAL_OIDC_CLIENT_ID, 'secret': SOCIAL_OIDC_SECRET } } ] }`
> 
> For this i have to make a workaround `'openid-connect_testAuth':{ 'OAUTH_PKCE_ENABLED': True }`

@muhmuhhum Thanks for letting me know! I've updated how the OpenID Connect per-server settings are accessed internally and added a corresponding unit test. It should now be possible to set `OAUTH_PKCE_ENABLED` like this:

```python
"openid-connect": {
    "SERVERS": [
        {
            "id": "testAuth",
            "name": "Test Auth",
            "server_url": SOCIAL_OIDC_URL,
            "OAUTH_PKCE_ENABLED": True,
            "APP": {
                "client_id": SOCIAL_OIDC_CLIENT_ID,
                "secret": SOCIAL_OIDC_SECRET,
            },
        }
    ]
}
```

andrey-toptal:
any review updates on this PR?

mikapfl:
Hi,

When testing this PR, I encountered two difficulties. I configured an oidc provider (a self-hosted Ory Hydra), using a config like (anonymized):
```
SOCIALACCOUNT_PROVIDERS = {
    'openid-connect': {
        'SERVERS': [
            dict(
                id='oidc.servers.tld',
                name="Custom Account System",
                server_url="https://accounts.example.org",
                APP=dict(
                    client_id="example",
                    secret="supersecure",
                ),
            ),
        ]
    }
}
```

Difficulties:
1. Apparently, the implementation uses `client_secret_post` as a token endpoint authentication method. In the default configuration, my OIDC Provider did not allow this. Maybe document this somewhere?
2. The generated provider ID was too long and wouldn't fit into the database, and I got a `django.db.utils.DataError: value too long for type character varying(30)` exception. Apparently, the provider factory adds at least 15 characters of provider ID because my id was 16 characters long. After switching to an ID with only 3 characters, things started working.

Debugging these problems was rather challenging because even with `DEBUG = True` in the django settings, no debugging information is displayed by the [authentication_error template](https://github.com/pennersr/django-allauth/blob/master/allauth/templates/socialaccount/authentication_error.html). I had to patch the template to find about the token endpoint authentication method mismatch.

After debugging and fixing these two issues, the provider works like a charm and pulls information. One final thing I noticed is that the `email_verified` claim in the oidc spec is not used. It could be used to mark the email as verified already and not send out the verification email.

Thanks for your work!

Cheers

Mika

smkent:
@mikapfl Thank you for your detailed feedback!

I've added some changes:

* I noted the provider ID 15-character limit in the documentation's configuration example. We could gain a few more characters if the `openid-connect_` internal ID prefix is changed to something shorter. I'm open to doing that if folks feel strongly.
* The OAuth2 provider base (which the OpenID Connect provider inherits from) supports token retrieval via both HTTP Basic authentication and HTTP POST data. I added support for both options in the OpenID Connect provider. Each server can now have an optional `token_auth_method` value (`client_secret_basic` or `client_secret_post`) to use a specific method. Otherwise, a method is selected from the server's `token_endpoint_auth_methods_supported` list in the `.well-known/openid-configuration` response. (This new behavior also applies to the Keycloak provider since it inherits from OpenID Connect.)
* I added support for the `email_verified` claim in the userinfo response by implementing `extract_email_addresses` on the provider.

Please let me know if you encounter any more issues with the OpenID Connect provider.

smkent:
Hi @pennersr -- just another friendly ping! A number of folks are finding this feature useful and would benefit from this PR being merged and included in a new release. Please let me know if you have any review feedback and/or what else I can do to help move this feature forward. Thank you!

smkent:
With @pennersr's announcement today in #3201 regarding their limited time for django-allauth, I'm not sure what the timeline might look like for getting this feature merged and released. I've started a discussion about making a continuity plan for django-allauth in https://github.com/pennersr/django-allauth/issues/3202 and would love to hear people's thoughts.

allauth/socialaccount/providers/openid_connect/provider.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+class OpenIDConnectProviderAccount(ProviderAccount):
+    def to_str(self):
+        dflt = super(OpenIDConnectProviderAccount, self).to_str()
+        return self.account.extra_data.get("name", dflt)
+
+
+class OpenIDConnectProvider(OAuth2Provider):
+    id = "openid-connect"
```
pennersr, line 15: The convention is that the `id` matches the package name, `openid_connect` in this case. Though, perhaps it would be best to rename the package (and the `id`) to `oidc`?

allauth/socialaccount/providers/openid_connect/urls.py
```diff
@@ -0,0 +1,45 @@
...
+            name=provider.id + "_login",
+        ),
+        path(
+            "login/callback/",
+            OAuth2CallbackView.adapter_view(adapter_class),
+            name=provider.id + "_callback",
+        ),
+    ]
+
+    return [path(provider.get_slug() + "/", include(urlpatterns))]
```
pennersr, line 40: I am not sure we should nest OIDC based providers below a  `/accounts/oidc/` path. The fact that such providers are using this subprovider mechanism is just an implementation detail that better not shine through on the outside.  Not having the "/oidc/" in between would make it easier to move to (or perhaps even away from) an OIDC based subprovider.

allauth/socialaccount/providers/openid_connect/views.py
```diff
@@ -0,0 +1,49 @@
...
+    def profile_url(self):
+        return self.openid_config["userinfo_endpoint"]
+
+    def complete_login(self, request, app, token, response):
+        response = requests.get(
+            self.profile_url, headers={"Authorization": "Bearer " + str(token)}
+        )
+        response.raise_for_status()
+        extra_data = response.json()
+        extra_data["id"] = extra_data["sub"]
```
pennersr, line 46: This looks like we're deviating from the standard -- is it really needed to do this by default, or just a mechanism to cope with legacy situations?

docs/providers.rst
```diff
@@ -1485,6 +1485,54 @@ The OpenID provider can be forced to operate in stateless mode as follows::
...
+The OpenID Connect provider provides a dynamic instance for each configured
+server. To expose an OpenID Connect server as an authentication method,
+configuration of one or more servers is required:
+
+.. code-block:: python
+
+    SOCIALACCOUNT_PROVIDERS = {
+        'openid-connect': {
+            'SERVERS': [
+                dict(
```
pennersr, line 1501: It would be more readable to stick to one way of defining structures (so `{` instead of `dict(`).

pennersr (changes requested):

allauth/socialaccount/providers/openid_connect/provider.py
smkent: I changed the provider ID to `openid_connect` as suggested. I'm happy to change it to `oidc` if folks feel strongly. If we were to rename to `oidc`, the `openid_connect` provider directory should also be renamed to `oidc`.

allauth/socialaccount/providers/openid_connect/urls.py
smkent: I have removed the `/oidc` component from the OpenID Connect URL paths.

allauth/socialaccount/providers/openid_connect/views.py
smkent: I have removed this. The OpenID Connect provider's `extract_uid` method now simply consumes `extra_data["sub"]`. 

[As described in the spec, `sub` in the userinfo endpoint response contains the user ID.](https://openid.net/specs/openid-connect-core-1_0.html#UserInfoResponse)

docs/providers.rst
smkent: I have replaced `dict()` constructor usage with `{}` literals.

smkent:
Hi @pennersr, thank you very much for your feedback! I have added updates based on your comments. This PR is now ready for another look.

allauth/socialaccount/providers/openid_connect/urls.py
smkent: Note that this is a change to the OpenID Connect provider's `get_slug` method via https://github.com/pennersr/django-allauth/pull/3165/commits/cc77a03aecf1ad68caac10753bba40e2ed075318.

pennersr:
When you add an OpenID connect provider with id `foo` the resulting `SocialAccount.provider` field will be filled with `openid_connect_foo`. This is problematic because that field has a max length of 30, 15 of those get consumed by `openid_connect_`.  I would prefer a shorter prefix, and also use a different separator, e.g. `oidc:foo`.  

Alternatively, you can also argue that we should not expose internal implementation details of provider `foo` at the database level, and just store `foo`. 

Currently this PR only supports configuration via the settings. We could/should extend this to also support `SocialApp` based configuration.  We don't need to do that as part of this PR, but if we were to pick that up, I think it would look something like this:

- `SocialApp` needs a new URL field (for `server_url`).
- We would need to turn `SocialApp.provider` into a free format text field.
- In that field, you would need to enter `oidc:foo`, or perhaps friendlier, add another new field `type` that would be filled with `oidc`, and set `provider` to just `foo`.

All things considered, I think having just `foo` makes most sense. What do you think?

smkent:
> When you add an OpenID connect provider with id `foo` the resulting `SocialAccount.provider` field will be filled with `openid_connect_foo`. This is problematic because that field has a max length of 30, 15 of those get consumed by `openid_connect_`. I would prefer a shorter prefix, and also use a different separator, e.g. `oidc:foo`.
> 
> Alternatively, you can also argue that we should not expose internal implementation details of provider `foo` at the database level, and just store `foo`.
> 
> [...]
>
> All things considered, I think having just `foo` makes most sense. What do you think?

@pennersr Removing the `openid_connect_` prefix works for me. I removed it and updated the PR.

> Currently this PR only supports configuration via the settings. We could/should extend this to also support `SocialApp` based configuration. We don't need to do that as part of this PR, but if we were to pick that up, I think it would look something like this:
> 
>     * `SocialApp` needs a new URL field (for `server_url`).
> 
>     * We would need to turn `SocialApp.provider` into a free format text field.
> 
>     * In that field, you would need to enter `oidc:foo`, or perhaps friendlier, add another new field `type` that would be filled with `oidc`, and set `provider` to just `foo`.

I agree that this is a great conversation to have separately.

docs/providers.rst
```diff
@@ -1486,6 +1486,54 @@ The OpenID provider can be forced to operate in stateless mode as follows::
...
+server. To expose an OpenID Connect server as an authentication method,
+configuration of one or more servers is required:
+
+.. code-block:: python
+
+    SOCIALACCOUNT_PROVIDERS = {
+        "openid_connect": {
+            "SERVERS": [
+                {
+                    "id": "my-server",  # 15 characters or less
```
pennersr, line 1503: **Issue**: By dropping the prefix, this 15 characters remark and the one below can be bumped to 30.

smkent: I updated both comments to say 30 characters or less.

pennersr:
Merged, thanks!

smkent:
@pennersr Thank you so much! I'm eager to consume this functionality in some other projects. Would you be able to release a new version of django-allauth? Thanks again!

em-:
Thank you both for implementing and shipping this! I am pretty sure we will be playing with this soon. :D

PR #3203, opened 2022-12-05
Feat pinterest

MerleLiuKun:
Hi, I found the `Pinterest` now upgrade the api version to `v5`.  So I add this support.

But this version has some breaking changes. 

1. Authorize url has changed, And `scopes` has changed
2. Not return `uid`, And use the `username` to replace `uid`
3. Profile response structure has changed.

Maybe need drop the `v1` api, What do you think of it? @pennersr 

For now I just make the `v5` works well.

And the migrate docs for `Pinterest` at [Migration](https://developers.pinterest.com/docs/getting-started/migration/)

# Submitting Pull Requests

## General

 ## Provider Specifics

### DIFF
--- a/allauth/socialaccount/providers/pinterest/provider.py
+++ b/allauth/socialaccount/providers/pinterest/provider.py
@@ -4,6 +4,9 @@ from allauth.socialaccount.providers.oauth2.provider import OAuth2Provider
 
 class PinterestAccount(ProviderAccount):
     def get_profile_url(self):
+        # v5 extra_data not same as v1
+        if "profile_image" in self.account.extra_data:
+            return self.account.extra_data.get("profile_image")
         return self.account.extra_data.get("url")
 
     def to_str(self):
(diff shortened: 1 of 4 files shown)


pennersr:
Given that the user `id` is replaced with `username` -- does that mean the existing users will lose access to their account? When they login, I assume a new account will be created because `username != id` ? Is there a way around this? If not, a note about `Backwards incompatible changes` is in order in the `ChangeLog.rst`.

MerleLiuKun:
If the api version set before `v5`, The `id` will still returned. 

So current app may not make any changes.

However, I am not sure if the app will automatically upgrade to `v5` after June 30, 2023 when EOF for `v3` and `v4` is reached. If the behavior is automatically migrated, it will be a breaking change.

And the current `v3` may still need some update. I will do this today.

MerleLiuKun:
It's very confused for the different version api.

`v1` has been deprecated, and can not visit. So the current code not worked anymore.

`v3` has changed, and not same as `v1`.  So add `v3` support.

PR #3225, opened 2022-12-30
fix: tox: Upgrade black, flake8, isort

cclauss:
# Submitting Pull Requests

## General

 - [x] All Python code must be formatted using Black, and clean from pep8 and isort issues.

 ## Provider Specifics

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,3 @@
 [flake8]
-max-line-length = 2424
+max-line-length = 88
 ignore = E203, E402, W503
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,3 @@
 [flake8]
-max-line-length = 88
-ignore = E203, W503, E501, E231  # Black
+max-line-length = 2424
+ignore = E203, E402, W503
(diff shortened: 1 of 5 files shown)


cclauss:
Flake v6.0.0 drops support for Python 3.7 so we need to use flake8 v5.
* https://github.com/PyCQA/flake8/blob/main/docs/source/release-notes/6.0.0.rst

setup.cfg
```diff
@@ -1,3 +1,3 @@
 [flake8]
-max-line-length = 88
-ignore = E203, W503, E501, E231  # Black
+max-line-length = 2424
```
pennersr, line 2: **Issue**:  This line length does not match:

https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length

setup.cfg
```diff
@@ -1,3 +1,3 @@
...
+ignore = E203, E402, W503
```
pennersr, line 3: **Issue**: We did not need E402 before, why now?

pennersr (changes requested):

setup.cfg
cclauss: Mods to `docs/conf.py`

setup.cfg
cclauss: Let’s try

setup.cfg
pennersr: I would prefer not to add ignore rules affecting the whole code base for an issue in `docs/conf.py`.  If there is an issue in that file, it can be marked using `# noqa`.

PR #3227, opened 2022-12-31
Fix instagram provider scope "basic" is not valid

### LINKED ISSUE #3226: Instagram provider scope "basic" is not valid
This issue has been discussed before at #2454 and #2514 but was never fixed

`basic` scope was used to read user's profile and media but now according to [documentation](https://developers.facebook.com/docs/instagram-basic-display-api/guides/getting-access-tokens-and-permissions) it has changed and `user_profile` scope is required and `user_media` is optional and `basic` scope does not work anymore.

image

When trying to login with `basic` scope this error response comes up:
```
{"error_type": "OAuthException", "code": 400, "error_message": "No valid scopes requested"}
```

Hisham-Pak:
Closes #3226

### DIFF
--- a/allauth/socialaccount/providers/instagram/provider.py
+++ b/allauth/socialaccount/providers/instagram/provider.py
@@ -23,7 +23,7 @@ class InstagramProvider(OAuth2Provider):
         return data
 
     def get_default_scope(self):
-        return ["basic"]
+        return ["user_profile", "user_media"]
 
     def extract_uid(self, data):
         return str(data["id"])


allauth/socialaccount/providers/instagram/provider.py
```diff
@@ -23,7 +23,7 @@ def extract_extra_data(self, data):
         return data
 
     def get_default_scope(self):
-        return ["basic"]
+        return ["user_profile", "user_media"]
```
pennersr, line 26: **Issue**: By default we should request the minimum scope to get the auth flow going. If a project needs more scopes, that can always be configured. Therefore, can we drop `user_media` here as it is not needed for auth purposes?

pennersr:
Merged via 30a3473b

PR #3230, opened 2023-01-09
fix(accounts): Friendlier implementation of account enumeration protection

francoisfreitag:
Since commit a807e5db411c4762310552bb4b882ca9285888f1, the user returned
by `SignupForm.save()` can be `None`. That causes issues for subclasses
not designed to handle receiving `None` when the saving failed (see [1]).

Maintaining the old behavior is possible, the view should generate the
response and send the email, based on the form state. It also makes the
interface clearer: the form validates data, the view decides on how to
present it to users.

Prevent users from inadvertently calling `SignupForm.save()` on a form
instance with `account_already_exists=True`, which could result in
overwriting another user data. For example, that mistake could happen
when subclassing SignupView and forgetting to call
`super().form_valid()`.

[1]
```python
class MySignupForm(SignupForm):
    def save(self, request):
        user = super().save(request)
        # Does not expect user to be `None`.
        do_something_with_user(user)
        return user
```

# Submitting Pull Requests

## General

 ## Provider Specifics

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -1,3 +1,13 @@
+UNRELEASED
+*******************
+
+Note worthy changes
+-------------------
+
+- Make the account enumeration prevention in ``BaseSignupForm`` friendlier to
+  inheritance.
+
+
 0.54.0 (2023-03-31)
 *******************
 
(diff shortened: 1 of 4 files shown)


pennersr:
Indeed. Though, moving it into the view would cause duplication, because the social signup form needs to take the same steps. I've refactored things over here: f58889cf

PR #3235, opened 2023-01-13
fix(accounts): Make `request` object available to AccountAdapter during sing-up

irfanka:
The `request` object is already forwarded from LoginView -> LoginForm -> DefaultAccountAdapter. This does the same for SignupView -> BaseSignupForm -> DefaultAccountAdapter.

The purpose is to enable lib users who override `DefaultAccountAdapter` to make sign-up decisions based on `request`. For example, handling signups differently for different Sites.

### DIFF
--- a/allauth/account/forms.py
+++ b/allauth/account/forms.py
@@ -283,6 +283,7 @@ class BaseSignupForm(_base_signup_form_class()):
     )
 
     def __init__(self, *args, **kwargs):
+        self.request = kwargs.pop("request", None)
         email_required = kwargs.pop("email_required", app_settings.EMAIL_REQUIRED)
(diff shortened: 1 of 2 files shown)


maxlienen:
I am very sorry, I used the wrong issue id!

pennersr:
This must be listed as a backwards incompatible change, as the signature of the form `__init__` is changed.

pennersr:
The fact that `request` is not available globally has left its mark on the code over the years. Some functions get explicitly passed a request, some don't, and it is time to put a halt to passing it around in just a few extra places. Instead, it would be better to move to a solution where `request` is just globally available through a thread local / context var.

pennersr:
The above comment is handled via: #3412

Issue #3237, opened 2023-01-17
OAuth 2.0 PKCE Code Verifier longer than allowed


maxlienen:
RFC 7636 states that the code verifier may be with a minimum length of 43 characters and a maximum length of 128 characters.

The following code, used in the `generate_code_challenge` method in [utils.py](https://github.com/pennersr/django-allauth/blob/master/allauth/socialaccount/providers/oauth2/utils.py), outputs a code verifier with a random length between 58 and 171 characters using Python 3.8.10:

```
nbytes = random.randint(43, 128)
code_verifier = token_urlsafe(nbytes)
```

I think the main issue here is that bytes are mistaken with characters.

Another question for me would be why the code verifier length is random, a I do not surely know what the randomness adds, which is also not mentioned in RFC 7636.

So, for example, a fix for this would be:

```
# Create a code verifier with 128 characters
code_verifier = token_urlsafe(96)
```

This would always output a `code_verifier` of length 128. Using the maximum length adds entropy and, by this, security.

pennersr:
https://codeberg.org/allauth/django-allauth/issues/3237

PR #3238, opened 2023-01-17
fix(oauth2): Set PKCE Code Verifier length to 128

maxlienen:
Set the length of the generated PKCE code verifier to exactly 128 characters, as it is the maximum length allowed by RFC 7636. Using the maximum length adds entropy and, by this, security.

fixes pennersr/django-allauth#3237

### DIFF
--- a/allauth/socialaccount/providers/oauth2/utils.py
+++ b/allauth/socialaccount/providers/oauth2/utils.py
@@ -1,6 +1,5 @@
 import base64
 import hashlib
-import random
 
 
 try:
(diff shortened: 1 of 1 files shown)


allauth/socialaccount/providers/oauth2/utils.py
```diff
@@ -17,9 +17,8 @@ def token_urlsafe(nbytes=None):
 
 
 def generate_code_challenge():
-    # minimum length of 43 characters, maximum length of 128 characters
-    nbytes = random.randint(43, 128)
-    code_verifier = token_urlsafe(nbytes)
+    # Create a code verifier with a length of 128 characters
+    code_verifier = token_urlsafe(96)
```
pennersr, line 21: **Issue**: `import random` has now become an unused import -- can you remove it?

PR #3259, opened 2023-02-13
feat(account): Add SEND_ACCOUNT_ALREADY_EXISTS_EMAIL config

half0wl:
When `ACCOUNT_PREVENT_ENUMERATION` is set to `True`, signing up with an existing email address will send an "account already exists" email to the user. I'd like to skip sending that email and let it no-op silently.

I'm currently overriding the method in the form, and I think this would be a good configuration option to have.

## General

 ## Provider Specifics

n/a

### DIFF
--- a/allauth/account/app_settings.py
+++ b/allauth/account/app_settings.py
@@ -51,6 +51,10 @@ class AppSettings(object):
         return self._setting("PREVENT_ENUMERATION", True)
 
     @property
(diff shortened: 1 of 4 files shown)


pennersr:
Can you elaborate why making this configurable is good to have? Under what circumstances would you turn sending those emails off? Users forgetting they already signed up before happens, and not sending a mail results in a bad UX as the user has no clue what is going on.

PR #3287, opened 2023-03-20
fix(socialaccount): Remove hard-coded redirect URL

c-w:
This pull request moves all uses of `reverse("socialaccount_connections")` into the social account adapter so that they can be consistently overwritten in just one place. There was already a helper method for to format the redirect URL, but it wasn't used in all edge cases.

Besides increasing code consistency, the changes in this pull request also fix using django-allauth in combination with [dj-rest-auth](https://github.com/iMerica/dj-rest-auth/) in the scenario where we have an existing account for a new social login. Previously, we would experience a crash unless a dummy URL is configured for `socialaccount_connections`. After the fix in this PR is merged, instead of setting this dummy URL we can now override the redirect URL in the adapter which is much cleaner.

### DIFF
--- a/allauth/socialaccount/adapter.py
+++ b/allauth/socialaccount/adapter.py
@@ -116,7 +116,6 @@ class DefaultSocialAccountAdapter(object):
         Returns the default URL to redirect to after successfully
         connecting a social account.
         """
-        assert request.user.is_authenticated
         url = reverse("socialaccount_connections")
         return url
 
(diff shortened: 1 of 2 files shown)


allauth/socialaccount/adapter.py
```diff
@@ -116,7 +116,6 @@ def get_connect_redirect_url(self, request, socialaccount):
         Returns the default URL to redirect to after successfully
         connecting a social account.
         """
-        assert request.user.is_authenticated
```
pennersr, line 119: **Question**: Why is this `assert` being removed? Same question for the one below.

c-w:
Thanks for the merge 🎉

PR #3329, opened 2023-06-21
fix(is_safe_url): respect allowed_hosts

kateEcobb:
Like those who have come before me (https://github.com/pennersr/django-allauth/pull/3264), I ran into an issue where I was unable to redirect to an absolute URL after a user successfully logs in. However, the above fix doesn't take into consideration wildcard ('*') domain matching. 

The main issue is that in `is_safe_url`, `allowed_hosts` defaults to `None` and doesn't take into account Django's standard ALLOWED_HOSTS setting.

Hopefully we can put this [issue](https://github.com/pennersr/django-allauth/issues/1766) to rest finally!

# Submitting Pull Requests

## General

### DIFF
--- a/allauth/account/adapter.py
+++ b/allauth/account/adapter.py
@@ -508,8 +508,8 @@ class DefaultAccountAdapter(object):
         allowed_hosts = {self.request.get_host()} | set(settings.ALLOWED_HOSTS)
 
         if "*" in allowed_hosts:
-            parsed_host = urlparse(url).netloc
-            return url_has_allowed_host_and_scheme(url, allowed_hosts={parsed_host})
+            parsed_host = {urlparse(url).netloc} if parsed_host else None
+            return url_has_allowed_host_and_scheme(url, allowed_hosts=parsed_host)
 
         return url_has_allowed_host_and_scheme(url, allowed_hosts=allowed_hosts)
 
--- a/allauth/account/adapter.py
+++ b/allauth/account/adapter.py
@@ -2,6 +2,7 @@ import html
 import json
 import warnings
 from datetime import timedelta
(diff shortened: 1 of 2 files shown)


allauth/account/adapter.py
```diff
@@ -503,7 +504,14 @@ def is_safe_url(self, url):
                 is_safe_url as url_has_allowed_host_and_scheme,
             )
 
-        return url_has_allowed_host_and_scheme(url, allowed_hosts=None)
+        # get_host already validates the given host, so no need to check it again
+        allowed_hosts = {self.request.get_host()} | set(settings.ALLOWED_HOSTS)
+
+        if "*" in allowed_hosts:
+            parsed_host = urlparse(url).netloc
```
pennersr, line 511: If `url` is relative,`parsed_host` will be the empty string. In that case, shouldn't `allowed_hosts` be `None`?

pennersr (changes requested):

allauth/account/adapter.py
kateEcobb: Looking at the source code for `url_has_allowed_host_and_scheme`, there's a check for `not url_info.netloc` which will return true, see here: https://github.com/django/django/blob/main/django/utils/http.py#L298

In this case it doesn't matter if allowed_hosts is None or a set because django assumes relative URLs are safe.

pennersr: Agreed, though if we write it like `allowed_hosts={parsed_host} if parsed_host else None` at least the code is understandable without having to rely on the inner workings of the Django code.

kateEcobb: fair enough, will make that change now!

Issue #3265, opened 2023-02-17
ACCOUNT_EMAIL_CONFIRMATION_COOLDOWN creates failing tests


Hafnernuss:
I upgraded all my dependencies today, and this setting caused me quite some headache.
I have some custom sign-up logic that I test quite extensively, and because of this setting, random tests kept failing, but not when executing them individually. 

Maybe there is a clean way to disable this setting when testing, or a hint in the documentation?  🤔

pennersr:
https://django-allauth.readthedocs.io/en/latest/faq.html#why-did-you-just-close-my-issue

pennersr:
https://codeberg.org/allauth/django-allauth/issues/3265

PR #3337, opened 2023-07-03
docs: update doc for ACCOUNT_EMAIL_CONFIRMATION_COOLDOWN 

Hafnernuss:
This PR adds a short section to the docs, hinting that during testing, the `ACCOUNT_EMAIL_CONFIRMATION_COOLDOWN` setting might cause issues with unexpected failures.

#3265

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -76,6 +76,9 @@ ACCOUNT_EMAIL_CONFIRMATION_COOLDOWN (=180)
   users from sending too many of these mails, a rate limit is in place that
   allows for one confirmation mail to be sent per the specified cooldown period
   (in seconds).
+  It is recommended to change this setting to ``"0"`` when running tests, 
+  as it can cause unintended and sporadic failures when your testcase
+  relies on sent mails.
 
 ACCOUNT_EMAIL_MAX_LENGTH(=254)
   Maximum length of the email field. You won't need to alter this unless using


pennersr:
I don't think recommendation is sound. I wouldn't recommend altering the behavior of the code under test by using a different setting compared to when running in production. If the cooldown behavior is causing you trouble during testing, it suggests you may have leakage going on between tests, possibly you are not using `"django.core.cache.backends.dummy.DummyCache"`?

Hafnernuss:
I have no caching backends activated at all, and in fact, the [test settings](https://github.com/pennersr/django-allauth/blob/aed4d5400951c0c519e288bcf4bf1e0598ad9ab4/test_settings.py#L211) have set this to `0` as well.

pennersr:
Let me rephrase a bit -- what I meant was, the cooldown behavior is not unexpected, it is part of the expected behavior. So to me, labeling that as unexpected and recommending for it to be turned off during testing is not something I would recommend. If there are test cases that rely on the cooldown behavior to be turned of, then those tests should explicitly set it (`@override_settings`) to 0 indeed. If there is leakage across tests, that could point to caching issues. What kind of issues did you run into?

Hafnernuss:
I guess the issue was, that all those tests were part of the same test class and this caused leakages (standard django tests).
I was surprised by the change overall, because it seemes to be a minor update that was greeted with a (to me) breaking chance,  and as ... I think _it is_. It's a great feature, but I would disable it during tests any day, because it's not something I would test in a consuming project. 
I just wanted to make sure that confirmation mails are sent directly, as I am providing my own signup endpoints. 

It could very well be that the fault was on my side from the beginning, but caching is definitly no issue here. 
Personally, I'd like to have had this setting set to `0` by default, but that can be argued both ways.

Hafnernuss:
>then those tests should explicitly set it (@override_settings) to 0 indeed.
Yes, that I'd agree on that. Would you be happy with a rephrase on the PR, so that the doc would recommend overriding this setting for said use-case?

pennersr:
> that all those tests were part of the same test class and this caused leakages (standard django tests).

Hmm, given 2 test cases within the same `TestCase` class, I don't think that cache keys set in one test case leak through the other test case. And, as you mention, you did not have any caching backends configured. So that suggests that your test case was actually triggering the normal/expected cooldown behavior within one test case?

Hafnernuss:
Yes, exactly. Let's say I have two test cases for signup, one tests signup with an optional field enabled, and one signup with this field disabled. In any case, the user is signed up. Either one of those tests did not trigger the email, because of that cooldown.

yamac-kurtulus:
I also have realized that there is leakage between tests. It's been really hard to identify the problem with all-auth. I upgraded from django 3.2 to 4.2 along with all the rest of the packages. My authorization package is normally another package (dj-rest-auth) but it uses all-auth underneath for registration, so it's been really hard to find the culprit.

Anyway, i have 2 tests using registration within same class (though the issue did not go away when I separated them). When I run them one by one, it is fine, but the test suite fails due to throttling. I had to override the ACCOUNT_EMAIL_CONFIRMATION_COOLDOWN setting to 0 during these two cases. Pinpointing the issue was really annoying -.- 

Is this really the expected behavior?

PR #3340, opened 2023-07-09
Add support for SAML 2.0 

pennersr:
# Submitting Pull Requests

## General

 ## Provider Specifics

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,6 +47,8 @@ jobs:
       uses: actions/setup-python@v4
       with:
         python-version: ${{ matrix.python-version }}
+    - name: Install xmlsec
+      run: sudo apt-get install -y xmlsec1 libxmlsec1-dev
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
@@ -77,6 +79,8 @@ jobs:
       uses: actions/setup-python@v4
       with:
         python-version: '3.11'
+    - name: Install xmlsec
+      run: sudo apt-get install -y xmlsec1 libxmlsec1-dev
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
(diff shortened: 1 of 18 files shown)


akx:
You might want to consider https://github.com/akx/smolsaml instead of onelogin.saml – it was born out of frustrations with the onelogin implementation (you can see a bunch of my PRs in the onelogin repo before I gave up)...

pennersr:
The reason for going with `python3-saml` is that is smaller in scope compared to e.g. `pysaml2`, as the former focuses only on the SP side whereas the later covers both SP and IdP. But more importantly, `python3-saml` does have a lot of mileage, it is the SAML implementation used by e.g. Sentry as well. In the end, the SAML dependency is just an implementation detail of allauth. Would definitely want to switch to something smaller, but as it is now the mileage made by `python3-saml` is a deciding factor.

As for optional, definitely -- you will need to explicitly `pip install django-allauth[saml]` to pull this in.

akx:
Yeah, I evaluated `pysaml2` too before deciding to write `smolsaml` for our (@valohai) uses. Anyway, glad it'll be optional – we had enough pains with various `xmlsec` versions...

pennersr:
> Anyway, glad it'll be optional – we had enough pains with various xmlsec versions...

@akx It would be good if you could elaborate a bit on this point. I figured that `smolsaml`, `pysaml2` and `python3-saml` all depend on `xmlsec` so whatever installation trouble that gives, I would expect it to be the same for all these packages?

akx:
@pennersr Yeah, I made the decision that smolsaml just shells out to the `xmlsec` tool for signing and verification because there have been notorious conflicts where different versions of libxml in the same python interpreter cause havoc (I'm out and about right now so I can't look up the issues) where things just don't work in a correct way.

PR #3346, opened 2023-07-12
feat(account): Add `ACCOUNT_CHANGE_EMAIL` flow

pennersr:
With these changes, when `ACCOUNT_MAX_EMAIL_ADRESSES=1`, the user can still change his email. The user is allowed to add an extra email, and when that email is confirmed, the former email is removed. So, the user is allowed to go over the maximum of 1 for the purpose of changing his email.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -15,6 +15,9 @@ indent_size = 4
 [*.py]
 indent_size = 4
 
+[*.html]
+indent_size = 4
+
 [Makefile]
 indent_style = tab
 indent_size = 8
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -16,6 +16,11 @@ Note worthy changes
 - It is now possible to manage OpenID Connect providers via the Django
   admin. Simply add a `SocialApp` for each OpenID Connect provider.
 
+- There is now a new flow for changing the email address. When enabled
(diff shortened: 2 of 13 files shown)


pennersr:
@9mido and @pktiuk -- does this solve your use cases?

e2489:
This still seems overly complicated and potentially confusing for users. Having to add an extra email in order to change your current email is an odd way to do things, and will probably not make sense to the average person.

I get that that's how it has to work on the back end, and that the end result is the same, but from the user's perspective, they should simply be able to change their email. That's the way every site has done things for decades now, and this shouldn't be tied into the 'multiple emails' feature, which offers something different, and is quite nonstandard these days.

pennersr:
@e2489 Valid remark, but isn't this solely a presentation concern? With a few changes to the templates, you turn "Add" into "Change" and instead of showing the user a manageable list of email addresses you can show the user just his current one and the one he is trying to change to, if any.

pennersr:
Changes:
- Added an explicit setting `ACCOUNT_CHANGE_EMAIL` for this, instead of hiding it behind `ACCOUNT_MAX_EMAIL_ADDRESSES=1`.
- Template tweaks -- "Change email" instead of "Add email".

pennersr:
There is now a dedicated setting for this and a separate template addressing the presentation concerns. Any other feedback?

PR #3348, opened 2023-07-14
fix(cern): Updated to new SSO w/ OIDC

nothingface0:
## General

 - [ ] ~~JavaScript code should adhere to [StandardJS](https://standardjs.com).~~ (Does not apply)
 - [ ] ~~If your change is substantial, feel free to add yourself to `AUTHORS`.~~

### CERN

CERN will be completely deprecating the current SSO solution by September 1st, 2023, making the current CERN `django-allauth` provider obsolete.

This PR updates the CERN provider to the ""new"" SSO solution, which is OIDC-based, using new URLs and response format. 
The most important thing to note is that two breaking changes have been introduced:

1. A formerly available attribute, `groups`, is no longer available. In its place, `cern_roles` have been added, and are provided by default by the JWT access token returned after logging in. Developers using this provider will have to update their applications to use `cern_roles` instead of `groups`.

	> **Note**
	> One solution could have been to just map `cern_roles` to `groups`, but, since they do not reflect the same information (i.e. multiple `groups` may be mapped to a single `cern_role`), it's better for the developers to update their applications to comply with this new concept. More information [here](https://auth.docs.cern.ch/applications/role-based-permissions).

2. The `uid` that was used previously (mapped to `id` in the old SSO response) **is no longer available.** This means that, if returning users attempt a login after `django-allauth` has been updated, there's no way to associate them with the existing `SocialAccount` entry, since the `uid` is now different. In its place, the `sub` field is now used, which is also [guaranteed to be unique](https://auth.docs.cern.ch/user-documentation/oidc/config/#what-will-be-in-your-tokens), and is already stored in the `extra_data` column of the `SocialAccount` table, so it would be easy to "migrate" the value to the `uid` column.
	I will try to come up with a "migration" script to update the existing `uid`s to be the `username` value in the `extra_data` column.
	To migrate:
	i. Enter your project's django shell:
	```bash
	python manage.py shell
	```
	ii. Use the `username` field (which has been renamed to `sub` in the new SSO) in the `extra_data` to update the `uid`:
	```python
	from allauth.socialaccount.models import SocialAccount
	sa = SocialAccount.objects.filter(provider="cern")
	for s in sa:
		if 'username' in s.extra_data:
			s.uid = s.extra_data["username"]
			s.save()
	```

 
All other functionality and setup should be intact.

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -16,6 +16,39 @@ Note worthy changes
 - It is now possible to manage OpenID Connect providers via the Django
   admin. Simply add a `SocialApp` for each OpenID Connect provider.
 
+- The CERN provider has been updated to use OIDC, in order to comply with
+  the planned deprecation of the current Single Sign-On solution. This change
+  introduces two breaking changes:
+    
+    - Egroups are no longer supported, and your
+      application must be modified to use Roles
+      through the `Application Portal <application-portal.web.cern.ch/>`_. 
+      See the `official CERN Auth docs <auth.docs.cern.ch/>`_
+      for more information.  
+    
+    - The new SSO does not provide the `id` value that was previously
+      used to uniquely map a logged in user with an existing `SocialAccount`
+      (stored in the `uid` column). We are now using the `sub` field, 
+      which is guaranteed to be unique for each user or service account. 
+      
+      This means that, once you update your `django-allauth` package, you will
+      need to update the `SocialAccount` entries too, so that all 
+      the existing CERN accounts in your database can be mapped to the users
+      which login through the new SSO. Luckily, all existing entries in your
+      database have a `username` field in the `extra_data` column of the `SocialAccount`
+      table. To update the existing entries, you will need to enter a 
+      django shell session (`python manage.py shell`)
+      and run the following: 
+
+      .. code-block:: python
+
+        from allauth.socialaccount.models import SocialAccount
+        sa = SocialAccount.objects.get(provider="cern")
+        for s in sa:
+          if 'username' in s.extra_data:
+            s.uid = s.extra_data["username"]
+            s.save()
+
 
 Security notice
 ---------------
--- a/allauth/socialaccount/providers/cern/provider.py
+++ b/allauth/socialaccount/providers/cern/provider.py
@@ -1,32 +1,40 @@
 from allauth.socialaccount.providers.base import ProviderAccount
 from allauth.socialaccount.providers.oauth2.provider import OAuth2Provider
+from allauth.socialaccount.providers.openid_connect.provider import (
+    OpenIDConnectProvider,
+    OpenIDConnectProviderAccount,
+)
 
 
-class CernAccount(ProviderAccount):
+class CernAccount(OpenIDConnectProviderAccount):
     def to_str(self):
         dflt = super(CernAccount, self).to_str()
         return self.account.extra_data.get("name", dflt)
 
 
-class CernProvider(OAuth2Provider):
+class CernProvider(OpenIDConnectProvider):
+    """
+    Child class for logging into CERN, using OIDC.
+    """
+
     id = "cern"
     name = "Cern"
+
+    # Well-known config for CERN
+    _server_url = (
+        "https://auth.cern.ch/auth/realms/cern/.well-known/openid-configuration"
+    )
     account_class = CernAccount
 
-    def get_auth_params(self, request, action):
-        data = super(CernProvider, self).get_auth_params(request, action)
-        data["scope"] = "read:user"
-        return data
-
     def extract_uid(self, data):
-        return str(data.get("id"))
+        return str(data.get("sub"))
 
     def extract_common_fields(self, data):
         return dict(
             email=data.get("email"),
-            username=data.get("username"),
-            first_name=data.get("first_name"),
-            last_name=data.get("last_name"),
+            username=data.get("preferred_username"),
+            first_name=data.get("given_name"),
+            last_name=data.get("family_name"),
             name=data.get("name"),
         )
 
--- a/allauth/socialaccount/providers/cern/tests.py
+++ b/allauth/socialaccount/providers/cern/tests.py
@@ -1,28 +1,48 @@
-from allauth.socialaccount.tests import OAuth2TestsMixin
+from allauth.socialaccount.tests import OpenIDConnectTests
 from allauth.tests import MockedResponse, TestCase
 
 from .provider import CernProvider
 
 
-class CernTests(OAuth2TestsMixin, TestCase):
+class CernTests(OpenIDConnectTests, TestCase):
     provider_id = CernProvider.id
 
     def get_mocked_response(self):
         return MockedResponse(
             200,
             """
-        {
-            "name":"Max Mustermann",
-            "username":"mmuster",
-            "id":8173921,
-            "personid":924225,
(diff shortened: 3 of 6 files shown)


nothingface0:
Converting to draft, as I've discovered another breaking feature (will update the description)

ChangeLog.rst
```diff
@@ -16,6 +16,12 @@ Note worthy changes
 - It is now possible to manage OpenID Connect providers via the Django
   admin. Simply add a `SocialApp` for each OpenID Connect provider.
 
+- The CERN provider has been updated to use OIDC, in order to comply with
+  the planned deprecation of the current Single Sign-On solution. This change
+  introduces a breaking change: egroups are no longer supported, and your
+  application must be modified to use Roles. See the official CERN Auth docs
+  for more information.  
+
```
pennersr, line 24: I think it would be good if you could add the documentation wrt to `uid` changes to the list of backwards incompatible changes in the changelog.

pennersr (changes requested):

ChangeLog.rst
nothingface0: Will do

nothingface0:
I will also have to update the tests, I'm not really sure what

```
 django.urls.exceptions.NoReverseMatch: Reverse for 'cern_callback' with keyword arguments '{'provider_id': ''}' not found. 1 pattern(s) tried: ['cern/login/callback/\\Z']
 ```
 means, yet.

**Comment by pennersr** [comment]
> I will also have to update the tests, I'm not really sure what...

On the main branch, the OpenID connect provider was considerably refactored, have a look at `urls.py`:

- After: https://github.com/pennersr/django-allauth/blob/main/allauth/socialaccount/providers/openid_connect/urls.py
- Before: https://github.com/pennersr/django-allauth/blob/66135fdcb313659d42046cedd48811a9505791a5/allauth/socialaccount/providers/openid_connect/urls.py

As you can see, before there was a bit of hacking/magic involved which is now all gone. So that affects URL reversal.

**Comment by pennersr** [comment]
I am unable to check right now, but I suspect that the current `cern/urls.py` right now is the problem:

```
from allauth.socialaccount.providers.oauth2.urls import default_urlpatterns

from .provider import CernProvider


urlpatterns = default_urlpatterns(CernProvider)
```

Here, you're using the OAuth2 URLs whereas the provider is OIDC based now. Unable to dive into this right now, but either you override `get_login_url` and `get_callback_url` on your Cern provider class, or, you make `urlpatterns = []` over at `cern/urls.py`. Then, be sure that your `app.provider_id = "cern"`. The latter approach does imply you will need to add the `openid_connect` provider to your `INSTALLED_APPS` which previously was not the case.

**Comment by pennersr** [comment]
Also, this:

`class CernTests(OAuth2TestsMixin, TestCase)`

... needs to be turned into:

```
class Cern(OpenIDConnectTests, TestCase):
    provider_id = CernProvider.id
```

**Comment by nothingface0** [comment]
> Here, you're using the OAuth2 URLs whereas the provider is OIDC based now. 

It seems that using the `openid_connect.default_urlpatterns`:
```diff
-from allauth.socialaccount.providers.oauth2.urls import default_urlpatterns
+from allauth.socialaccount.providers.openid_connect.urls import default_urlpatterns
```
along with inheriting from `OpenIDConnectTests`:
```diff
-from allauth.socialaccount.tests import OAuth2TestsMixin
+from allauth.socialaccount.tests import OpenIDConnectTests

-class CernTests(OAuth2TestsMixin, TestCase):
+class CernTests(OpenIDConnectTests, TestCase):
```
lets the url revesal tests pass. 

Does this look OK?

nothingface0:
Now there's one more test failing, `test_login_with_pkce_enabled`:

```
_____________________________________________________________________________ CernTests.test_login_with_pkce_enabled ______________________________________________________________________________

self = <allauth.socialaccount.providers.cern.tests.CernTests testMethod=test_login_with_pkce_enabled>

    @override_settings(SOCIALACCOUNT_AUTO_SIGNUP=False)
    def test_login_with_pkce_enabled(self):
        provider_settings = app_settings.PROVIDERS.get(self.app.provider, {})
        provider_settings_with_pkce_enabled = provider_settings.copy()
        provider_settings_with_pkce_enabled["OAUTH_PKCE_ENABLED"] = True
        with self.settings(
            SOCIALACCOUNT_PROVIDERS={
                self.app.provider: provider_settings_with_pkce_enabled
            }
        ):
            resp_mock = self.get_mocked_response()
            if not resp_mock:
                warnings.warn(
                    "Cannot test provider %s, no oauth mock" % self.provider.id
                )
                return
    
>           resp = self.login(
                resp_mock,
            )

allauth/socialaccount/tests/__init__.py:259: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
allauth/socialaccount/tests/__init__.py:322: in login
    self.assertEqual("code_challenge" in q, pkce_enabled)
E   AssertionError: False != True
```
I'll try to look into it. In the meantime, any insight is much appreciated! :sweat_smile:

ChangeLog.rst
nothingface0: Does this look better now? Do you think it needs any rephrasing? In my head is clear what I want to say, but I can't judge how clearly I'm explaining it.

pennersr:
> It seems that using the `openid_connect.default_urlpatterns`:

This does not make sense -- `openid_connect.urls` has no `default_urlpatterns` on main.

pennersr:
I am wondering, is having a separate `cern` provider still needed? What if you just use the `openid_connect` provider as is?

nothingface0:
> I am wondering, is having a separate `cern` provider still needed? What if you just use the `openid_connect` provider as is?

This did pass my mind at some point; I will take a look on Monday. I have to see how tedious or not the migration will be.

Thank you for your time!

nothingface0:
I've actually tested it and most of the functionality I need is there! The bad news is that some of the information is encoded into the access token, which needs to be decoded (see [here](https://github.com/pennersr/django-allauth/pull/3348/files#diff-228a341cbd07f43c91516d675600c430d803dbb5e634ca0e34c00e3675b567c1R20-R27)). Is there any idea to get it directly from `django-allauth` through the signals?

nothingface0:
In the end I figured I can update the `extra_data` of the `SocialAccount` during `pre_social_login`, and then just check if the account exists or not with `sociallogin.account.pk is None`; if it does, I save it, else, I just leave the `django-allauth` login flow to save it later.

The question remains, what should be done with the `cern` provider now? Should it be removed altogether?

pennersr:
So, the current implementation will cease to work on September? Then, for now we can leave the provider in, but indeed, it makes most sense to remove it once that date is passed. So, this PR could be turned into a PR that removes `cern` and adds a backwards incompatible note stating that people should just use the regular `openid_connect` provider instead.

pennersr:
Closing -- different approach needed (rm -rf cern)

pennersr:
@nothingface0 See: d22d95a6

nothingface0:
Awesome, thanks a lot for your time!

PR #3364, opened 2023-08-02
Settings for checking outdated data when authorizing via telegram.

Graphanton:
Added the ability to set the period for checking outdated data when authorizing via telegram.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -178,3 +178,4 @@ William Li
 Yaroslav Muravsky
 Yuri Kriachko
 Youcef Mammar
+Anton Goncharov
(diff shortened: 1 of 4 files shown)


pennersr:
> If the time of the telegram server differs from the time of your server...

Wouldn't it be better in this case to actually fix the root cause and make sure your server is running NTP so that the time is not out of sync?

Graphanton:
Yes. I removed it myself. But since Telegram does not strictly regulate the authorization expiration time, I can set my own expiration time, which can be less than the default.

pennersr:
I am okay with adding this functionality, though I think it would be good to reword the documentation. Because "If the time of the telegram server differs from the time of your server..." you need NTP and not this change.

pennersr:
Thanks for rewording. One last thing, the code is not black formatted causing the build to fail. Can you look into that?

Graphanton:
> 

Ready!

PR #3370, opened 2023-08-07
changed all templates using django crispy forms

ahn1305:
changes templates using django crispy forms


pennersr:
Providing a special flavor of templates is best served in an external package of its own. Crispy forms are not a de facto Django standard, therefore I do not think it is wise to force these templates and the accompanying crispy forms dependency onto all users of django-allauth.

ahn1305:
I personally felt customisation was a bit difficult, could you put any tutorials on that. Thankyou!

ahn1305:
Like, extending the templates, something like that

PR #3374, opened 2023-08-08
feat(socialaccount): Authenticate by email

pennersr:
Consider a scenario where a social login occurs, and the social account comes with a verified email address (verified by the account provider), but that email address is already taken by a local user account. Additionally, assume that the local user account does not have any social account connected. Now, if the provider can be fully trusted, you can argue that we should treat this scenario as a login to the existing local user account even if the local account does not already have the social account connected, because -- according to the provider -- the user logging in has ownership of the email address.  This is how this scenario is handled when `SOCIALACCOUNT_EMAIL_AUTHENTICATION` is set to `True`. As this implies that an untrustworthy provider can login to any local account by fabricating social account data, this setting defaults to `False`. Only set it to `True` if you are using providers that can be fully trusted.

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -30,6 +30,9 @@ Note worthy changes
 
 - Fixed Twitter OAuth2 authentication by using basic auth and adding scope `tweet.read`.
 
+- Added (optional) support for authentication by email for social logins (see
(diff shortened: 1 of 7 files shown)


svanellewee:
:+1:

Issue #3129, opened 2022-07-22
filter_users_by_email does not have supporting index Case-insensitive exact match match on email column.


rpoyau:
filter_users_by_email uses iexact to filter users by email and account_emailaddress table does not have an supporting
index on email column for this. 

Basically this function is doing table scan. 

Example calling:
filter_users_by_email('foo@void.net', is_active=True)
```SQL
SELECT "account_emailaddress"."id", "account_emailaddress"."user_id", "account_emailaddress"."email", "account_emailaddress"."verified", "account_emailaddress"."primary" FROM "account_emailaddress" INNER JOIN "auth_user" ON ("account_emailaddress"."user_id" = "auth_user"."id") WHERE (UPPER("account_emailaddress"."email"::text) = UPPER('foo@void.net') AND "auth_user"."is_active")
```
[query-plan.txt](https://github.com/pennersr/django-allauth/files/9167695/query-plan.txt)

Also, unfortunately auth_user (User table) also does not have an index for this as well:
```SQL
NO SCROLL CURSOR WITH HOLD FOR SELECT "auth_user"."id", "auth_user"."password", "auth_user"."last_login", "auth_user"."is_superuser", "auth_user"."username", "auth_user"."first_name", "auth_user"."last_name", "auth_user"."email", "auth_user"."is_staff", "auth_user"."is_active", "auth_user"."date_joined" FROM "auth_user" WHERE (UPPER("auth_user"."email"::text) = UPPER('foo@void.net') AND "auth_user"."is_active")
```

An index on upper(email) would resolve this problems.
create index idx_account_emailaddress_upper_email on account_emailaddress(upper(email))

pennersr:
A migration for this could be added, though would only support Django 3.2 and up: https://docs.djangoproject.com/en/3.2/releases/3.2/#functional-indexes

varunsaral:
Hey @pennersr, i have added the migration and waiting for your review. As told by you the tox test is failing for version <= 3.2.

pennersr:
Support for 3.1 has just been dropped on the main branch, so that will not be an issue.

varunsaral:
We can close this @pennersr , sorry about the blunder, first i made the index as "idx_account_emailaddress_upper_email" but it had 30 char limit that's why i shortend it. But while making changes after your review i forgot to double check. Maybe we can have a test to see if migration are getting applied without any problem.

pennersr:
@varunsaral No worries... Thanks for moving things forward.

[...]

PR #3382, opened 2023-08-23
Add Upper index for email in EmailAddress model

varunsaral:
This PR add index for email field in EmailAddress Model, this is to improve email__iexact filter in function filter_users_by_email


varunsaral:
This PR fixes #3129

allauth/account/models.py
```diff
@@ -41,6 +43,7 @@ class Meta:
                     condition=Q(verified=True),
                 )
             ]
+        indexes = [Index(Upper("email"), name="idx_upper_email")]
```
pennersr, line 46: The other indexes are named like this:

```
Indexes:
    "account_emailaddress_pkey" PRIMARY KEY, btree (id)
```

Could you rename this one to `account_emailaddress_upper_email` ?

varunsaral: I have renamed the index and manually renamed in migration file, i hope that's the correct way instead of deleting and regenerating the migration file.

pennersr:
One final thing, can you make sure the code is black formatted? The build is failing on that now. Thanks.

varunsaral:
I did the black formatting, sorry for the inconvenience.

varunsaral:
applied isort as suggested by tests

PR #3383, opened 2023-08-23
Multi-Factor Authentication

pennersr:
Introduces `allauth.mfa` -- adding support for MFA.


ryanhiebert:
This is very exciting, and I expect that I'll be trying to implement it. I'll note one thing for your consideration: we would like to be able to require 2FA, except when the user logs in via a SAML SSO integration, so if it were convenient to make that possible via a hook or signal or something, I know that would be helpful for what we'd like to be able to do.

pennersr:
Also, wondering, wouldn't it be better to require 2FA over at the SAML Identity Provider instead, instead of within your own Django project?

akx:
_sad django-allauth-2fa maintainer noises_

Issue #3297, opened 2023-04-03
Cache stores CSRF on login and register


mfvictor:
LoginView and SignupView are caching the csrf token when back is pressed after Login/Signup. 

https://user-images.githubusercontent.com/8157518/229463400-b1e5b320-c240-4388-b7ec-1d6a8fa292a4.mp4

This can be reproduced in the example app.

varunsaral:
@mfvictor if you are not working on this, can i take up the work?

varunsaral:
@pennersr can i take up the work for this?

pennersr:
Definitely...

pennersr:
Closed via #3390

pennersr:
https://codeberg.org/allauth/django-allauth/issues/3297

PR #3390, opened 2023-08-25
Add never_cache decorator to prevent csrf_token caching

varunsaral:
This pr solves #3297


varunsaral:
applied isort as suggested by tests

Issue #3396, opened 2023-08-28
Keycloak provider broken in version 0.55


antoinehumbert:
In version 0.55 of django-allauth, the keycloak provider is broken because the callback url is hardcoded by OpenIDConnectAdapter at https://github.com/pennersr/django-allauth/blob/0.55.0/allauth/socialaccount/providers/openid_connect/views.py#L60, using "openid_connect_callback" URL name instead of "keycloak_callback").

The get_callback_url method should be redefined on class KeycloakOAuth2Adapter https://github.com/pennersr/django-allauth/blob/0.55.0/allauth/socialaccount/providers/keycloak/views.py#L13

pennersr:
Ouch, can you prep a PR for this? :pray:

antoinehumbert:
I'll do.

Instead of redefining get_callback_url in KeycloakOAuth2Adapter with the same implementation as in the base class OAuth2Provider, I'm thinking about:
- create a BaseOpenIDConnectAdapter implementing methods of actual OpenIDConnectAdapter except constructor (so without provider_id set on the class nor initiated on instances by constructor) and the current get_callback_url
- make OpenIDConnectAdapter inherit this base class and redifining the constructor (to set provider_id) and get_callback_url
- make KeycloakOAuth2Adapter inherit this base class

This way, we could also modify implementation of keycloak login and callback views by providing the KeycloakOAuth2Adapter class to .adapter_view() instead of an instance as the latter would not be required anymore

What do think about that ?

antoinehumbert:
Hi @pennersr, will you create a branch for version 0.55 hotfixes I should start from for this fix, or would you like this PR to occur on the main branch and corresponding commit will ve cherrypicked for creation of a 0.55.1 version ?

antoinehumbert:
The main reason is because when not using `allauth/urls.py` in a project (that's my case), but only `allauth/socialaccount/providers/keycloak/urls.py` (because we only want to allow keycloak authentication and do not want signin, signup views - we do auto signup - ...), then the "openid_connect_callback" URL cannot be reversed as it's not defined in project's urls.

The second reason is, as you said, because it is not pretty :
- we, indeed, have two URL patterns in defined urls that match the same path, and it works because openid_connect URLs were explicitly moved at the end of url patterns, but from a semantic point of view I'd rather reverse `keycloak_callback` when I want to work with keycloak rather than reversing `openid_connect_callabck` (even if the result is the same)
[...]

PR #3398, opened 2023-08-29
fix(keycloak): use keycloak_callback reversed URL for Keycloak provider instead of openid_connect_callback reversed URL.

antoinehumbert:
Issue #3396 

Fix retrieval of callback URL for Keycloak provider which was computed by reversing `openid_connect_callback` instead
  of `keycloak_callback`. Even if resulting URL would be the same when all `allauth.urls` are installed, this could
  result in a `NoReverseMatch` error when OpenIDConnect provider's URLs are not included in project's URLs.

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -4,7 +4,11 @@
 Note worthy changes
 -------------------
 
-- ...
+- Fix retrieval of callback URL for Keycloak provider which was computed by
+  reversing `openid_connect_callback` instead of `keycloak_callback`. Even if
+  both reversed URL would be the same, this resulted in a `NoReverseMatch`
+  error when `allauth.socialaccount.providers.openid_connect` was not present
+  in `INSTALLED_APPS`.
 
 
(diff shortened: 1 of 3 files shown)


pennersr:
This would break the build:

```
black --check allauth setup.py
would reformat /home/pennersr/src/django-allauth/allauth/socialaccount/providers/openid_connect/views.py

Oh no! 💥 💔 💥
1 file would be reformatted, 737 files would be left unchanged.
```

antoinehumbert:
> 

Sorry, I will correct

antoinehumbert:
> This would break the build:
> 
> ```
> black --check allauth setup.py
> would reformat /home/pennersr/src/django-allauth/allauth/socialaccount/providers/openid_connect/views.py
> 
> Oh no! 💥 💔 💥
> 1 file would be reformatted, 737 files would be left unchanged.
> ```

Corrected

pennersr:
Merged via f7de05e0

PR #3410, opened 2023-09-03
feat(accounts): Added prefilling of forms with GET query strings

### LINKED ISSUE #2861: Feature Request: Prefill option for E-Mail on signup form

varunsaral:
This pr adds email prefilling suport in signup form, closes #2861

### DIFF
--- a/docs/socialaccount/providers/openid.rst
+++ b/docs/socialaccount/providers/openid.rst
@@ -12,15 +12,21 @@ builtin templates can be configured as follows:
     SOCIALACCOUNT_PROVIDERS = {
         'openid': {
             'SERVERS': [
-                dict(id='yahoo',
-                     name='Yahoo',
-                     openid_url='http://me.yahoo.com'),
-                dict(id='hyves',
-                     name='Hyves',
-                     openid_url='http://hyves.nl'),
-                dict(id='google',
-                     name='Google',
-                     openid_url='https://www.google.com/accounts/o8/id'),
+                {
+                    'id': 'yahoo',
+                    'name': 'Yahoo',
+                    'openid_url': 'http://me.yahoo.com',
+                },
+                {
+                    'id': 'hyves',
+                    'name': 'Hyves',
+                    'openid_url': 'http://hyves.nl',
+                },
+                {
+                    'id': 'google',
+                    'name': 'Google',
+                    'openid_url': 'https://www.google.com/accounts/o8/id',
+                },
             ]
         }
     }
@@ -30,13 +36,19 @@ You can manually specify extra_data you want to request from server as follows::
     SOCIALACCOUNT_PROVIDERS = \
         { 'openid':
             { 'SERVERS':
-                 [ dict(id='mojeid',
-                      name='MojeId',
-                      openid_url='https://mojeid.cz/endpoint/',
-                      extra_attributes = [
-                          ('phone', 'http://axschema.org/contact/phone/default', False),
-                          ('birth_date', 'http://axschema.org/birthDate', False,),
-                      ])]}}
+                [
+                    {
+                        'id': 'mojeid',
+                        'name': 'MojeId',
+                        'openid_url': 'https://mojeid.cz/endpoint/',
+                        'extra_attributes': [
+                            ('phone', ''http://axschema.org/contact/phone/default', False),
+                            ('birth_date', 'http://axschema.org/birthDate', False),
+                        ]
+                    },
+                ]
+            }
+        }
 
 Attributes are in form (id, name, required) where id is key in extra_data field of socialaccount,
 name is identifier of requested attribute and required specifies whether attribute is required.
@@ -54,8 +66,13 @@ The OpenID provider can be forced to operate in stateless mode as follows::
     SOCIALACCOUNT_PROVIDERS = \
         { 'openid':
             { 'SERVERS':
-                [ dict(id='steam',
-                    name='Steam',
-                    openid_url='https://steamcommunity.com/openid',
-                    stateless=True,
-                )]}}
+                [
+                    {
+                        'id': 'steam',
+                        'name': 'Steam',
+                        'openid_url': 'https://steamcommunity.com/openid',
+                        'stateless': True,
+                    },
+                ]
+            }
+        }


allauth/account/views.py
```diff
@@ -290,6 +290,13 @@ def get_context_data(self, **kwargs):
         )
         return ret
 
+    def get_initial(self):
+        if self.request.GET.get("email", None):
+            initial = super().get_initial()
+            initial["email"] = self.request.GET.get("email")
```
pennersr, line 296: In case of `app_settings.SIGNUP_EMAIL_ENTER_TWICE` there is also an `email2` field. That would need to be prefilled as well.

allauth/account/views.py
```diff
@@ -290,6 +290,13 @@ def get_context_data(self, **kwargs):
         )
         return ret
 
+    def get_initial(self):
+        if self.request.GET.get("email", None):
```
pennersr, line 294:
```suggestion
        if "email" in self.request.GET:
```

allauth/account/views.py
pennersr, line 296: I think we need to validate (as in https://docs.djangoproject.com/en/4.2/ref/validators/#emailvalidator) the email to avoid prefilling the form with garbage data?

pennersr (changes requested):

allauth/account/views.py
varunsaral: Do we also need to respect EMAIL_MAX_LENGTH?

pennersr: No, the prefilled data will be validated for that anyway, but preventing prefilling to appear when you do `email=no-email-just-spam` would be nice.

allauth/account/views.py
```diff
@@ -290,6 +290,13 @@ def get_context_data(self, **kwargs):
...
+            initial = super().get_initial()
+            initial["email"] = self.request.GET.get("email")
+            return initial
+        return super().get_initial()
```
pennersr, line 298: Would also be nice to make this a bit more DRY. You're invoking super().get_initial twice. How about:

```
initial = super().get_initial():
if ...:
  prefill
return initial
```

allauth/account/views.py
varunsaral: Should we restrict if user is prefilling only one field, ex: only email or email2. Or is it fine to prefill anyone or both?

allauth/account/views.py
varunsaral: noted

allauth/account/views.py
pennersr: It is fine to prefill both.

allauth/account/views.py
```diff
@@ -291,11 +293,23 @@ def get_context_data(self, **kwargs):
...
-        if self.request.GET.get("email", None):
-            initial = super().get_initial()
-            initial["email"] = self.request.GET.get("email")
-            return initial
-        return super().get_initial()
+        intial = super(SignupView, self).get_initial()
+        twice_email = app_settings.SIGNUP_EMAIL_ENTER_TWICE
+        request_get = self.request.GET
+
+        def validate_and_set_email(email_key):
```
pennersr, line 300: This looks a little bit complex to me, e.g. there is no need to validate the email twice. How about simplifying into:

```
from django.core.validators import validate_email

def get_initial(self):
     intial = super().get_initial()
     email = self.request.GET("email")
     if email:
         try:
             email = validate_email(email)
         except ValidationError:
             pass
         else:
             intial["email"] = email
             if app_settings.SIGNUP_EMAIL_ENTER_TWICE:
                 intial["email2"] = email
     return intial
```

varunsaral:
I think this should be fine @pennersr, don't know why black,isort are failing i applied them before pushing

allauth/account/views.py
```diff
@@ -46,7 +46,8 @@
     sync_user_email_addresses,
     url_str_to_user_pk,
 )
-
+from django.core.validators import EmailValidator
+from django.forms import ValidationError
```
fdobrovolny, line 50: The isort is failing on these two lines if you take a look at the logs:

```python
--- /home/runner/work/django-allauth/django-allauth/allauth/account/views.py:before	2023-09-03 10:56:32.162313
+++ /home/runner/work/django-allauth/django-allauth/allauth/account/views.py:after	2023-09-03 10:56:51.994464
@@ -2,6 +2,8 @@
 from django.contrib.auth import REDIRECT_FIELD_NAME
 from django.contrib.auth.decorators import login_required
 from django.contrib.sites.shortcuts import get_current_site
+from django.core.validators import EmailValidator
+from django.forms import ValidationError
 from django.http import (
     Http404,
     HttpResponsePermanentRedirect,
@@ -46,8 +48,7 @@
     sync_user_email_addresses,
     url_str_to_user_pk,
 )
-from django.core.validators import EmailValidator
-from django.forms import ValidationError
+
```

There should be two lines between the first code line and imports. You removed one.

 The imports should be moved to be sorted alphabetically, as can be seen in the patch by isort.

allauth/account/views.py
```diff
@@ -290,6 +291,19 @@ def get_context_data(self, **kwargs):
         )
         return ret
 
+    def get_initial(self):
+        intial = super(SignupView,self).get_initial()
```
fdobrovolny, line 295:
```suggestion
        intial = super(SignupView, self).get_initial()
```

To make black happy you need a space here.

fdobrovolny (commented):
Not sure why your black and isort, are not working. But the following changes seem like the reason the tests are failing.

varunsaral:
@pennersr this can be closed. Everything changed according to the suggestions

varunsaral:
Along with #3003

allauth/account/views.py
varunsaral: Thanks @fdobrovolny for the suggestion, turns out the isort and black installation were messed up.

allauth/account/tests/test_signup.py
```diff
@@ -315,3 +315,17 @@ def test_django_password_validation(self):
                 "password1",
                 ["This password is too short. It must contain at least 9 characters."],
             )
+
+    def test_get_initial_with_valid_email(self):
```
pennersr, line 319: The codebase is transitioning towards using `pytest` based tests. Can you turn this into a regular (pytest) function? As in, not make it a mapper of a class, let it accept `rf` (request factory) as a parameter, and so on...

allauth/account/views.py
```diff
@@ -290,6 +292,19 @@ def get_context_data(self, **kwargs):
         )
         return ret
 
+    def get_initial(self):
+        intial = super(SignupView, self).get_initial()
```
pennersr, line 296: Typ-o: in-i-tial

allauth/account/views.py
```diff
@@ -290,6 +292,19 @@ def get_context_data(self, **kwargs):
...
+        email = self.request.GET.get("email")
+        if email:
+            try:
+                EmailValidator()(email)
```
pennersr, line 300: There is no need to spawn a validator yourself, just call this method:

https://docs.djangoproject.com/en/4.2/ref/validators/#validate-email

pennersr (changes requested):

allauth/account/views.py
pennersr, line 296: Just use `super()` -- no need to be Python 2 compliant anymore.

varunsaral: I think the spelling is correct, it's in-i-tial

pennersr: But it says `intial = ...`

varunsaral:
@pennersr Mistakenly rebased wrong and now whole branch got deleted 😭 , sorry about the trouble. I have redone all the changes and raised the pr #3447. I have checked it throughly and changed everything with your review suggestions.

PR #3412, opened 2023-09-03
refactor: Provide request context

pennersr:
The availability of `request` throughout the code is not consistent. Some methods get `request` passed explicitly, some don't, and some constructs have it available both as a parameter and as `self.request`. Having `request` available is essential, especially when trying to implement adapter hooks, though littering the code base with an explicit `request` is not very great.

With the changes made here, there is always a request available if you need it: `allauth.core.context.request`.

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -6,6 +6,16 @@ Note worthy changes
 
 - Added builtin support for Two-Factor Authentication via the ``allauth.mfa`` app.
 
+- The fact that ``request`` is not available globally has left its mark on the
+  code over the years. Some functions get explicitly passed a request, some do
+  not, and some constructs have it available both as a parameter and as
+  ``self.request``.  As having request available is essential, especially when
+  trying to implement adapter hooks, the request has now been made globally
+  available via::
+
+    from allauth.core import context
+    context.request
+
 
 Backwards incompatible changes
 ------------------------------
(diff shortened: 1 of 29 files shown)


akx:
I'm not sure this is an excellent idea – it kind of goes against the Zen of Python's "Explicit Is Better Than Implicit"...

akx:
```python
get_adapter(request).get_login_redirect_url(request)
```
sounds like it should be just
```python
get_adapter(request).get_login_redirect_url()
```
which keeps it clear that the adapter is per-request. 🤷

pennersr:
@akx Why then is the signature of Django's `View.get()` method like this:

```    def get(self, request, *args, **kwargs):```

It has the request both available via the parameter `request` and member `self.request`.  Isn't it clear that views are per request?

Well, you could argue that in order to implement `get` you will likely need access to the `request`, and therefore explicitly pass it along.  But, coming back to our `get_login_redirect_url()`, that is exactly what people are doing here as well -- inspect the URL or session, and redirect somewhere. So, per that reasoning, `request` would need to be an explicit parameter.

And then you get into a fuzzy calculation of the likelihood of a method needing the request, and if the likelihood is large enough it needs to be passed explicitly. And that fuzziness results in inconsistent signatures and tickets like #3235.

akx:
I agree that it's a bit weird that Django's CBVs have both `self.request` and `request`, but they _do_ have `self.request`, not a global contextvar.

PR #3428, opened 2023-09-12
password_change_done

amirreza8002:
i added a page so when you change password it redirects to that instead of staying in the password change view

it's my first time doing this so go easy on me :)

pls let me know what is missing


pennersr:
Currently, when you change your password you indeed stay on the page, and from the user's point of view that looks like this:

image

With the changes made here, you end up on a separate page. This raises several questions:

- The page content will state "Password` successfully changed.", yet, that exact content is already in posted as a message. So, if you are aiming for something like this, shouldn't the message (as in the `django.contrib.messages` one) be dropped?

- Given that the success page is just a GET request away on its own URL, you can easily scare someone by navigating to that page while they are fetching coffee. 

So I have some doubts on whether or not this is truly an improvement over the existing situation. Now, I definitely understand why people would want to navigate away from the password change form, so perhaps that should be made easier by means of adding some form of `get_password_change_redirect_url()` to the adapter. Then, overriding the default behavior is easier, and people could redirect to e.g. the main dashboard/profile page or something like that.

pennersr:
See 088a4f1a -- that introduces a `get_password_change_redirect_url()` adapter method. Opting for this so that projects are free to do whatever they see fit.

PR #3441, opened 2023-09-21
Fix LOGOUT_ON_PASSWORD_CHANGE settings not tested

SebCorbin:

Let's test this and find out that it is bugged, then I'll ad a new commit to fix (TDD style)

### DIFF
--- a/allauth/account/tests/test_reset_password.py
+++ b/allauth/account/tests/test_reset_password.py
@@ -309,4 +309,4 @@ def test_password_change_ACCOUNT_LOGOUT_ON_PASSWORD_CHANGE(settings, auth_client
         "password2": "newpass123"
     }
     resp = auth_client.post(reverse("account_change_password"), data)
-    assert resp['location'] == settings.ACCOUNT_LOGOUT_REDIRECT_URL
+    assert resp['location'] == settings.LOGOUT_REDIRECT_URL
--- a/allauth/account/tests/test_reset_password.py
+++ b/allauth/account/tests/test_reset_password.py
@@ -309,4 +309,4 @@ def test_password_change_ACCOUNT_LOGOUT_ON_PASSWORD_CHANGE(settings, auth_client
         "password2": "newpass123"
     }
     resp = auth_client.post(reverse("account_change_password"), data)
-    assert resp['location'] == reverse("account_login")
+    assert resp['location'] == settings.ACCOUNT_LOGOUT_REDIRECT_URL
--- a/allauth/account/tests/test_reset_password.py
+++ b/allauth/account/tests/test_reset_password.py
@@ -280,6 +280,17 @@ class ResetPasswordTests(TestCase):
         # EmailVerificationMethod.MANDATORY sends us to the confirm-email page
         self.assertRedirects(resp, "/confirm-email/")
 
+    @override_settings(ACCOUNT_LOGOUT_ON_PASSWORD_CHANGE=True)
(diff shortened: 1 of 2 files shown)


allauth/account/tests/test_reset_password.py
```diff
@@ -280,6 +280,17 @@ def test_password_reset_ACCOUNT_LOGIN_ON_PASSWORD_RESET(self):
         # EmailVerificationMethod.MANDATORY sends us to the confirm-email page
         self.assertRedirects(resp, "/confirm-email/")
 
+    @override_settings(ACCOUNT_LOGOUT_ON_PASSWORD_CHANGE=True)
+    def test_password_change_ACCOUNT_LOGOUT_ON_PASSWORD_CHANGE(self):
```
pennersr, line 284: We're in the process of moving to pytest, and therefore adding new old style tests is strongly discouraged. Can you turn this into a pytest? Just to get you started:

``` python
def test_passowrd_change_account_logout_on_password_change(settings, auth_client):
    settings.ACCOUNT_LOGOUT_ON_PASSWORD_CHANGE = True
    resp = auth_client.post(...)
    assert resp['location'] == ...

File: allauth/account/tests/test_reset_password.py [hunk 2]
```diff
@@ -300,3 +300,13 @@ def _create_user_and_login(self, usable_password=True):
...
+
+def test_password_change_ACCOUNT_LOGOUT_ON_PASSWORD_CHANGE(settings, auth_client, user_password):
+    settings.ACCOUNT_LOGOUT_ON_PASSWORD_CHANGE = True
+    data = {
+        "oldpassword": user_password, 
+        "password1": "newpass123", 
+        "password2": "newpass123"
+    }
+    resp = auth_client.post(reverse("account_change_password"), data)
+    assert resp['location'] == reverse("account_login")
```
pennersr, line 312: The build is failing on this point. Shouldn't you assert on `== settings.ACCOUNT_LOGOUT_REDIRECT_URL` ?

SebCorbin: Exactly yes, I mixed my project's settings for the defaults

allauth/account/views.py
```diff
@@ -628,13 +628,7 @@ def dispatch(self, request, *args, **kwargs):
...
 
     def render_to_response(self, context, **response_kwargs):
-        if self.request.user.is_anonymous:
-            # We end up here when `ACCOUNT_LOGOUT_ON_PASSWORD_CHANGE = True`.
-            redirect_url = get_adapter(self.request).get_logout_redirect_url(
-                self.request
-            )
-            return HttpResponseRedirect(redirect_url)
-        elif not self.request.user.has_usable_password():
+        if not self.request.user.has_usable_password():
```
pennersr, line 631: Why did you alter this code? AFAIK, Django's `AnonymousUser` does not have a `has_usable_password()` method.

SebCorbin: AFAIK the comment _We end up here when `ACCOUNT_LOGOUT_ON_PASSWORD_CHANGE = True`_ is not valid (this is why I'm adding a test), which means : 
 - if we end up on this view as anonymous, we should be redirected somewhere, LOGIN_URL maybe ?
 - we're connected and if we submit this view and successfully change the password:
   - we have LOGOUT_ON_PASSWORD_CHANGE to True and should be logged out
   - LOGOUT_ON_PASSWORD_CHANGE is False and we just redirect to success_url

pennersr:
Refactored via 088a4f1a

PR #3448, opened 2023-09-23
feat(socialaccount): Add new socialaccount signal after user and socialaccount save

Frohus:
The post_user_and_socialaccount_save signal is triggered after the user and the user's socialaccount are saved to the db.

I found myself in need to create/update user's related model using data from `socialaccount.extra_data` after user logs in with a social account but I couldn't use any of the existing signals, because:
- `allauth.socialaccount.signals.pre_social_login` is triggered before the user and the user's socialaccount are saved to the DB, so can't create a related model instance that has ForeignKey / OneToOne relation to the user.
- `allauth.socialaccount.signals.social_account_added` is only triggered after connection to an existing account,
- `allauth.socialaccount.signals.social_account_updated` in my case is only triggered after the user logs out and logs back in because of the settings I use:
```python
SOCIALACCOUNT_EMAIL_AUTHENTICATION = True
SOCIALACCOUNT_EMAIL_AUTHENTICATION_AUTO_CONNECT = True
SOCIALACCOUNT_EMAIL_VERIFICATION = "none"
```
\+ `"VERIFIED_EMAIL": True,` for every provider.

### DIFF
--- a/allauth/socialaccount/adapter.py
+++ b/allauth/socialaccount/adapter.py
@@ -21,7 +21,7 @@ from ..utils import (
     serialize_instance,
     valid_email_or_none,
 )
-from . import app_settings
+from . import app_settings, signals
 
 
 class DefaultSocialAccountAdapter(object):
@@ -88,6 +88,7 @@ class DefaultSocialAccountAdapter(object):
         else:
             get_account_adapter().populate_username(request, u)
         sociallogin.save(request)
+        signals.post_user_and_socialaccount_save.send(self.__class__, sociallogin=sociallogin)
         return u
 
     def populate_user(self, request, sociallogin, data):
--- a/allauth/socialaccount/adapter.py
+++ b/allauth/socialaccount/adapter.py
@@ -21,7 +21,7 @@ from ..utils import (
     serialize_instance,
     valid_email_or_none,
 )
-from . import app_settings, signals
+from . import app_settings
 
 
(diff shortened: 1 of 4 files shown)


pennersr:
I am not sure it makes sense to add an out of the box signal for this, given that there are already existing options:

- Add a custom adapter to your project and do this:

``` python
    def save_user(self, request, sociallogin, form=None):
        ret = super().save_user(reuqest, sociallogin, form=form)
        # ... INSERT YOUR OWN LOGIC HERE...
        return ret
```

- Use Django's out of the box `post_save` signal (https://docs.djangoproject.com/en/4.2/ref/signals/#django.db.models.signals.post_save)

The latter makes perhaps most sense for your use case?

Frohus:
Custom adapter with overridden `save_user()` is I how handled it initially but thought signals would be a better approach since there are other signals already triggered at different stages of the social login lifecycle.
It can be useful for other people with similar setup to mine as well.

I don't think Django's `post_save()` would be a good idea here since my logic relies on 2 different models (User and SocialAccount) being saved to the DB.

Frohus:
I've updated PR: moved the signal send() to SocialLogin.save(). I think it makes more sense to be there.

pennersr:
Given that it is already possible to implement this, and given that this signal would really be quite specific, I don't think it is needed to offer something for this out of the box.

Issue #3453, opened 2023-09-27
account.EmailAddress: (models.W043) MariaDB does not support indexes on expressions.


vaughanchandler:
After upgrading from 0.54.0 to 0.57.0 I'm getting these warnings:

> account.EmailAddress: (models.W036) MariaDB does not support unique constraints with conditions.
> 	HINT: A constraint won't be created. Silence this warning if you don't care about it.
> account.EmailAddress: (models.W043) MariaDB does not support indexes on expressions.
> 	HINT: An index won't be created. Silence this warning if you don't care about it.

W036 is covered by #3385 but W043 doesn't seem to have been reported as yet.

JamesParrott:
This seems like it could be related to the additions of these lines 3 months ago:
```
class EmailAddress(models.Model):
...
    class Meta:
    ...
        if app_settings.UNIQUE_EMAIL:
            constraints = [
                UniqueConstraint(
                    fields=["email"],
                    name="unique_verified_email",
                    condition=Q(verified=True),
                )
            ]
        indexes = [Index(Upper("email"), name="account_emailaddress_upper")]
```
https://github.com/pennersr/django-allauth/blob/701ef63cfd4508af0faf0671ac89c87e0d52a801/allauth/account/models.py#L37

pennersr:
The index was added here:

https://github.com/pennersr/django-allauth/pull/3382

... which was needed to solve performance issues here:

https://github.com/pennersr/django-allauth/issues/3129

Now, if your database does not support this then compared to previous versions the behavior is the same, and you can safely ignore that warning. Other than storing all email addresses as lower case -- which is according to standards not the right thing to do (https://github.com/pennersr/django-allauth/issues/1873#issuecomment-1740703742) -- there is nothing that can be done here besides silencing the warning and or upgrading your database.

Closing this ticket -- it is not actionable. Storing as lower case can be discussed over at #1873.

pennersr:
https://codeberg.org/allauth/django-allauth/issues/3453

PR #3460, opened 2023-10-01
feat(account): Store emails lower case, treat case insensitive

pennersr:
Formally, email addresses are case sensitive because the local part (the part before the "@") can be a case sensitive user name.  To deal with this, workarounds have been in place for a long time that store email addresses in their original case, while performing lookups in a case insensitive style. This approach led to subtle bugs in upstream code, and also comes at a performance cost (``__iexact`` lookups). The latter requires case insensitive index support, which not all databases support. Re-evaluating the approach in current times has led to the conclusion that the benefits do not outweigh the costs.  Therefore, email addresses are now always stored as lower case, and migrations are in place to address existing records.

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -4,6 +4,12 @@
 
 - New providers: TikTok, Lichess.
 
+- Starting since version 0.62.0, new email addresses are always stored as lower
+  case. In this version, we take the final step and also convert existing data
+  to lower case, alter the database indices and perform lookups
+  accordingly. Migrations are in place.  For rationale, see the note about email
+  case sensitivity in the documentation.
+
 
 0.62.1 (2024-04-24)
 *******************
--- a/allauth/account/managers.py
+++ b/allauth/account/managers.py
@@ -1,4 +1,3 @@
-import functools
 from datetime import timedelta
 
 from django.db import models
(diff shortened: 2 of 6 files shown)


derek-adair:
Worth looking at https://docs.djangoproject.com/en/4.1/releases/4.0/#functional-unique-constraints?

pennersr:
@derek-adair MariaDB/MySQL don't support those: https://github.com/pennersr/django-allauth/issues/3453

pennersr:
Even though what is mentioned over at https://github.com/pennersr/django-allauth/pull/3460#issue-1920726894 is correct, I am quite reluctant to break with RFC5321. Also, one might run into migration errors when rolling this out. TBD how to proceed.

pennersr:
Related: https://ellen.dev/til-firebase-auth-lowercases-email-addresses.html

pennersr:
Interesting: https://www.reddit.com/r/todayilearned/comments/2r264r/comment/cnbuk7w/

pennersr:
Let's break this down in smaller steps. First, 80088ce3 lands which stores email addresses in lower case without altering the database index or lookups. The remainder -- this PR -- will land at a later time.

PR #3467, opened 2023-10-06
feat(socialaccounts):  Check in User.email for account matching email address

CarlSchwan:
# Submitting Pull Requests

## General

Currently when EMAIL_AUTHENTIFICATION is True, we only checks for other socialaccount accounts and not for accounts in User. The usecase is when migrating the authentification system to one using User.email to store emails to allauth which use EmailAddress.emails

This probably should be behind a configuration flag, I'm not an Django expert so maybe there is better way to do this.

### DIFF
--- a/allauth/socialaccount/models.py
+++ b/allauth/socialaccount/models.py
@@ -338,6 +338,14 @@ class SocialLogin(object):
                 self.connect(context.request, address.user)
             else:
                 self.user = address.user
+            return
+
+        user = get_user_model().objects.filter(email__in=emails).first()
+        if user:
+            if app_settings.EMAIL_AUTHENTICATION_AUTO_CONNECT:
+                self.connect(context.request, user)
+            else:
+                self.user = user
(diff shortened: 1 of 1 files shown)


allauth/socialaccount/models.py
```diff
@@ -338,6 +338,14 @@ def _lookup_by_email(self):
                 self.connect(context.request, address.user)
             else:
                 self.user = address.user
+            return
+
+        user = get_user_model().objects.filter(email__in=emails).first()
```
pennersr, line 343: Not all user models have an `email` field. Also, email casing needs to be taken into account, so this should likely be replaced by `filter_users_by_email()`. I think that the existing `EmailAddress.objects.lookup(emails).order_by("-verified", "user_id").first()` line could be replaced by it, so no need to duplicate the `if` that follows here.

pennersr (changes requested):

PR #3483, opened 2023-10-14
feat(ebay): Starting work on eBay OAuth Provider

kareemassad:
# Submitting Pull Requests

## General

 - [ ] Make sure you use [semantic commit messages](https://seesparkbox.com/foundry/semantic_commit_messages).

 ## Provider Specifics

- [x] Add documentation to `docs/providers/<provider name>.rst` and `docs/providers/index.rst` Provider Specifics toctree.


allauth/socialaccount/providers/ebay/provider.py
```diff
@@ -0,0 +1,20 @@
...
+    pass
+
+
+class EBayProvider(OAuth2Provider):
+    id = "ebay"
+    name = "EBay"
+    account_class = EBayAccount
+
+    def extract_uid(self, data):
+        # NOTE: Ensure that the user_id exists
```
pennersr, line 16: This comment is a bit unclear. It refers to `user_id` while the code uses `userid`. Also, what is meant by this comment? Who is to ensure this?

allauth/socialaccount/providers/ebay/views.py
```diff
@@ -0,0 +1,65 @@
...
+    "sandbox": {
+        "auth_url": "https://auth.sandbox.ebay.com/oauth2/authorize",
+        "token_url": "https://api.sandbox.ebay.com/identity/v1/oauth2/token",
+    },
+}
+
+ENV = (
+    getattr(settings, "SOCIALACCOUNT_PROVIDERS", {})
+    .get("ebay", {})
+    .get("ENVIRONMENT", "production")
```
pennersr, line 28: Whether the production or sandbox environment is used is actually a setting that we can store/tweak per app. I think it would be best to retrieve it from the `SocialApp.settings`:

``` python
  SOCIALACCOUNT_PROVIDERS = {
      "ebay": {
          "APPS": [
              {
                  "client_id": "<insert-id>",
                  "secret": "<insert-secret>",
                  "settings": {
                      "environment": "production",
                  }
              }
          ]
      }
  }
```

(note: the lower case "environment" is intentional here)

pennersr (changes requested):
Thanks for contributing. I added a few comments -- can you look into these?

allauth/socialaccount/providers/ebay/provider.py
kareemassad: I honestly was not to sure if I was meant to extract user_id or userid. That's why that comment is there. I was following Dwolla and Dropbox implementation as an example.

pennersr: Looking at their docs over here https://developer.ebay.com/api-docs/commerce/identity/overview.html they only seem to be offering `userId` ?

allauth/socialaccount/providers/ebay/views.py
kareemassad: Sounds great! Where should this change go? I can't seem to find the file. Should it be within the test_settings.py? Also is there a Contributing document I should have a look at or some way to test it? I'm struggling to know if I am on the right track.

allauth/socialaccount/providers/ebay/provider.py
kareemassad: That looks right, will be updated next push

kareemassad:
Hello,

EBay uses a two part authentication system. First, we need to get the user's consent (and specify scopes of the application), then we place the authorization code grant request to get a User access token and a refresh token. (I am following the docs here: https://developer.ebay.com/api-docs/static/oauth-consent-request.html, https://developer.ebay.com/api-docs/static/oauth-auth-code-grant-request.html.

Is it possible to support this? We would need to have a way to pass scopes. Would that be part of settings aswell? 

Sorry for the questions!

Thanks!

allauth/socialaccount/providers/ebay/views.py
pennersr: In your adapter class, you can do:

```
app = self.get_provider().app
environment = app.settings.get("environment", "production")
```

So, this is a change to your adapter class, not related to `test_settings.py`.

pennersr:
> Is it possible to support this? We would need to have a way to pass scopes. Would that be part of settings aswell?

What you describe is regular OAuth2 functionality -- the other providers do similar things. I suggest you have a look at the `OAuth2Provider` class, specifically the `get_scope()` method. As you can see there, there is already a uniform way in which providers get ahold of the scopes, so this is not something you need a custom implementation for.

kareemassad:
I implemented some of the changes discussed. I wrote some tests (probably more than necessary), but I am not too sure how to test it

pennersr:
Abandoned, closing.

PR #3484, opened 2023-10-15
fix(openstreetmap): Update provider url

danieldegroot2:
### Switch to api.openstreetmap.org API host

Use `api.openstreetmap.org/api/` -and HTTPS- instead of `www.openstreetmap.org/api/*`.

(Is: https://github.com/openstreetmap/operations/issues/951)

### DIFF
--- a/allauth/socialaccount/providers/openstreetmap/views.py
+++ b/allauth/socialaccount/providers/openstreetmap/views.py
@@ -12,7 +12,7 @@ from .provider import OpenStreetMapProvider
 
 
 class OpenStreetMapAPI(OAuth):
-    url = "https://www.openstreetmap.org/api/0.6/user/details"
+    url = "https://api.openstreetmap.org/api/0.6/user/details"
 
     def get_user_info(self):
         raw_xml = self.query(self.url)


danieldegroot2:
Hey @pennersr,

If you are using OAuth2 for OpenStreetMap, please use the old URL for the OAuth2 handshake for now.
See the original issue linked above and https://github.com/osmlab/osm-auth/pull/123#issuecomment-1777330094

If you are only using OAuth 1 or Basic Auth for this provider, you can leave it as is until you have switched to OAuth 2.
See https://github.com/openstreetmap/operations/issues/867
I can open an issue for this if you want.

pennersr:
@danieldegroot2 Can you file a PR? Thanks.

danieldegroot2:
No, I don't think I can help here, I do not have enough coding experience for splitting URLs (nor switching to OAuth 2). You could alternatively revert the commit.

pennersr:
But, the allauth provider is of type OAuth 1 -- so this change is valid and need not be reverted then?

danieldegroot2:
Yes, then it is okay. I have opened new issue #3502

Issue #3488, opened 2023-10-18
Issue with Migration when the database is recreated from the beginning


elementzonline:
The following code block seems to be incorrect when the SITES_ENABLED is True. The AUTH_USER_MODEL dependency is not getting added to the dependency list when the SITES_ENABLED is True and causes the migration to fail.
```
class Migration(migrations.Migration):
    dependencies = (
        [
            ("sites", "0001_initial"),
        ]
        if app_settings.SITES_ENABLED
        else []
        + [
            migrations.swappable_dependency(settings.AUTH_USER_MODEL),
        ]
    )
```

Error during migration is as follows

    raise ValueError('Related model %r cannot be resolved' % self.remote_field.model)
ValueError: Related model 'auth.user' cannot be resolved

I resolved this issue by modifying the Migration file part like the one below
```
class Migration(migrations.Migration):
    dependencies = (
        [
            ("sites", "0001_initial"),
            migrations.swappable_dependency(settings.AUTH_USER_MODEL),
        ]
        if app_settings.SITES_ENABLED
        else []
        + [
            migrations.swappable_dependency(settings.AUTH_USER_MODEL),
        ]
    )
```

Hope this will get addressed in the next release.

pennersr:
https://codeberg.org/allauth/django-allauth/issues/3488

PR #3489, opened 2023-10-18
Update 0001_initial.py for migration bug fix

elementzonline:
# Submitting Pull Requests for fixing bug for the issue #3488

### DIFF
--- a/allauth/socialaccount/migrations/0001_initial.py
+++ b/allauth/socialaccount/migrations/0001_initial.py
@@ -12,7 +12,6 @@ class Migration(migrations.Migration):
     dependencies = (
         [
             ("sites", "0001_initial"),
-            migrations.swappable_dependency(settings.AUTH_USER_MODEL),
         ]
         if app_settings.SITES_ENABLED
         else []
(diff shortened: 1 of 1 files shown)


allauth/socialaccount/migrations/0001_initial.py
```diff
@@ -12,6 +12,7 @@ class Migration(migrations.Migration):
     dependencies = (
         [
             ("sites", "0001_initial"),
+            migrations.swappable_dependency(settings.AUTH_USER_MODEL),
         ]
         if app_settings.SITES_ENABLED
         else []
```
pennersr, line 18: I don't think this is correct. The user model is always a dependency, and the site model conditionally is. So, the original code is almost correct except for missing some brackets:

```
    dependencies = (
        [
            ("sites", "0001_initial"),
        ]
        if app_settings.SITES_ENABLED
        else []
    ) + [
        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
    ]
```

elementzonline:
Hi, 
updated the migrations file as noted. It also fixed the issue #3488

Issue #3503, opened 2023-10-28
Some templates missing {% load allauth %}


msapiro:
At least the templates `django-allauth/allauth/templates/socialaccount/login_cancelled.html`and `django-allauth/allauth/templates/account/verified_email_required.html` are missing {% load allauth %} to define the {% element %} tag.

pennersr:
https://codeberg.org/allauth/django-allauth/issues/3503

PR #3504, opened 2023-10-28
Load allauth

msapiro:
Fix https://github.com/pennersr/django-allauth/issues/3503 by adding missing {% load allauth %} to define {% element %} tag.


pennersr:
Thanks. I added coverage for this here: f2d7aa1b

PR #3536, opened 2023-11-27
feat(urls): Add error logging for provider imports

Antvirf:
# Background

At the moment, failing to import a provider and set up its URLs fails silently, leading to confusing error messages when certain paths aren't found:

```
NoReverseMatch at /login/
Reverse for 'saml_login' not found. 'saml_login' is not a valid view function or pattern name.
```

Most recently, I encountered this with the `SAML` provider as it currently relies on `libxmlsec` - the correct installation of which on an ARM Mac is difficult at the moment. It took me a long time to figure out that my URL paths weren't working due to an import issue, as I would have expected the library to tell me if something is wrong. [Someone else had also encountered something similar with openssl previously](https://forum.djangoproject.com/t/all-google-login-urls-unavailable-reverse-for-google-login-not-found/21795/6).

# Solution

Since the current `allauth/urls.py` chooses to `continue` despite `ImportError`s, I opted not to change that design. Instead, this PR just adds an `ERROR`-level logging message to inform the user that something went wrong during import.

# Output example

In my case, with a broken `libxmlsec` install, the output when starting the server is:

```
ERROR   2023-11-27 09:39:10,013 Unable to import allauth provider: saml: (dlopen(/Users/myuser/myproject/venv/lib/python3.11/site-packages/xmlsec.cpython-311-darwin.so, 0x0002): symbol not found in flat namespace '_xmlSecDSigNs').
```

# Submitting Pull Requests checklist


Antvirf:
@pennersr would be great to get your thoughts - do you want to keep catching these exceptions as of current, or would you prefer just throwing the exception?

pennersr:
Except for backwards compatibility, there is no good reason to catch those import errors. I've also bumped into a few occasions where catching the `ImportError` was silently hiding a different issue. So I would prefer to drop catching import errors.

PR #3543, opened 2023-11-30
fix(mfa): Use "one-time-code" autocomplete hint

julianwachholz:
This small change also makes it a little easier to detect the field in custom form templates (e.g. with `django-widget-tweaks`).

Reference: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete#sect16


pennersr:
There is one more case here:

```
class ActivateTOTPForm(forms.Form):
    code = forms.CharField(label=_("Authenticator code"))
```

Can you give that the same treatment?

Thanks.

PR #3579, opened 2023-12-25
Added ALLOW_SIGNUP setting in accounts to enable/disable signups in t…

musab-tatek:
Added ALLOW_SIGNUP setting in accounts to enable/disable signups in the Accounts Adapte

### DIFF
--- a/allauth/account/adapter.py
+++ b/allauth/account/adapter.py
@@ -243,7 +243,7 @@ class DefaultAccountAdapter(object):
         Next to simply returning True/False you can also intervene the
         regular flow by raising an ImmediateHttpResponse
         """
-        return True
+        return app_settings.ALLOW_SIGNUP
 
     def new_user(self, request):
         """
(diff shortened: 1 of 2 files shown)


allauth/account/adapter.py
```diff
@@ -243,7 +243,7 @@ def is_open_for_signup(self, request):
         Next to simply returning True/False you can also intervene the
         regular flow by raising an ImmediateHttpResponse
         """
-        return True
+        return app_settings.ALLOW_SIGNUP
```
pennersr, line 246: Naming wise this is inconsistent with the adapter method, would have accepted IS_OPEN_FOR_SIGNUP?

allauth/account/app_settings.py
```diff
@@ -144,6 +144,13 @@ def UNIQUE_EMAIL(self):
         Enforce uniqueness of email addresses
         """
         return self._setting("UNIQUE_EMAIL", True)
+    
+    @property
+    def ALLOW_SIGNUP(self):
+        """
+        Allow new registrations.
```
pennersr, line 151: This is a new setting -- it requires documentation.

pennersr (changes requested):

pennersr:
Wondering, do we really need a new setting for this -- why not just override the adapter method?

PR #3585, opened 2024-01-04
feat(ping): Add Ping Identity provider.

timw6n:
# Submitting Pull Requests

## General

 ## Provider Specifics

- [x] Add documentation to `docs/providers/<provider name>.rst` and `docs/providers/index.rst` Provider Specifics toctree.
- [x] Add an entry to the list of supported providers over at `docs/overview.rst`.  - *This doesn't seem to exist now.*

### DIFF
--- a/allauth/socialaccount/providers/pingidentity/provider.py
+++ b/allauth/socialaccount/providers/pingidentity/provider.py
@@ -1,0 +1,38 @@
+from allauth.account.models import EmailAddress
+from allauth.socialaccount.providers.base import ProviderAccount
+from allauth.socialaccount.providers.oauth2.provider import OAuth2Provider
+
+
+class PingIdentityAccount(ProviderAccount):
+    pass
+
+
+class PingIdentityProvider(OAuth2Provider):
+    id = "pingidentity"
+    name = "Ping Identity"
+    account_class = PingIdentityAccount
+
+    def get_default_scope(self):
+        return ["openid", "profile", "email"]
+
+    def extract_uid(self, data):
+        return str(data["preferred_username"])
+
+    def extract_extra_data(self, data):
+        return data
+
+    def extract_email_addresses(self, data):
+        # Assume emails are verified as there is no such claim, but this is a corporate (not end-user) provider.
+        return [EmailAddress(email=data["email"], verified=True, primary=True)]
+
+    def extract_common_fields(self, data):
+        email = data["email"]
+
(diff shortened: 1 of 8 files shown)


allauth/socialaccount/providers/pingidentity/provider.py
```diff
@@ -0,0 +1,38 @@
...
+    pass
+
+
+class PingIdentityProvider(OAuth2Provider):
+    id = "pingidentity"
+    name = "Ping Identity"
+    account_class = PingIdentityAccount
+
+    def get_default_scope(self):
+        return ["openid", "profile", "email"]
```
pennersr, line 16: It looks like PingIdentity is just a regular OpenID Connect provider. So, do we really need a dedicated provider -- can't we just use the existing `openid_connect` provider as is ?

pennersr (changes requested):

allauth/socialaccount/providers/pingidentity/provider.py
timw6n: Thanks for that. I had actually missed that there was an OpenID Connect provider as well as the OAuth2 one. 

Having checked the two across it looks like the only real difference would be that the name wouldn't come across as it is looking for `name` but you get split `firstName`, `lastName` from Ping. (Or at least from the instance thereof that I'm integrating with.)

The other thing is that Ping doesn't seem to output an `email_verified` field. The code I've sent is assuming that is True. In the context of this being a big enterprise identity provider that doesn't seem unreasonable (certainly in my usage it is safe enough) but I'll just flag that in case it's too liberal an assumption for being merged.

Whether the above differences are big enough to warrant a separate provider is up to you I guess...

pennersr: With respect to email verified, that is something that can already be configured, like so:

```
        This can be configured with a ``"verified_email"`` key in the provider
        app settings, or a ``"VERIFIED_EMAIL"`` in the global provider settings
        (``SOCIALACCOUNT_PROVIDERS``).  Both can be set to ``False`` or
        ``True``, or, a list of domains to match email addresses against.
```

So only the name issue remains, which I think does not warrant a separate provider. Using the existing (e.g. `pre_social_login`) adapter hooks you can already handle the `firstName` issue yourself, though I am wondering ... aren't they adhering to the payload as described here?:

https://openid.net/specs/openid-connect-basic-1_0.html

PR #3586, opened 2024-01-05
fix steam login problem

dchoruzy:
Fixes following error when trying to sign in using Steam provider
```

TypeError at /accounts/steam/callback/

Session.get() takes 2 positional arguments but 3 were given
```

### DIFF
--- a/allauth/socialaccount/providers/steam/provider.py
+++ b/allauth/socialaccount/providers/steam/provider.py
@@ -41,7 +41,7 @@ def request_steam_account_summary(api_key, steam_id):
     method = "ISteamUser/GetPlayerSummaries/v0002/"
     params = {"key": api_key, "steamids": steam_id}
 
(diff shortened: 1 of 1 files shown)


allauth/socialaccount/providers/steam/provider.py
```diff
@@ -41,7 +41,7 @@ def request_steam_account_summary(api_key, steam_id):
     method = "ISteamUser/GetPlayerSummaries/v0002/"
     params = {"key": api_key, "steamids": steam_id}
 
-    resp = get_adapter().get_requests_session().get(api_base + method, params)
+    resp = get_adapter().get_requests_session().get(api_base + method, prams=params)
```
pennersr, line 44:
```suggestion
    resp = get_adapter().get_requests_session().get(api_base + method, params=params)
```

pennersr:
Closed via 0bb3cec7

Issue #3599, opened 2024-01-18
Social Authentication Error when setting SOCIALACCOUNT_STORE_TOKENS


miqsoft:
Hello guys,

When setting SOCIALACCOUNT_STORE_TOKENS I get error ``save() prohibited to prevent data loss due to unsaved related object 'app'``. 

I assume it occurs since the SocialApp instance is not saved to database and instead created from data from the settings. However I do not know how resolve it properly.

Versions:
-----------
Django==4.2
django-allauth==0.60.1

Stacktrace:
------------

``` python3

Traceback (most recent call last):
  File "X:\venv\lib\site-packages\allauth\socialaccount\models.py", line 302, in _lookup_by_socialaccount
    t = SocialToken.objects.get(
  File "X:\venv\lib\site-packages\django\db\models\manager.py", line 87, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "X:\venv\lib\site-packages\django\db\models\query.py", line 637, in get
    raise self.model.DoesNotExist(
allauth.socialaccount.models.SocialToken.DoesNotExist: SocialToken matching query does not exist.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "X:\venv\lib\site-packages\django\core\handlers\exception.py", line 55, in inner
    response = get_response(request)
  File "X:\venv\lib\site-packages\django\core\handlers\base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "X:\venv\lib\site-packages\django\views\decorators\csrf.py", line 56, in wrapper_view
    return view_func(*args, **kwargs)
  File "X:\venv\lib\site-packages\django\views\generic\base.py", line 104, in view
    return self.dispatch(request, *args, **kwargs)
  File "X:\venv\lib\site-packages\django\utils\decorators.py", line 46, in _wrapper
    return bound_method(*args, **kwargs)
  File "X:\venv\lib\site-packages\django\views\decorators\debug.py", line 92, in sensitive_post_parameters_wrapper
    return view(request, *args, **kwargs)
  File "X:\venv\lib\site-packages\dj_rest_auth\views.py", line 48, in dispatch
    return super().dispatch(*args, **kwargs)
  File "X:\venv\lib\site-packages\rest_framework\views.py", line 509, in dispatch
    response = self.handle_exception(exc)
  File "X:\venv\lib\site-packages\rest_framework\views.py", line 469, in handle_exception
[...]

PR #3600, opened 2024-01-18
Fix issue #3599 

miqsoft:
Fix issue #3599 by checking before a token update if app instance is saved to database. If not set app instance to None.


pennersr:
Thanks -- I also wanted to capture this in a test case, so it's fixed here: 5f71a37b

PR #3609, opened 2024-01-22
New provider: Garmin Connect

soh55:
# Submitting Pull Requests

## General

 ## Provider Specifics

- [X] Make sure unit tests are available.
- [X] Add an entry of your provider in `test_settings.py::INSTALLED_APPS` and `docs/installation.rst::INSTALLED_APPS`.
- [X] Add documentation to `docs/providers/<provider name>.rst` and `docs/providers/index.rst` Provider Specifics toctree.
- [ ] Add an entry to the list of supported providers over at `docs/overview.rst`. -> Couldn't find this file


allauth/socialaccount/providers/garminconnect/client.py
```diff
@@ -0,0 +1,52 @@
+import requests
+from urllib.parse import parse_qsl
+
+from django.utils.translation import gettext as _
+from requests_oauthlib import OAuth1
```
pennersr, line 5: Question: this provider is based on OAuth 1.0 -- yet, reading over here https://developer.garmin.com/connect-iq/core-topics/authenticated-web-services/ it seems as if they also support OAuth 2.0. If that is the case, shouldn't 2.0 instead of 1.0 be integrated?

allauth/socialaccount/providers/garminconnect/client.py
```diff
@@ -0,0 +1,52 @@
...
+from allauth.socialaccount.providers.oauth.client import (
+    OAuthClient,
+    OAuthError,
+    get_token_prefix,
+)
+from allauth.utils import get_request_param
+
+class GarminConnectOAuthClient(OAuthClient):
+
+    def get_access_token(self):
```
pennersr, line 16: This seems like an almost exact copy of `OAuthClient.get_access_token()` -- can't we reuse that, if not, can't this be made more DRY?

allauth/socialaccount/providers/garminconnect/provider.py
```diff
@@ -0,0 +1,44 @@
+from allauth.socialaccount.providers.base import ProviderAccount
+from allauth.socialaccount.providers.oauth.provider import OAuthProvider
+
+
+class GarminConnectAccount(ProviderAccount):
+    def get_profile_url(self):
+        """
+        Garmin Connect API doesn't provide a profile URL
+        """
+        return None
```
pennersr, line 10: `ProviderAccount` already has this method as its default, so you can skip this.

allauth/socialaccount/providers/garminconnect/provider.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def get_avatar_url(self):
```
pennersr, line 12: Same -- skip

allauth/socialaccount/providers/garminconnect/provider.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def extract_uid(self, data):
+        return str(data['userId'])
+
+    def extract_common_fields(self, data):
+        # Garmin Connect API doesn't return any other data than userId
+        return {}
+
+    def sociallogin_from_response(self, request, response):
+        return super(GarminConnectProvider, self).sociallogin_from_response(request, response)
```
pennersr, line 41:
```suggestion
        return super().sociallogin_from_response(request, response)
```

No need to support Python 2.

pennersr (changes requested):

allauth/socialaccount/providers/garminconnect/provider.py
```diff
@@ -0,0 +1,44 @@
...
+class GarminConnectProvider(OAuthProvider):
+    id = 'garmin_connect'
+    name = 'Garmin Connect'
+    account_class = GarminConnectAccount
+
+    def get_default_scope(self):
+        return ['read', 'write']
+
+    def extract_uid(self, data):
+        return str(data['userId'])
```
pennersr, line 34: `userId` is already a string. Is `str()` really needed?

allauth/socialaccount/providers/garminconnect/client.py
soh55: Yeah I'm not 100% familiar with `OAuth1` api, but the major change I made was to explicitly pass `verifier` param when constructing `oauth` object. I can hop on a debugger and see if we can use vanilla `OAuthClient` class later.

soh55: Yeah verified that using the original implementation results in 400 response from the Garmin server as it expects verifier in the header. Please let me know if you'd like extra data to review.

pennersr: This likely benefits other providers as well.The base class is already checking for the `oauth_verifier` here: https://github.com/pennersr/django-allauth/blob/382e9afaa47dad7f0590d5026f5b128866b73ec4/allauth/socialaccount/providers/oauth/client.py#L113

Could you change that so that it passes that verifier to the OAuth1 client, if available?

soh55: Does modifying the base `OAuthClient`'s `get_access_token` method like this work for you?

```python
    def get_access_token(self):
        """
        Obtain the access token to access private resources at the API
        endpoint.
        """
        if self.access_token is None:
            request_token = self._get_rt_from_session()
            at_url = self.access_token_url

            # Passing along oauth_verifier is required according to:
            # http://groups.google.com/group/twitter-development-talk/browse_frm/thread/472500cfe9e7cdb9#
            # Though, the custom oauth_callback seems to work without it?
            oauth_verifier = get_request_param(self.request, "oauth_verifier")
            if oauth_verifier:
                at_url = at_url + "?" + urlencode({"oauth_verifier": oauth_verifier})

            oauth = OAuth1(
                self.consumer_key,
                client_secret=self.consumer_secret,
                resource_owner_key=request_token["oauth_token"],
                resource_owner_secret=request_token["oauth_token_secret"],
                verifier=oauth_verifier if oauth_verifier else None,
            )

            response = get_adapter().get_requests_session().post(url=at_url, auth=oauth)
            if response.status_code not in [200, 201]:
                raise OAuthError(
                    _("Invalid response while obtaining access token" ' from "%s".')
                    % get_token_prefix(self.request_token_url)
                )
            self.access_token = dict(parse_qsl(response.text))

            self.request.session[
                "oauth_%s_access_token" % get_token_prefix(self.request_token_url)
            ] = self.access_token
        return self.access_token

soh55: Hmm, seems like Garmin throws a 400 error because of below section. Would be grateful for some help here- would we want to implement an option to not pass this as a URL param?

```python
      if oauth_verifier:
          at_url = (
              at_url + "?" + urlencode({"oauth_verifier": oauth_verifier})
          )

soh55: @pennersr Sorry to bother you, but would be grateful for some clarification here for implementing `get_access_token`.

pennersr: I would suggest to add a `access_token_requires_oauth_verifier=True` parameter to the `OAuthClient.__init__`, and use this in the code that builds `at_url`.  Then, alter `_get_client()` to copy a similarly named `access_token_requires_oauth_verifier` attribute over from the adapter to the client. That way, this functionality is encapsulated in the base class and you do not need to alter `_get_client()` yourself.

Issue #3612, opened 2024-01-22
Error while retrieving session from request in middleware in async environment


jvdboog:
Hello!

I get the following error when I upgraded django-allauth from 55.2 to 60.1.

```
django  | 2024-01-22 12:51:24,637 ERROR    Internal Server Error: /cms/sprite-f0864303/
django  | Traceback (most recent call last):
django  |   File "/usr/local/lib/python3.11/site-packages/django/contrib/sessions/backends/base.py", line 187, in _get_session
django  |     return self._session_cache
django  |            ^^^^^^^^^^^^^^^^^^^
django  | AttributeError: 'SessionStore' object has no attribute '_session_cache'
django  | 
django  | During handling of the above exception, another exception occurred:
django  | 
django  | Traceback (most recent call last):
django  |   File "/usr/local/lib/python3.11/site-packages/asgiref/sync.py", line 534, in thread_handler
django  |     raise exc_info[1]
django  |   File "/usr/local/lib/python3.11/site-packages/django/core/handlers/exception.py", line 42, in inner
django  |     response = await get_response(request)
django  |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^
django  |   File "/usr/local/lib/python3.11/site-packages/allauth/account/middleware.py", line 18, in middleware
django  |     _remove_dangling_login(request, response)
django  |   File "/usr/local/lib/python3.11/site-packages/allauth/account/middleware.py", line 52, in _remove_dangling_login
django  |     if "account_login" in request.session:
django  |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
django  |   File "/usr/local/lib/python3.11/site-packages/django/contrib/sessions/backends/base.py", line 50, in __contains__
django  |     return key in self._session
django  |                   ^^^^^^^^^^^^^
django  |   File "/usr/local/lib/python3.11/site-packages/django/contrib/sessions/backends/base.py", line 192, in _get_session
django  |     self._session_cache = self.load()
django  |                           ^^^^^^^^^^^
django  |   File "/usr/local/lib/python3.11/site-packages/django/contrib/sessions/backends/db.py", line 42, in load
django  |     s = self._get_session_from_db()
django  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
django  |   File "/usr/local/lib/python3.11/site-packages/django/contrib/sessions/backends/db.py", line 32, in _get_session_from_db
django  |     return self.model.objects.get(
[...]

PR #3614, opened 2024-01-22
fix(account): Async session access

pennersr:
See #3612


bkbirr:
I was having the same error as https://github.com/pennersr/django-allauth/issues/3612. This works. Thanks!

unique1o1:
When will this be released in PYPI?

PR #3622, opened 2024-01-29
Use the panel element in account/email.html

foarsitter:
Discovered the new element api and, if I understood it correctly, email.html is missing a panel.


pennersr:
The current page looks like this:

image

With these changes, that becomes:

image

I don't think a panel is appropriate here.

foarsitter:
Indeed, that isn't an improvement. Thanks for having a look at it!

PR #3633, opened 2024-02-08
fix(microsoft): Fixed issue where get_avatar_url does not work

arowley-ai:
Issue
---------
This pull request fixes an with get_avatar_url for the Microsoft provider which as far as I can tell, does not work at all.

Solution
---------
To fix the issue two additional requests were necessary. Firstly, to get a little metadata about the return type for the user photo and then to pull the user photo itself. I proposed a new setting that will enable/disable this behaviour as clearly we have the opportunity to save time for the majority of requests that do not need a user photo.

I chose to entirely replace the get_avatar_url method in the provider. That could be a little controversial since maybe somewhere out there somebody has an app that actually returns a valid URL but I am reasonably confident that this is not possible. As far as I can tell, the only endpoint from which any user photo can be accessed is via "https://graph.microsoft.com/v1.0/me/photo" and the only way to access the photo is as a binary stream (rather than a url).

The get_avatar_url method will provide a base64 encoded image stream directly instead of a url. I would have preferred the url option but apparently not with Microsoft. Nonetheless this solution will still render the image in place exactly as a valid url would. I think therefore that it is consistent with the original intent of this function but please let me know if you disagree.

To do
---------
I am happy to update the documentation on this if it looks like something that can be accepted.

### DIFF
--- a/allauth/socialaccount/providers/microsoft/provider.py
+++ b/allauth/socialaccount/providers/microsoft/provider.py
@@ -6,7 +6,10 @@ from allauth.socialaccount.providers.oauth2.provider import OAuth2Provider
 
 class MicrosoftGraphAccount(ProviderAccount):
     def get_avatar_url(self):
-        return self.account.extra_data.get("photo")
+        return "data:{format};base64, {data}".format(
+            format=self.account.extra_data.get('photo_metadata', {}).get('@odata.mediaContentType', 'image/jpeg'),
+            data=self.account.extra_data.get('photo')
+        )
 
     def to_str(self):
         dflt = super(MicrosoftGraphAccount, self).to_str()
--- a/allauth/socialaccount/providers/microsoft/views.py
+++ b/allauth/socialaccount/providers/microsoft/views.py
@@ -1,6 +1,9 @@
 from __future__ import unicode_literals
 
 import json
+import base64
+
+from django.conf import settings
 
 from allauth.core import context
 from allauth.socialaccount import app_settings
@@ -13,6 +16,12 @@ from allauth.socialaccount.providers.oauth2.views import (
 )
 
 from .provider import MicrosoftGraphProvider
+
+FETCH_PHOTO = (
+    getattr(settings, "SOCIALACCOUNT_PROVIDERS", {})
+    .get("microsoft", {})
+    .get("FETCH_PHOTO", False)
+)
 
 
 def _check_errors(response):
@@ -35,6 +44,10 @@ def _check_errors(response):
 
 class MicrosoftGraphOAuth2Adapter(OAuth2Adapter):
     provider_id = MicrosoftGraphProvider.id
+    fetch_photo = FETCH_PHOTO
+    profile_url = "https://graph.microsoft.com/v1.0/me"
+    photo_metadata_url = "https://graph.microsoft.com/v1.0/me/photo"
+    photo_data_url = "https://graph.microsoft.com/v1.0/me/photo/$value"
 
     def _build_tenant_url(self, path):
         settings = app_settings.PROVIDERS.get(self.provider_id, {})
@@ -52,8 +65,6 @@ class MicrosoftGraphOAuth2Adapter(OAuth2Adapter):
     @property
     def authorize_url(self):
         return self._build_tenant_url("/oauth2/v2.0/authorize")
-
-    profile_url = "https://graph.microsoft.com/v1.0/me"
 
     user_properties = (
         "businessPhones",
@@ -84,6 +95,34 @@ class MicrosoftGraphOAuth2Adapter(OAuth2Adapter):
             )
         )
         extra_data = _check_errors(response)
+
+        if self.fetch_photo:
+
+            photo_metadata_response = (
+                get_adapter()
+                .get_requests_session()
+                .get(
+                    self.photo_metadata_url,
+                    headers=headers,
+                )
+            )
+            if not photo_metadata_response.ok:
+                raise OAuth2Error("Request to user photo metadata failed")
+
+            photo_data_response = (
+                get_adapter()
+                .get_requests_session()
+                .get(
+                    self.photo_data_url,
+                    headers=headers,
+                )
+            )
+            if not photo_data_response.ok:
+                raise OAuth2Error("Request to user photo data failed")
+
+            extra_data['photo_metadata'] = photo_metadata_response.json()
+            extra_data['photo'] = base64.b64encode(photo_data_response.content).decode('utf-8')
+
         return self.get_provider().sociallogin_from_response(request, extra_data)
 
 


arowley-ai:
Sorry for remaking the PR I got myself stuck into a battle with git and threw in the towel.

arowley-ai:
Hello @pennersr this looks a little stuck at the moment? Is there something I need to do?

allauth/socialaccount/providers/microsoft/provider.py
```diff
@@ -6,7 +6,10 @@
 
 class MicrosoftGraphAccount(ProviderAccount):
     def get_avatar_url(self):
-        return self.account.extra_data.get("photo")
+        return "data:{format};base64, {data}".format(
+            format=self.account.extra_data.get('photo_metadata', {}).get('@odata.mediaContentType', 'image/jpeg'),
+            data=self.account.extra_data.get('photo')
```
pennersr, line 11: **Issue**: I would expect that if `extra_data` does not contain `photo_metadata` or `photo`, then `get_avatar_url()` would return `None`, but in this case it returns a malformatted `data:...` string.

allauth/socialaccount/providers/microsoft/views.py
```diff
@@ -84,6 +95,34 @@ def complete_login(self, request, app, token, **kwargs):
...
+
+            photo_metadata_response = (
+                get_adapter()
+                .get_requests_session()
+                .get(
+                    self.photo_metadata_url,
+                    headers=headers,
+                )
+            )
+            if not photo_metadata_response.ok:
```
pennersr, line 109: **Question**: How does this behave if the user has no photo set up? Suppose it returns a 404, I don't think the OAuth handshake should hard fail on that...

allauth/socialaccount/providers/microsoft/views.py
```diff
@@ -84,6 +95,34 @@ def complete_login(self, request, app, token, **kwargs):
...
+                .get(
+                    self.photo_data_url,
+                    headers=headers,
+                )
+            )
+            if not photo_data_response.ok:
+                raise OAuth2Error("Request to user photo data failed")
+
+            extra_data['photo_metadata'] = photo_metadata_response.json()
+            extra_data['photo'] = base64.b64encode(photo_data_response.content).decode('utf-8')
```
pennersr, line 124: **Question**: I am not sure storing photo's as data URIs is in the `SocialAccount` table is a wise plan. Not sure about the Microsoft API, but photos can be really large, and turning that into data URIs results in an even bigger footprint.

Also see: https://stackoverflow.com/questions/695151/data-protocol-url-size-limitations

Also, always fetching the photo at login seems inefficient. I am really wondering if this should be builtin behavior, or whether this is best left as a project specific feature.

allauth/socialaccount/providers/microsoft/views.py
arowley-ai: Yes from the [documentation](https://learn.microsoft.com/en-us/graph/api/profilephoto-get?view=graph-rest-1.0&tabs=http#response-for-getting-the-photo) it looks like it would fail if the user has not set up a photo.

I can change to an "If self.fetch_photo is ok:" and put the rest of the logic underneath that if statement.

allauth/socialaccount/providers/microsoft/views.py
arowley-ai: > I am not sure storing photo's as data URIs is in the SocialAccount table is a wise plan ... uri size

There is a way to limit the uri size. 

Provided the user has any photo set up, Microsoft auto generates the 48x48 photo automatically and when making the api call you can specify 48x48 as a fixed size for example https://graph.microsoft.com/v1.0/me/photos/48x48/$value. A simple solution would be to just use that as the endpoint for the photo request and the equivelant for the metadata request. I have a feeling that endpoint is specifically designed to help get around this exact issue. 

> Always fetching the photo at login seems inefficient

This is a little trickier. I think if we're down to 48x48 we're back into reasonable data size so not sure if this is still a huge issue. I think we'd be talking 10 kilobytes at most. The downside of not requesting at login is that if the user photo changes then we would be using an old photo. There's no way to check if the user data has changed since last logon either as far as I can tell. I did check for a 'last updated' date in metadata and there didn't seem to be one. 

> I am really wondering if this should be builtin behavior, or whether this is best left as a project specific feature.

Of course that's really up to you. In favour of including this feature:

- We already provide get_avatar_url and for a project supporting multiple social logins this simply enables/makes the feature work equivelantly across all users and all platforms. 
- The solution works as users would expect it to and fits within the existing solution with minimal change

In favour of not including it:

- It feels a bit like a workaround and storing base64 photo data in the user object isn't quite the intent of the solution you've designed

allauth/socialaccount/providers/microsoft/provider.py
arowley-ai: Yes this seems super obvious upon reflection so I apologise, it's an easy change to make.

maruthiaws0808: from __future__ import unicode_literals
from allauth.socialaccount.providers.base import AuthAction, ProviderAccount
from allauth.socialaccount.providers.oauth2.provider import OAuth2Provider

class MicrosoftGraphAccount(ProviderAccount):
    def get_avatar_url(self):
        return self.account.extra_data.get("photo")
        return "data:{format};base64, {data}".format(
            format=self.account.extra_data.get('photo_metadata', {}).get('@odata.mediaContentType', 'image/jpeg'),
            data=self.account.extra_data.get('photo')
Owner
@[pennersr](https://github.com/pennersr) pennersr 2 days ago
Issue: I would expect that if extra_data does not contain photo_metadata or photo, then get_avatar_url() would return None, but in this case it returns a malformatted data:... string.

Author
@[arowley-ai](https://github.com/arowley-ai) arowley-ai yesterday
Yes this seems super obvious upon reflection so I apologise, it's an easy change to make.

maruthiaws0808 (commented):
from __future__ import unicode_literals
from allauth.socialaccount.providers.base import AuthAction, ProviderAccount
from allauth.socialaccount.providers.oauth2.provider import OAuth2Provider

class MicrosoftGraphAccount(ProviderAccount):
    def get_avatar_url(self):
        return self.account.extra_data.get("photo")
        return "data:{format};base64, {data}".format(
            format=self.account.extra_data.get('photo_metadata', {}).get('@odata.mediaContentType', 'image/jpeg'),
            data=self.account.extra_data.get('photo')
Owner
@[pennersr](https://github.com/pennersr) pennersr 2 days ago
Issue: I would expect that if extra_data does not contain photo_metadata or photo, then get_avatar_url() would return None, but in this case it returns a malformatted data:... string.

Author
@[arowley-ai](https://github.com/arowley-ai) arowley-ai yesterday
Yes this seems super obvious upon reflection so I apologise, it's an easy change to make.

allauth/socialaccount/providers/microsoft/views.py
pennersr: The methods exposed by `ProviderAccount` including `get_avatar_url` are intended to extract common information from the metadata payload of the account. In this case, the photo is clearly not available in the metadata, and I would prefer not to forcefully store it there. Even if we shrink it to 48x48 it all becomes quite project specific.

If you rely on this functionality within your project you can easily place that logic inside the `pre_social_login()` method of the social account adapter.  So, I prefer not to include this out of the box.

arowley-ai: Am I welcome to add an example of how to do so into docs?

pennersr: Use `pre_social_login()` et al easily becomes more than a simple few-liners example, it does not really fit well into the Microsoft provider documentation. Perhaps https://github.com/pennersr/django-allauth/discussions/categories/show-and-tell is a more suitable place?

arowley-ai: Ok if there's a space for me to put it in show and tell and we can include a one liner/link into the provider documentation just so it's able to be found if anyone needs it, then that sounds like a sensible plan.

Issue #1946, opened 2018-02-13
ACCOUNT_EMAIL_VERIFICATION customizable dinamically in AccountAdapter


rocchidavide:
Is it possible to set or change ACCOUNT_EMAIL_VERIFICATION in AccountAdapter? I'm trying to set as 'mandatory' for the company user profile, and 'optional' for the private user profile.

derek-adair:
I solved this w/ a single account adapter that has logic for this by reading an account_type field.  Not sure how worthy this is to add BUT i would use it.

Would it be silly to have it look for a defined field and let users override w/ a form defaulting to the setting?  How else might this be achieved?

pennersr:
It would definitely be useful to have a `adapter.get_email_verification_method(user)` method available.

derek-adair:
Could probably add "help wanted" tag, this *seems* like a good entry level PR for someone looking to get involved.

Hk669:
hey @pennersr can you give me a brief on this issue, i wanna work on this issue.

pennersr:
This should be it:
- Add a `get_email_verification_method(self, email): return app_settings.app_settings.EMAIL_VERIFICATION` to the account adapter
- Replace references to `app_settings.EMAIL_VERIFICATION` with a call to that method.

muneeb706:
I have tried adding this feature in PR: https://github.com/pennersr/django-allauth/pull/3648

pennersr:
Moved to https://codeberg.org/allauth/django-allauth/issues/1946

PR #3648, opened 2024-02-19
Custom account email verification through adapter #1946

muneeb706:
# Submitting Pull Requests

## General

 ## Provider Specifics

N/A


allauth/socialaccount/providers/authentiq/provider.py
```diff
@@ -58,14 +59,16 @@ class AuthentiqProvider(OAuth2Provider):
...
+        adapter = get_adapter()
         scope = set(super(AuthentiqProvider, self).get_scope(request))
         scope.add("openid")
 
         if Scope.EMAIL in scope:
             modifiers = ""
             if app_settings.EMAIL_REQUIRED:
                 modifiers += "r"
-            if app_settings.EMAIL_VERIFICATION:
+            #TODO: How to get user email here for get_email_verification_method ?
```
muneeb706, line 70: How can I get user's email in this method to pass it to get_email_verification_method ?

pennersr: This method is called to get a list of scopes to ask permission for as part of the OAuth handshake. So, at this point in time you do not know what user is going to be authenticated, yet.

allauth/account/adapter.py
```diff
@@ -752,6 +752,9 @@ def get_reauthentication_methods(self, user):
                 )
         return ret
 
+    def get_email_verification_method(self, email=None):
```
pennersr, line 755:
```suggestion
    def get_email_verification_method(self, login):
```

Now that we have a `Login` class representing the current login, I think it is more useful to base this method on an instance of that. See:

https://github.com/pennersr/django-allauth/blob/main/allauth/account/utils.py#L158

So, for that to happen, I think we need to:

- For every case where the default value (`email_verification=app_settings.EMAIL_VERIFICATION`) is passed, pass `None`. Then, inside the `__init__` of `Login`, when the email verification is not explicitly set to a value, call this adapter method.
-

muneeb706: Thank you for reviewing I'll work on it.

muneeb706: In assess_unique_email method https://github.com/pennersr/django-allauth/blob/main/allauth/account/utils.py#L496

We only have access to email not the Login instance as it is called before we complete signup. 
Is it a good idea to keep both login and email ?

allauth/socialaccount/providers/authentiq/provider.py
muneeb706: So, in this case login parameter in get_email_verification_method can be None ?

allauth/account/adapter.py
pennersr: Would it work if we change the signature of `assess_unique_email` to include a `email_verification` parameter?

muneeb706: How will it solve the problem in the sign up workflow ? Still login instance is not created before assess_unique_email method is invoked.

I think in the get_email_verification_method, if we have **kwargs, which will most likely contain, user fields (username, password, ...). Then developer overriding this method can make email_verification decision based on these field values. This will be beneficial, in case of custom sign up / login form (e-g add a field to determine email_verification method).

In the sign up and login workflow we can pass form fields and their values.

pennersr:
https://codeberg.org/allauth/django-allauth/pulls/3648

PR #3669, opened 2024-03-04
fix(templates): HTML validation issues

reggieriser:
I ran across a few HTML validation issues in the Django templates:
- an interleaved `slot` template tag that caused a closing `span` HTML tag to be dropped (for a field element with help text)
- some empty `for` attributes for `label`s on several pages

# Submitting Pull Requests

## General

### DIFF
--- a/allauth/templates/account/email_change.html
+++ b/allauth/templates/account/email_change.html
@@ -23,7 +23,7 @@
                 {% endelement %}
             {% endif %}
             {% if new_emailaddress %}
-                {% element field id="new_emailaddress" name="email" value=new_emailaddress.email disabled=True type="email" %}
+                {% element field id="new_emailaddress" value=new_emailaddress.email disabled=True type="email" %}
                     {% slot label %}
                         {% if not current_emailaddress %}
                             {% translate "Current email" %}:
@@ -44,7 +44,7 @@
                     {% endslot %}
                 {% endelement %}
             {% endif %}
-            {% element field id="email" name="email" value=form.email.value errors=form.email.errors type="email" %}
+            {% element field id=form.email.auto_id name="email" value=form.email.value errors=form.email.errors type="email" %}
                 {% slot label %}
                     {% translate "Change to" %}:
                 {% endslot %}
--- a/allauth/templates/account/email_change.html
+++ b/allauth/templates/account/email_change.html
@@ -16,14 +16,14 @@
         {% slot body %}
             {% csrf_token %}
             {% if current_emailaddress %}
-                {% element field disabled=True type="email" value=current_emailaddress.email %}
+                {% element field id="current_emailaddress" disabled=True type="email" value=current_emailaddress.email %}
                     {% slot label %}
                         {% translate "Current email" %}:
                     {% endslot %}
                 {% endelement %}
             {% endif %}
             {% if new_emailaddress %}
-                {% element field name="email" value=new_emailaddress.email disabled=True type="email" %}
+                {% element field id="new_emailaddress" name="email" value=new_emailaddress.email disabled=True type="email" %}
                     {% slot label %}
                         {% if not current_emailaddress %}
                             {% translate "Current email" %}:
@@ -44,7 +44,7 @@
                     {% endslot %}
                 {% endelement %}
             {% endif %}
-            {% element field name="email" value=form.email.value errors=form.email.errors type="email" %}
+            {% element field id="email" name="email" value=form.email.value errors=form.email.errors type="email" %}
                 {% slot label %}
                     {% translate "Change to" %}:
                 {% endslot %}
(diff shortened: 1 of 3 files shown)


allauth/templates/account/email_change.html
```diff
@@ -16,14 +16,14 @@
         {% slot body %}
             {% csrf_token %}
             {% if current_emailaddress %}
-                {% element field disabled=True type="email" value=current_emailaddress.email %}
```
pennersr, line 19: I don't think these (id=) changes belong here as part of this PR?

allauth/templates/mfa/totp/activate_form.html
```diff
@@ -16,7 +16,7 @@
             {% element img src=totp_svg_data_uri alt=form.secret tags="mfa,totp,qr" %}
             {% endelement %}
             {% csrf_token %}
-            {% element field type="text" value=form.secret disabled=True %}
+            {% element field id="authenticator_secret" type="text" value=form.secret disabled=True %}
```
pennersr, line 19: Same for this `id`.

pennersr (changes requested):

allauth/templates/account/email_change.html
reggieriser: These changes were intentional to address the empty `for` attributes mentioned in the PR description.  See [this line from field.html](https://github.com/pennersr/django-allauth/blob/ef0f66b8131964941f9b1dbe93086aa509602383/allauth/templates/allauth/elements/field.html#L3).  If you don't include an `id`, then the `label` value will be empty and not point to any input tag, which is invalid HTML.

Is there a better way to solve this?  I see `id` attributes used [for other field elements](https://github.com/pennersr/django-allauth/blob/ef0f66b8131964941f9b1dbe93086aa509602383/allauth/templates/mfa/recovery_codes/index.html#L11).

reggieriser: Sorry, it's not clear to me how I could leverage Django's `auto_id` since that's associated with a Django form and form field, and `{% element field %}` is sometimes used for fields that aren't part of a Django form or even have a `name` attribute (like this disabled one).  Or are you suggesting to change the `field.html` template to default to something like `id_%s` like a Django form does if an explicit `id` is not provided?  That could get a little tricky too since `name` is not required, although we could add it.

pennersr: The `fields.html` template over at the example project already uses `auto_id` as the default by using:

`id=bound_field.auto_id`

So, indeed, for fields that do not originate from a form field we can populate an `id` ourselves like you are doing. Though there is one field on line which could use `auto_id` I think.

reggieriser: Thanks for the pointer to the example project.  I think we're on the same page now.  I just changed the one email field backed by a form to use `auto_id`.

allauth/templates/account/email_change.html
```diff
@@ -16,14 +16,14 @@
...
+                {% element field id="current_emailaddress" disabled=True type="email" value=current_emailaddress.email %}
                     {% slot label %}
                         {% translate "Current email" %}:
                     {% endslot %}
                 {% endelement %}
             {% endif %}
             {% if new_emailaddress %}
-                {% element field name="email" value=new_emailaddress.email disabled=True type="email" %}
+                {% element field id="new_emailaddress" value=new_emailaddress.email disabled=True type="email" %}
```
reggieriser, line 26: Note that I removed the `name="email"` from this disabled element because that results in two inputs named "email" on the same form.  A disabled input shouldn't get sent with the form submission either way, but could be confusing.

allauth/templates/account/email_change.html
```diff
@@ -16,14 +16,14 @@
         {% slot body %}
             {% csrf_token %}
             {% if current_emailaddress %}
-                {% element field disabled=True type="email" value=current_emailaddress.email %}
+                {% element field id="current_emailaddress" disabled=True type="email" value=current_emailaddress.email %}
```
pennersr, line 19:
```suggestion
                {% element field id="current_email" disabled=True type="email" value=current_emailaddress.email %}
```

allauth/templates/account/email_change.html
pennersr, line 26:
```suggestion
                {% element field id="new_email" value=new_emailaddress.email disabled=True type="email" %}
```

pennersr (changes requested):

kedare:
Sorry for the question but what is the goal of this custom element tag ? Why not just put HTML tags directly ?

reggieriser:
> Sorry for the question but what is the goal of this custom element tag ? Why not just put HTML tags directly ?

It streamlines custom styling across all templates.  See [Styling the Existing Templates](https://docs.allauth.org/en/latest/common/templates.html#styling-the-existing-templates) from the docs.

Issue #3650, opened 2024-02-21
SAML EntityID can't be used in MS Azure AD / Entra ID SSO 


wrabit:
### The issue

The `EntityID` used in the xml metadata file of the ServiceProvider SP ( / the allauth host) is built automatically using the metadata endpoint url for the provider:

The value is automatically generated from [allauth/socialaccount/providers/saml/utils.py#L43-L45](https://github.com/pennersr/django-allauth/blob/main/allauth/socialaccount/providers/saml/utils.py#L43-L45)

```
    ...
    metadata_url = request.build_absolute_uri(reverse("saml_metadata", args=[org]))
    sp_config = {
        "entityId": metadata_url,
    ...
```

`build_absolute_uri` enforces a trailing slash which is not changeable by django or allauth configuration (unless I missed something)

Producing this metadata xml excerpt:
```<md:EntityDescriptor ... entityID="https://ngrok.xxx.com/accounts/saml/xxx/metadata/">```

MS Entra ID does not permit a trailing slash for the application ID:

<img width="568" alt="image" src="https://github.com/pennersr/django-allauth/assets/5918271/5ecc8235-4e82-4426-8533-ca8a48708e95">

### Possible solutions

For discussion:

- We make the EntityId configurable from settings
- We have some kind of `entity_id_without_trailing_slash` in the settings for `saml` providers which will `metadata_url[:-1]` it off

**Comment by pennersr** [comment]
> which is not changeable by django or allauth

Django allows for that, though indeed the allauth `urls.py` for the SAML provider includes:

```
                path(
                    "metadata/",
                    views.metadata,
                    name="saml_metadata",
                ),
```

It is a matter of removing the slash there, though that may cause backwards compatibility issues as it changes the metadata/entity URL. So the only way around this is to introduce a new setting `SAML_METADATA_URL_TRAILING_SLASH` which controls that URL. And, given that the current setting causes trouble it likely needs to default to `False`.

wrabit:
Having no slash might not a global standard - I know it's not permitted for Microsoft AD. If the default becomes False, it could still present breaking changes for existing setups. So maybe the other way would be better - default True to match the existing behaviour. What do you think?

pennersr:
See discussion over at #3670 -- entity ID is now configurable, this is no longer an issue.

pennersr:
https://codeberg.org/allauth/django-allauth/issues/3650

PR #3670, opened 2024-03-05
Add SAML_METADATA_URL_TRAILING_SLASH option

wrabit:
**As per**

- #3650

**This PR does 2 primary things:**

- Switches the default behaviour of placing a trailing slash on the published metadata url to removing it
- Allows control by setting a SAML_METADATA_URL_TRAILING_SLASH in settings.py


pennersr:
Do we still need this given that we now support a configurable entity ID?

See #3674

Issue #3673, opened 2024-03-06
[SAML] Make Service Provider (SP) settings configurable


mecampbellsoup:
In our application we implemented SAML so that customers could bring their own IdP SAML app configurations in order to authenticate their internal users with our application.

Initially we implemented SAML before it was incorporated into allauth natively.

Now, I am working on migrating away from our home-rolled code in order to use allauth's SAML implementation.

One obstacle I've encountered has to do with `allauth.socialaccount.providers.saml.utils.build_sp_config`: https://github.com/pennersr/django-allauth/blob/23be15ae7ecbf4b55ccc8ef4f947fc267c82b5ac/allauth/socialaccount/providers/saml/utils.py#L43

Our existing customer-provided SAML configurations were issued w/ a sort of "placeholder" SP entity ID as you can see [in our docs](https://docs.coreweave.com/cloud-ui/cloud-account-management/sso#configure-the-idp), i.e. `https://cloud.coreweave.com/sso/sp/metadata.xml`.

Further, we unfortunately instructed customers to [use that placeholder as the Audience URI](https://docs.coreweave.com/cloud-ui/cloud-account-management/sso/okta#configure-saml) in their IdP configuration.

As a result, it's likely that our customers' applications are configured w/ the placeholder as we instructed them, which means that there is a mismatch between `build_sp_config(...)["entityId"]` and the Audience URI constraint on their IdP side when I try to use the allauth SAML provider and related code with existing customer configurations/SAML settings.

One possible workaround would be to allow the SAML `SocialApp` settings to include an `"sp"` key where the service provider config can be declared, e.g.:

```python
SOCIALACCOUNT_PROVIDERS = {
    "saml": {
        # Here, each app represents the SAML provider configuration of one
        # organization.
        "APPS": [
            {
                # Used for display purposes, e.g. over by: {% get_providers %}
                "name": "Acme Inc",

                # Accounts signed up via this provider will have their
                # `SocialAccount.provider` value set to this ID. The combination
                # of this value and the `uid` must be unique. The IdP entity ID is a
                # good choice for this.
                "provider_id": "urn:dev-123.us.auth0.com",

                # The organization slug is configured by setting the
                # `client_id` value. In this example, the SAML login URL is:
                #
                #     /accounts/saml/acme-inc/login/
                "client_id": "acme-inc",

                # The fields above are common `SocialApp` fields. For SAML,
                # additional configuration is needed, which is placed in
                # `SocialApp.settings`:
[...]

PR #3674, opened 2024-03-06
refactor(saml): Make SP entityId configurable via app settings

mecampbellsoup:
Resolves https://github.com/pennersr/django-allauth/issues/3673.

See issue for context.


mecampbellsoup:
@pennersr I wasn't able to find a "making contributions" doc. It's possible I'm blind, I didn't look terribly hard... but I would be happy to add a test for this behavior once I can easily run them!

pennersr:
Could you also extend the example config payload over at the docs (`saml.rst`) to showcase this?

pennersr:
As for testing, you can extend the test case here:

https://github.com/pennersr/django-allauth/blob/main/allauth/socialaccount/providers/saml/tests.py#L160

So:
- Add another `provider_config` entry at line 163 (so that we have one with, and one without a custom SP entity ID).
- Add an assert `assert config["sp"]["entitiyId"] = ...` accordingly.

mecampbellsoup:
@pennersr test added!

Sorry for the probably dumb Q but, how can I run tests in my local repo checkout?

pennersr:
> @pennersr test added!
> 
> Sorry for the probably dumb Q but, how can I run tests in my local repo checkout?

That depends a bit on how you normally run things, e.g. docker Y/N... but if you have `tox` installed, you should be able to just run `tox`.  If you're running Nix, `nix-shell; pytest` will do....

mecampbellsoup:
@pennersr got tox running locally and everything and fixed the CI failure now. Thanks for your help.

mecampbellsoup:
@pennersr ping! 😄

PR #3710, opened 2024-03-29
feat(mfa): Add RequireMFAMiddleware

amickan:
This implementation was worked on independently, but it does bear a lot of resemblance to https://github.com/valohai/django-allauth-2fa/blob/main/allauth_2fa/middleware.py simply because there isn't really a different way of implementing middleware like this. Up to you whether and how you would want to reference the other implementation.  

# Submitting Pull Requests

## General

 ## Provider Specifics


allauth/mfa/middleware.py
```diff
@@ -0,0 +1,56 @@
...
+    """Force multi-factor authentication for certain users."""
+
+    allowed_urls = [
+        "account_login",
+        "account_change_password",
+        "account_logout",
+        "account_reset_password",
+        "mfa_activate_totp",
+    ]
+
```
N3it, line 21: I would highly recommend adding at least `account_reauthenticate` to the `allowed_urls` list to avoid infinite redirections. If MFA is enforced for a user, not activated yet and your session expire you will encounter the following loop:

- `account_reauthenticate`
- `mfa_activate_totp`
- `account_reauthenticate`
- ...

Adding `account_reauthenticate` to the list will solve that issue. More globally, it might be interesting to review all the URLs (e.g `account_inactive`...)

N3it (changes requested):

allauth/mfa/middleware.py
amickan: Thanks! That is a very good suggestion. I will also take another closer look at all other URLs. 
I have some deadlines at work this week, but should be able to work on this again next week.

amickan: I have updated the allowed urls. I had missed quite a few on second thought. The sign-in and sign-up urls don't need to be added in my opinion since the user won't be authenticated on those views yet and hence they are implicitly allowed.

N3it: Great! Already using it on my side. I hope it will be merged.

amickan:
Has anyone had a chance to look at this? Or am I missing a step to get this into review?

pennersr:
I must admit, I am not sure about this approach -- whether or not a middleware is the best way forward. For example, another way of nudging users into turning on MFA would be to add a login stage that must be completed before the user can proceed.  Also, the scope of such functionality is more about policies than authentication, and without a clear de facto way of handling things I am not sure it is wise to bring that into allauth. As for implementation, the `allowed_urls` approach feels a bit tricky. So, given all of this, and the fact that projects can easily add this themselves, I am leaning towards not bringing this into scope.

amickan:
Fair enough. I hadn't thought of adding this as a login stage. Trying that out just now, however, it seems that you would end up in an infinite loop since MFA set-up requires reauthentication?

pennersr:
Yeah, that won't be trivial either. For now, let's leave this all out of scope...

pennersr:
@jmsmkn I should have done a better job at explaining the reasoning. Here goes, hopefully.

First, the implementation over at #3710 is not without issues. It uses a middleware, with an `allowed_urls` list, which may be a bit brittle to maintain and get right. For example, changing your email address is not allowed, yet, activating MFA only works if you have zero verified email addresses.  UX wise it may also be regarded as a bit intrusive, you cannot even visit "/" without being promptly redirected to the MFA activation page.

Another way of implementing this would be to gently nudge the user to turn on MFA when the user signs in. Perhaps even allowing for a time window of a few days before this kicks in as mandatory.

And, in case you are solely doing this for staff users, perhaps another way of handling this would be to override `user.has_perm()` and always return `False` unless MFA is turned on.

Which leads me to the "Clear de facto way of handling things" remark -- given that there may be multiple approaches to the problem, and given that the proposed solution is not trivial, I prefer to not take that on board. It is really rather trivial for any project to add this middleware themselves, it's just a 60 lines or so middleware.

On the short term, I would be open to adding a more simple solution, e.g. a `@mfa_required` decorator that you can slap on your own views where needed. Such a decorator does not come at a maintenance cost, although it may not serve your needs if you are looking for a more site wide approach.

PR #3721, opened 2024-04-09
feat: Vendor OAuth1 from requests-oauthlib

akx:
## General

---

This PR vendors [the single file used from `requests_oauthlib`](https://github.com/requests/requests-oauthlib/blob/master/requests_oauthlib/oauth1_auth.py) to avoid an extra dependency on `requests_oauthlib` itself. The only changes made to the vendored file are mechanical (isort, black) and adding the ISC license header from `requests-oauthlib`.

### DIFF
--- a/allauth/socialaccount/providers/oauth/client.py
+++ b/allauth/socialaccount/providers/oauth/client.py
@@ -11,9 +11,8 @@ from django.http import HttpResponseRedirect
 from django.utils.http import urlencode
 from django.utils.translation import gettext as _
 
-from requests_oauthlib import OAuth1
-
 from allauth.socialaccount.adapter import get_adapter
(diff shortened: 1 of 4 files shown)


akx:
No particular problem, was just looking at dependency chains while looking at routine upgrades for projects.

Seems a waste to have one extra 1K LoC dependency for the single file that's being used from it.

pennersr:
I see that requests-oauthlib is also being used for OAuth2 purposes over at #3889 (which I still need to look into in detail btw).

akx:
Welp, #3889 is apparently now closed 😅

pennersr:
> Alternately, how about making requests-oauthlib an optional dependency for the integrations that still need OAuth1?

That would break backwards compatiblity. While possible, I don't see any real gain here. Let's skip this...

PR #3736, opened 2024-04-19
feat(i18n): add Estonian (`et`) translations

gekoke:
Provides Estonian translations.

Some formatting and metadata in the header might be off since I was getting the following error when running `makemessages`:
```
CommandError: Unable to find a locale path to store translations for file examples/react-spa/backend/backend/__init__.py. Make sure the 'locale' directory exists in an app or LOCALE_PATHS setting is set.
```

Please let me know if there is anything else that needs to be done before the changes can be accepted.

Thanks!


pennersr:
Looks good (the build actually tests compilation of messages) -- thanks!

PR #3738, opened 2024-04-20
🖌️ Use vector badges

Jamim:
_Hello @pennersr,_

I noticed that you recently dropped some `SVG` badges in b24e4310552a03839ad352e3316a06afd66489d8 and f4e4276ced68318fe1d6327978b25515a3857d66, but I believe this change might be reconsidered taking into account how poorly scaled `PNG`s might look.

Raster badges look blurry at scales different from 100%, so it's worth switching to vector badges which look fine at any scale.

_Raster badges at 170% scale on my machine:_
raster-badges

_Vector badges at 170% scale on my machine:_
vector-badges

## General

_Best regards!_

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2,32 +2,32 @@
 Welcome to django-allauth!
 ==========================
 
-.. |ci| image:: https://img.shields.io/github/actions/workflow/status/pennersr/django-allauth/ci.yml.png
+.. |ci| image:: https://img.shields.io/github/actions/workflow/status/pennersr/django-allauth/ci.yml.svg
    :target: https://github.com/pennersr/django-allauth/actions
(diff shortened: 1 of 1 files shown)


pennersr:
@Jamim Definitely, I don't think these PNGs are an improvement, but I couldn't get it to work with SVGs. Without any change on my end -- not sure if GitHub rendering changed somehow -- things looked like:

https://github.com/pennersr/django-allauth/tree/0.60.x

image

I tried various things, and it all did not work... so I am a bit puzzled why it works with your changes.

Jamim:
@pennersr I suppose you fixed the root cause of the issue when you inlined the badges in 1e081b35c18bcc5904b0d74d9079591d1deb275b.

pennersr:
I could swear that after trying this it still did not work, but in any case, thanks!

PR #3754, opened 2024-04-24
fix: add types for fields overridden in subclasses & correct some imports

akx:
* Adding some types/annotations makes IDEs complain a little less.
* Mypy was also complaining when a superclass implies a field is `None` but it gets assigned to something else in a subclass.
* On that note, Mypy was also complaining about imports done via implicit re-exports, so those get fixed too.

### DIFF
--- a/allauth/account/stages.py
+++ b/allauth/account/stages.py
@@ -5,7 +5,7 @@ from allauth.utils import import_callable
 
 
 class LoginStage:
-    key = None
+    key: str  # Set in subclasses
 
     def __init__(self, controller, request, login):
         if not self.key:
--- a/allauth/socialaccount/providers/base/views.py
+++ b/allauth/socialaccount/providers/base/views.py
@@ -5,7 +5,7 @@ from allauth.socialaccount.providers.base.utils import respond_to_login_on_get
 
 
 class BaseLoginView(View):
-    provider_id = None
+    provider_id: str  # Set in subclasses
 
     def dispatch(self, request, *args, **kwargs):
         provider = self.get_provider()
--- a/allauth/socialaccount/providers/oauth2/provider.py
+++ b/allauth/socialaccount/providers/oauth2/provider.py
@@ -1,4 +1,5 @@
 import warnings
+from typing import Optional, Type
 from urllib.parse import parse_qsl
 
 from django.http import HttpResponseRedirect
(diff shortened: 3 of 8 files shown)


allauth/account/stages.py
```diff
@@ -1,11 +1,13 @@
...
+
 from allauth.account.adapter import get_adapter
 from allauth.account.app_settings import EmailVerificationMethod
 from allauth.account.utils import resume_login, stash_login, unstash_login
 from allauth.utils import import_callable
 
 
 class LoginStage:
-    key = None
+    key: Optional[str] = None  # Set in subclasses
```
pennersr, line 10: Given a `LoginStage` instance, its `key` property can/should never be `None`. So typing wise, it should be `: str`, not `Optional`.

allauth/socialaccount/providers/base/views.py
```diff
@@ -1,11 +1,13 @@
...
+
 from django.views import View
 
 from allauth.socialaccount.adapter import get_adapter
 from allauth.socialaccount.providers.base.utils import respond_to_login_on_get
 
 
 class BaseLoginView(View):
-    provider_id = None
+    provider_id: Optional[str] = None
```
pennersr, line 10: The call to `.get_provider()` down always requires a `provider_id` -- typing wise this should not be optional.

pennersr (changes requested):

allauth/socialaccount/providers/base/views.py
akx: Sure. Looks like `saml.views.LoginView` doesn't set `provider_id` though?

allauth/socialaccount/providers/oauth2/provider.py
```diff
@@ -12,11 +13,12 @@
 from allauth.socialaccount.providers.oauth2.utils import (
     generate_code_challenge,
 )
+from allauth.socialaccount.providers.oauth2.views import OAuth2Adapter
 
 
 class OAuth2Provider(Provider):
     pkce_enabled_default = False
-    oauth2_adapter_class = None
+    oauth2_adapter_class: Optional[Type[OAuth2Adapter]] = None
```
pennersr, line 21: This shouldn't be optional.

pennersr (changes requested):

allauth/socialaccount/providers/base/views.py
pennersr: Correct -- but it overrides the `get_provider()` method. If you make `provider_id` optional, I would expect `mypy` to complain about this method:

```
    def get_provider(self):
        provider = get_adapter().get_provider(self.request, self.provider_id)
        return provider
```

... as here we're passing `self.provider_id` to a method that does not allow `None`. SAML overrides that method.

allauth/socialaccount/providers/oauth2/provider.py
akx: Removed the default and changed (and moved) the instantiation check to raise a real error instead of a warning.

PR #3767, opened 2024-04-26
fix: Cannot serialize BinaryField wihout a value

radek-senfeld:
When serializing BinaryField(blank=True, null=True) containing None value an exception is thrown:

```
TypeError at /accounts/google/login/callback/ a bytes-like object is required, not 'None Type'
```

This patch kinda fixes the problem.

### DIFF
--- a/allauth/utils.py
+++ b/allauth/utils.py
@@ -168,7 +168,8 @@ def serialize_instance(instance):
         try:
             field = instance._meta.get_field(k)
             if isinstance(field, BinaryField):
-                v = force_str(base64.b64encode(v))
+                if v:
+                    v = force_str(base64.b64encode(v))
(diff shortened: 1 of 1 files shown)


allauth/utils.py
```diff
@@ -168,7 +168,8 @@ def serialize_instance(instance):
         try:
             field = instance._meta.get_field(k)
             if isinstance(field, BinaryField):
-                v = force_str(base64.b64encode(v))
+                if v:
```
pennersr, line 171:
```suggestion
                if v is not None:
```

Otherwise, `b""` would be turned into `None`?

radek-senfeld: I see. Good catch!

PR #3291, opened 2023-03-28
Lint upgrade

akx:
This PR:

* upgrades the version of Black used from 22.3 to 23.1, configures it to apply to all files in the repo, and runs it.
  * This removes some newlines within class definitions.
* switches linting from flake8 + isort over to [ruff](https://beta.ruff.rs/), a new fast linter for Python.
  * The isort rules weren't 100% portable (notably, no custom `django` section), but I tried to keep changes to the minimum.
  * I used [flake8-to-ruff](https://pypi.org/project/flake8-to-ruff/) for porting the selected rules; more could be added later.
  * There is currently no make target to run `ruff --fix`, but there could be (or the configuration could be switched to pre-commit, see e.g. https://github.com/valohai/django-allauth-2fa/pull/150 or #3085 here.

PR #3768, opened 2024-04-26
chore: switch linting and formatting to ruff

akx:
This PR proposes to switch the linting + formatting toolchain from flake8 + isort + black to [Ruff](https://github.com/astral-sh/ruff), a fast, very capable linter + formatter. 

I didn't enable any Ruff rules beyond those that were already enabled by the flake8 + isort configuration.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -4,7 +4,7 @@
 
  - [ ] Make sure you use [semantic commit messages](https://seesparkbox.com/foundry/semantic_commit_messages).
        Examples: `"fix(google): Fixed foobar bug"`, `"feat(accounts): Added foobar feature"`.
- - [ ] All Python code must formatted using Black, and clean from pep8 and isort issues.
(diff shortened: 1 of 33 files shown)


pennersr:
See: https://github.com/pennersr/django-allauth/pull/3291#issuecomment-1491998363

akx:
Hah, I had forgotten I'd proposed this last year. A lot of stuff has happened in Ruff land since, though – it now does formatting too :)

pennersr:
No problem. The issue of the binary seems to be unresolved though...

akx:
Well, it _is_ a Rust binary, but they're shipping `musllinux` wheels now too, if that helps.

akx:
What is the issue with pip installing the binary? 🤔 You end up doing that for e.g. `cryptography` too (though an .so in that case, not an executable).

pennersr:
This: https://github.com/astral-sh/ruff/issues/1699

PR #3770, opened 2024-04-26
Add API support: allauth.headless

pennersr:


docs/_static/headless/openapi-specification/openapi.yaml
```diff
@@ -0,0 +1,2442 @@
...
+        "404":
+          $ref: "#/components/responses/TOTPAuthenticatorNotFound"
+        "200":
+          $ref: "#/components/responses/TOTPAuthenticator"
+    post:
+      tags:
+        - "Account: 2FA"
+      summary: Activate TOTP
+      description: |
+        Activate TOTP authentication by means of inputting a code. The code
```
grahamhealy20, line 740: I feel like this could be rephrased as something like:

> The code should be provided from the consuming TOTP authenticator application which was generated using the TOTP authenticator secret retrieved from the TOTP authenticator status endpoint.

docs/_static/headless/openapi-specification/openapi.yaml
```diff
@@ -0,0 +1,2442 @@
...
+      allOf:
+        - $ref: "#/components/schemas/BaseAuthenticator"
+        - type: object
+          properties:
+            type:
+              type: string
+              description: |
+                The authenticator type.
+              enum:
+                - recovery_cdes
```
grahamhealy20, line 1989: Should this be:

```suggestion
                - recovery_codes
```

docs/_static/headless/openapi-specification/openapi.yaml
```diff
@@ -0,0 +1,2442 @@
...
+                  $ref: "#/components/examples/InvalidEmail"
+        "401":
+          $ref: "#/components/responses/AuthenticationOrReauthentication"
+    put:
+      tags:
+        - "Account: Email"
+      summary: Request email verification
+      description: |
+        Requests for (another) email verification email to be sent. Note that
+        sending emails is rate limitted, so when you send too many requests the
```
grahamhealy20, line 629: Small typo: 

```suggestion
        sending emails is rate limited, so when you send too many requests the
```

allauth/headless/internal/restkit/response.py
```diff
@@ -0,0 +1,28 @@
+from django.http import JsonResponse
+from django.utils.cache import add_never_cache_headers
+
+
+class APIResponse(JsonResponse):
+    def __init__(self, request, errors=None, data=None, meta=None, status=200):
+        d = {"status": status}
```
grahamhealy20, line 7: Small aside: Is it useful to include the status in the response body? Usually the http client library would return the status code alongside the response body. e.g. 

```
response.status
```

Instead of 

```
response.data.status
```

grahamhealy20:
Hi @pennersr. Thank you so much for this, this is some amazing work! 🏆 This is a really well done, well thought out, un-opinionated REST API implementation that works with any Django API framework, and the high quality really shows here. 💪 

I've been building a reference implementation using Next JS to test things out and it's been working flawlessly so far, I'll just be replying with comments in here as I find things if that's okay! 

Thanks again!

docs/_static/headless/openapi-specification/openapi.yaml
pennersr: Applied.

docs/_static/headless/openapi-specification/openapi.yaml
pennersr: Applied

allauth/headless/internal/restkit/response.py
pennersr: This is done to help out with Javascript, where we typically have this:

```
fetch(url)
    .then((response) => {
      return response.json();
    })
    .then((data) => {
      // here, response is not available, so we cannot access the status.
    })
```

... or:

```
async function fetchFoo(url) {
   return await fetch(url).json()
}
```

In both cases the `response` is gone.

pennersr:
@grahamhealy20 Thanks for the positive feedback, as well as for reviewing and testing! This is a big new chunk of code, and knowing that others are actively engaging with this really helps.

allauth/headless/internal/restkit/response.py
grahamhealy20: Got it, thanks for the clarification! 😄

Issue #3728, opened 2024-04-15
Is allauth open to including an (optional) honeypot field on registration?


RileyMathews:
Hey there! First off thanks for making such a great library! I've used it in virtually every Django project I start.

I have a public Django app using allauth that was recently hit by a steady swarm of bot login attempts. I was able to quickly add a 'honeypot' system by extended the registration form class that seemed to stem the tide for now. That had me curious if this system is something that Allauth would be interested in having as a built in option? I would be open to working on a PR for the implementation. My thought is it would be configurable with at the very least having options for...
* enabling/disabling the honeypot system
* an option for naming the honeypot field so it does not conflict with any other expected registration fields. (for example I named my field 'phone_number' as my app does not actually collect them but this might conflict with other apps registration forms)

Like I said just curious if this is something you are open to including? Open to discussion on this!

pennersr:
You mention "by a steady swarm of bot login attempts" and "by extended the registration form class" -- so does your honeypot solution protect only the signup form, or both?

Must admit, I don't have any actual insights on whether or not honeypot fields are still effective in 2024, but if so, then it is indeed something useful to add. Are there at all any stats / publications on this?

I guess the only setting needed would be a `ACCOUNT_SIGNUP_FORM_HONEYPOT_FIELD = "phone_number"` (or, `""` / `None` to have it turned off).

RileyMathews:
Ahh my mistake in saying 'steady stream of login' it was indeed a steady stream of signups. I was able to identify them relatively easy because I have not marketed this app publically at all and have only sent it to a few close friends for 'beta testing'. The signups had garbage usernames and never verified their emails or took any other actions in the app.

To be completely honest I was slightly surprised it worked as well. I was already also looking into how to add captcha but decided to throw in the honeypot quickly as I knew captcha would be more involved just to see if it would stem the tide. I'll spend some time doing some more research on how effective honeypots actually are in general today and bring my findings back here before I start work on this.

RileyMathews:
[...]

PR #3774, opened 2024-04-30
Add honeypot to sign up form

RileyMathews:
# Submitting Pull Requests

## General

 - [ ] Make sure you use [semantic commit messages](https://seesparkbox.com/foundry/semantic_commit_messages).

This PR is for https://github.com/pennersr/django-allauth/issues/3728. I am adding an optional honeypot field to the sign up form that will not create an account if filled out.

### DIFF
--- a/allauth/account/forms.py
+++ b/allauth/account/forms.py
@@ -328,7 +328,7 @@ class BaseSignupForm(_base_signup_form_class()):
         if honeypot_field_name := app_settings.SIGNUP_FORM_HONEYPOT_FIELD:
             # TODO: hide input field
             self.fields[honeypot_field_name] = forms.CharField(
-                label=None,
+                label=False,
                 widget=forms.TextInput(
                     attrs={
                         "style": "position: absolute; right: -99999px;"
--- a/allauth/account/forms.py
+++ b/allauth/account/forms.py
@@ -329,6 +329,7 @@ class BaseSignupForm(_base_signup_form_class()):
             # TODO: hide input field
             self.fields[honeypot_field_name] = forms.CharField(
                 label=False,
+                required=False,
                 widget=forms.TextInput(
                     attrs={
                         "style": "position: absolute; right: -99999px;",
--- a/allauth/account/forms.py
+++ b/allauth/account/forms.py
@@ -325,6 +325,13 @@ class BaseSignupForm(_base_signup_form_class()):
             self, getattr(self, "field_order", None) or default_field_order
         )
 
+        if honeypot_field_name := app_settings.SIGNUP_FORM_HONEYPOT_FIELD:
+            # TODO: hide input field
+            self.fields[honeypot_field_name] = forms.CharField(
+                label=honeypot_field_name,
+                widget=forms.TextInput(),
+            )
+
     def clean_username(self):
         value = self.cleaned_data["username"]
         value = get_adapter().clean_username(value)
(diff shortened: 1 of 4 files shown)


allauth/account/forms.py
```diff
@@ -371,7 +385,17 @@ def try_save(self, request):
...
+
+        If the honeypot was filled out then also send 'email verification sent'
+        response so that the bot thinks things worked.
         """
+        if honeypot_field_name := app_settings.SIGNUP_FORM_HONEYPOT_FIELD:
+            if self.cleaned_data[honeypot_field_name]:
+                user = None
+                adapter = get_adapter()
+                resp = adapter.respond_email_verification_sent(request, None)
+                return user, resp
```
RileyMathews, line 397: Open to discussion on exactly what to do when the honeypot is filled out. Best practices say to make it seem to the bot that 'everything worked' so this seemed like an easy response to copy.

pennersr: Given that there isn't any hard evidence of what works and what doesn't, I think it's best to not over complicate things and take the simplest approach possible.  So reusing the existing response seems fine.

ChangeLog.rst
```diff
@@ -9,6 +9,8 @@
   accordingly. Migrations are in place.  For rationale, see the note about email
   case sensitivity in the documentation.
 
+- Added an optional honeypot field that will not create an account if filled out
```
pennersr, line 12:
```suggestion
- Added an optional honeypot field to the signup form preventing an account to be created when it gets filled out.
```

pennersr (changes requested):

pennersr:
The build is failing -- likely due to use of the walrus operator for python 3.7. Also, black is complaining. Can you have a look? Thanks.

RileyMathews:
@pennersr thanks for the feedback! I still have not been able to get all of the tox environments running locally for me but I was able to check black against my changes and also I believe this should now be 3.7 compatible. I think I'm ready for another review when you have time.

pennersr:
Thanks. Though, according to the build there is still a walrus operator here:

```if honeypot_field_name := app_settings.SIGNUP_FORM_HONEYPOT_FIELD:```

RileyMathews:
@pennersr Ok this should be ready now. I was finally able to get the full tox run working for me and everything except flake is passing. However it appears to be a file I did not touch and the flake error I get locally wasn't there in the last CI run here (though I did fix the error in the test file I did introduce) so I am guessing the CI here only runs flake on changed files?

pennersr:
I noticed that when running this over at the example project, `style` and `tabindex` are not picked up:
 
```<input name="telephone" placeholder="['\n            False\n        ']" class="  form-control rounded-3" id="id_telephone" autocomplete="nope" value="" type="text">```

There is also something odd going on with the placeholder attribute. TBD.

pennersr:
Merged via 8281bd77

RileyMathews:
@pennersr Thanks for merging this! Did you ever get the example app working with this feature? If not I can take a look and try to figure out what is going on there. As I was developing this I installed the local package into one of my own Django projects to test with and everything seemed to come out correctly there.

pennersr:
@RileyMathews Yes -- all is good, thanks

ish91:
This is a great idea; thanks for implementing it!
IMHO It would be even better if we added the honeypot field to the login form as well. What do you think?

pennersr:
https://codeberg.org/allauth/django-allauth/pulls/3774

PR #3798, opened 2024-05-12
feat(templates): Add additional links to the base template menu

brianhelba:
These views are otherwise orphaned and undiscoverable via normal user navigation.

### DIFF
--- a/allauth/templates/allauth/layouts/base.html
+++ b/allauth/templates/allauth/layouts/base.html
@@ -46,7 +46,7 @@
                         {% url 'usersessions_list' as usersessions_list_url %}
                         {% if usersessions_list_url %}
                             <li>
-                                <a href="{{ usersessions_list_url }}">{% trans "User Sessions" %}</a>
+                                <a href="{{ usersessions_list_url }}">{% trans "Sessions" %}</a>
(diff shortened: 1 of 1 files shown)


allauth/templates/allauth/layouts/base.html
```diff
@@ -31,6 +31,24 @@
...
+                        {% url 'mfa_index' as mfa_url %}
+                        {% if mfa_url %}
+                            <li>
+                                <a href="{{ mfa_url }}">{% trans "Two-Factor Authentication" %}</a>
+                            </li>
+                        {% endif %}
+                        {% url 'usersessions_list' as usersessions_list_url %}
+                        {% if usersessions_list_url %}
+                            <li>
+                                <a href="{{ usersessions_list_url }}">{% trans "User Sessions" %}</a>
```
pennersr, line 49:
```suggestion
                                <a href="{{ usersessions_list_url }}">{% trans "Sessions" %}</a>
```

brianhelba: Sure, that string is already translated too.

pennersr:
Merged via 2511802f -- thanks

PR #3809, opened 2024-05-14
docs: Change admin site workaround documentation

### LINKED ISSUE #3786: Allauth admin integration issues
# Problem
To force allauth workflow in Django admin the docs propose this change:

``` python
from django.conf import settings
from django.contrib import admin
from django.contrib.admin.views.decorators import staff_member_required

admin.site.login = staff_member_required(
    admin.site.login, login_url=settings.LOGIN_URL
)
```

# Django cookiecutter workaround
This however leads to `ERR_TOO_MANY_REDIRECTS`.

Django cookiecutter uses this as a workaround:

``` python
from django.contrib.auth import decorators

if settings.DJANGO_ADMIN_FORCE_ALLAUTH:
    # Force the `admin` sign in process to go through the `django-allauth` workflow:
    # https://django-allauth.readthedocs.io/en/stable/advanced.html#admin
    admin.site.login = decorators.login_required(admin.site.login)  # type: ignore[method-assign]
```

That however makes it possible to go around the allauth by logging in as a non-superuser.
When logging in as a normal user, the user gets redirected to the Django admin site with "You are logged in but do not have access to this page. Do you want to log in as a different user?".

# Discussion
We discussed the issue and possible workarounds here:
cookiecutter/cookiecutter-django#4766

We figured out a possible way to avoid the "You are logged in but do not have access to this page. Do you want to log in as a different user?" and drop connections, that are not authenticated as superusers.

I am not sure if that is the best thing to do.
@pennersr do you have any recommendations here?
I am happy to contribute I am just not sure where to take this from here.

jkaeske:
The documented workaround was resulting in ERR_TOO_MANY_REDIRECTS. To solve this issue a workaround using a custom decorator was proposed.

Fixes #3786

# Submitting Pull Requests

## General

### DIFF
--- a/docs/common/admin.rst
+++ b/docs/common/admin.rst
@@ -15,9 +15,32 @@ the normal Django allauth workflow.
     * Any other custom workflow that overrides the Django allauth adapter's
       login method will not be applied.
 
-An easy workaround for this is to require users to login before going to the
-Django admin site's login page (note that the following would need to be applied to
-every instance of ``AdminSite``):
+An easy workaround for this is to require users to login before going to the Django
+admin site's login page. For this workaround the following custom decorator is needed:
+
+.. code-block:: python
+
+  from django.conf import settings
+  from django.contrib.admin.views.decorators import staff_member_required
+  from django.contrib.auth import REDIRECT_FIELD_NAME
+  from django.contrib.auth.decorators import login_required
+
+  def staff_member_required_or_redirect(
+      view_func=None,
+      redirect_field_name=REDIRECT_FIELD_NAME,
+      login_url=settings.LOGIN_URL,
+      fail_url=settings.LOGIN_REDIRECT_URL,
+  ):
+      return login_required(
+          staff_member_required(
+          view_func, redirect_field_name=None, login_url=fail_url
+          ),
+          redirect_field_name=redirect_field_name,
+          login_url=login_url,
+      )
+
+This decorator can then be used to force the allauth login flow in the admin site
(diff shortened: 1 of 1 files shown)


pennersr:
IMO protecting the admin is really important, so I think it would make sense to promote that code to an actual decorator that just works out of the box, so that people do not need to copy chunks of code around. How about we add that decorator to `allauth.account.decorators`?

jkaeske:
Sounds great to me!
Let me update the code in the next days.

pennersr:
If I use this decorator, while logged in as a regular user, and then visit an admin page, I am redirected to `LOGIN_REDIRECT_URL`, but it is not really obvious why that happens. Wouldn't it be more clear to just raise a `PermissionDenied` ?

pennersr:
Also, perhaps nitpicking, but the name `staff_member_required_or_redirect` is a bit long. Given that this decorator has only one purpose, securing the admin login, perhaps name it like that, so that we get:

```
from allauth.account.decorators import secure_admin_login

admin.site.login = secure_admin_login(admin.site.login)
```

pennersr:
While I was toying with this anyway, I figured I could just as well file a PR right away. Could you take #3818 for a test spin?

pennersr:
The Django admin also uses `PermissionDenied` in case you try to access things you don't have access to, so this way the behavior is conform to what the admin is already doing. I am not sure a redirect makes much sense if there is no explanation/message on why you are ending up there...

PR #3817, opened 2024-05-15
Adjusted check for verified email in mfa activate form clean method

je-bugshell:
There is an issue where a user can have multiple email addresses associated with their account. Even though their primary email is verified and valid, if they have a secondary email address that is not verified, they cannot activate mfa.

 I'm not entirely sure the reason why you would want this behavior.


pennersr:
There is a risk that somebody else signs up using your email address, and prevents you from creating an account yourself. Normally, you can easily get out of such situation -- your email address is attached, so you can easily recover the account. However, if that person turns on 2FA, you will not be able to gain access to the account with your email address on it.

PR #3832, opened 2024-05-21
fix(kakao): Change deprecated properties to profile information

juchajam:
# Submitting Pull Requests

## General

### DIFF
--- a/allauth/socialaccount/providers/kakao/provider.py
+++ b/allauth/socialaccount/providers/kakao/provider.py
@@ -14,7 +14,9 @@ class KakaoAccount(ProviderAccount):
         return self.account.extra_data.get("kakao_account", {}).get("profile", {})
 
     def get_avatar_url(self):
-        return self.profile.get("profile_image_url", self.properties.get("profile_image"))
+        return self.profile.get(
+            "profile_image_url", self.properties.get("profile_image")
+        )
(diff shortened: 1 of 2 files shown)


allauth/socialaccount/providers/kakao/provider.py
```diff
@@ -6,15 +6,15 @@
...
 class KakaoAccount(ProviderAccount):
     @property
-    def properties(self):
-        return self.account.extra_data.get("properties")
+    def profile(self):
+        return self.account.extra_data.get("kakao_account", {}).get("profile")
 
     def get_avatar_url(self):
-        return self.properties.get("profile_image")
+        return self.profile.get("profile_image_url")
```
pennersr, line 13: Suppose there are existing accounts with the original (legacy) data, then the changes above would crash. I think we should fallback to the original fields in case the new lookups fail?

pennersr (changes requested):

allauth/socialaccount/providers/kakao/provider.py
juchajam: I fixed the code to use both profile and properties.

pennersr:
Thanks for the backwards compatibility fix. Only thing left is to make `black` happy...

PR #3835, opened 2024-05-21
feat: add app setting 'READONLY_EMAIL_FIELD'

dgatland:
# Submitting Pull Requests

## General

### DIFF
--- a/allauth/account/app_settings.py
+++ b/allauth/account/app_settings.py
@@ -160,6 +160,13 @@ class AppSettings(object):
         return self._setting("SIGNUP_EMAIL_ENTER_TWICE", False)
 
     @property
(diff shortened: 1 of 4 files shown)


pennersr:
Thoughts:
- How is social signup handled? So, I click on an invitation, see the form with the read-only email, but then click e.g. Google to signup. That would bypass the read-onlyness.
- If I receive an invitation link I clearly have access to the inbox the link was mailed to. Now, while you can make it read-only at signup, there is nothing prevent a user to change the email after signing up. So if the email can be freely changed anyway, what is the use of making it read-only?

dgatland:
Thanks for your comments @pennersr 

I think you're right that the feature probably isn't as generalised as we initially expected. Closing this PR and we will address separately.

PR #3866, opened 2024-06-04
fix(mfa): Add types to `allauth.mfa` flows

akx:
This PR adds a bunch of types to `allauth.mfa`'s internals.

Our app needs to set up MFA with a slightly custom flow, so we can't use the forms or headless features, so I needed to delve a bit into how things work under the hood, and figured it'd be useful if types were documented technically too, for IDEs etc.

Feels like `TOTP` and `RecoveryCodes` should have a superclass, but that's beyond the scope of this PR...

### DIFF
--- a/allauth/mfa/models.py
+++ b/allauth/mfa/models.py
@@ -47,6 +47,6 @@ class Authenticator(models.Model):
             self.type
         ](self)
 
-    def record_usage(self):
+    def record_usage(self) -> None:
         self.last_used_at = timezone.now()
         self.save(update_fields=["last_used_at"])
--- a/allauth/mfa/recovery_codes.py
+++ b/allauth/mfa/recovery_codes.py
@@ -3,6 +3,7 @@ import hmac
 import os
 import struct
 from hashlib import sha1
+from typing import List, Optional
 
 from allauth.mfa import app_settings
 from allauth.mfa.models import Authenticator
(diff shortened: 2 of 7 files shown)


allauth/mfa/models.py
```diff
@@ -14,7 +14,7 @@
 
 
 class AuthenticatorManager(models.Manager):
-    def delete_dangling_recovery_codes(self, user):
+    def delete_dangling_recovery_codes(self, user) -> "Authenticator":
```
pennersr, line 17: **Issue**: This can also return `None`.

allauth/mfa/recovery_codes.py
```diff
@@ -79,7 +80,7 @@ def get_unused_codes(self):
             ret.append(code)
         return ret
 
-    def _validate_migrated_code(self, code):
+    def _validate_migrated_code(self, code: str) -> bool:
```
pennersr, line 83: **Issue**: Returns `None` as well.

pennersr (changes requested):

akx:
> Can you please rebase? Thanks.

Done, let's see if it passes! 🤞 

> adding type hints without enforcing them easily leads to errors.

Tangentially: I know we've talked about Ruff before and I know how you feel about it, but it was pretty useful here, as the [ANN series](https://docs.astral.sh/ruff/rules/#flake8-annotations-ann) of rules can narrow down places that are missing annotations, and a human can then figure them out.

pennersr:
See #3905 -- some of your changes are not Python 3.7 compatible, that will be addressed by #3905.

akx:
@pennersr On that note, what's your stance on using `from __future__ import annotations` and futuristic annotations (e.g. `list[Foo] | None`)? I don't think anything in allauth is expecting to introspect annotations (where unevaluated futured annotations could be an issue).

PR #3871, opened 2024-06-07
chore(mfa): make qrcode an optional dependency

akx:
## Why?

Users may wish to use an alternative QR code implementation such as https://pypi.org/project/segno/, or generate QR codes from TOTP URLs client-side, and don't need the extra dependencies.

## General

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -1,3 +1,12 @@
+Unreleased
+**********
+
+Note worthy changes
+-------------------
+
+- MFA functionality no longer has a hard dependency on `qrcode`;
+  install with the `qrcode` extra if you need QR code generation.
+
 0.63.3 (2024-05-31)
 *******************
 
--- a/allauth/mfa/totp.py
+++ b/allauth/mfa/totp.py
@@ -10,9 +10,6 @@ from urllib.parse import quote
 from django.core.cache import cache
 from django.utils.http import urlencode
 
-import qrcode
-from qrcode.image.svg import SvgPathImage
-
 from allauth.core import context
 from allauth.mfa import app_settings
 from allauth.mfa.models import Authenticator
@@ -74,6 +71,11 @@ def build_totp_url(label, issuer, secret):
 
 
 def build_totp_svg(url):
+    try:
+        import qrcode
+        from qrcode.image.svg import SvgPathImage
+    except ImportError as ie:
+        raise ImportError("The qrcode library is required to generate QR codes") from ie
     img = qrcode.make(url, image_factory=SvgPathImage)
     buf = BytesIO()
     img.save(buf)
--- a/docs/mfa/introduction.rst
+++ b/docs/mfa/introduction.rst
@@ -14,6 +14,10 @@ Note that in order to use this functionality you need to install the ``mfa`` ext
 
   pip install "django-allauth[mfa]"
 
+If you also wish to render QR codes using the `python-qrcode` library, add the `qrcode` extra:
+
+  pip install "django-allauth[mfa,qrcode]"
+
 Remember to add the app to
 the ``settings.py`` of your project::
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -51,9 +51,11 @@ install_requires =
 
 [options.extras_require]
 mfa =
-    qrcode >= 7.0.0
+    # dummy to avoid "missing extra" warnings
 openid =
     python3-openid >= 3.0.8
+qrcode =
+    qrcode >= 7.0.0
 saml =
     python3-saml>=1.15.0,<2.0.0
 steam =
--- a/tox.ini
+++ b/tox.ini
@@ -33,6 +33,7 @@ deps =
 extras =
     mfa
     openid
+    qrcode
     socialaccount
     steam
 commands =


pennersr:
This is a backwards incompatible change that breaks existing installations silently. 

How about we move that `build_totp_svg` into the MFA adapter instead? Then, just don't install allauth using the `[mfa]` extra and you are free to plugin your own rendering engine, *without* breaking backwards compatibility.

pennersr:
> There's precedent for this sort of thing...

True, and that one caused a lot of issues as well. There were a lot of complaints and bugs reported about installs where Google wouldn't work anymore because `[socialaccount]` was now all of a sudden needed (e.g. https://stackoverflow.com/questions/78453278/i-have-used-to-google-provider-in-social-auth-in-django-but-provider-is-not-sho).  Though, that  was about transitioning from always "installing the world" to at least install only what's needed depending on what `allauth.<subapp>` you are using -- that does add value. I do not think that it is worth breaking backwards compatibility over the `qrcode` package.

> That feels a bit unfortunate too. What if MFA functionality gets new actually mandatory deps later? Then you'd have installations that are not using the extra, and they'll break in an interesting way too.

But here, we would be breaking installations for people who consciously decided to deviate from standard allauth by not installing `[mfa]` and thereby willingly taking on the consequences of that choice. That is far less an issue (if one at all) compared to breaking installations for people who just follow the regular installation guides.

akx:
> There were a lot of complaints and bugs reported about installs where Google wouldn't work anymore [...]

Which is why we read changelogs when upgrading and docs when setting things up... 😅 

>  But here, we would be breaking installations for people who consciously decided to deviate from standard allauth by not installing `[mfa]`

But I'd _want to_ have the `[mfa]` extra for future-proofing... 😄

What if I added an early check in `allauth.mfa.views` for `qrcode` being installed? IOW, if you've hooked up the MFA frontend, it'd remind you to install `qrcode` if it's otherwise not installed?

akx:
Moving `qrcode` to a late import (like in this PR) + adding a (currently empty) extra `[mfa-min]` (and documenting it) sounds like a good deal to me?

mecampbellsoup:
Is Segno somehow preferable to qrcode?

mecampbellsoup:
> Hmm. As it is, `build_totp_svg` is only used by the MFA view (and as such is only a dependency for headed installations) – the adapter doesn't feel like the right place for it...

Just a heads up/possibly related... in [this recent PR I opened](https://github.com/pennersr/django-allauth/pull/3884), I added `totp_url` attribute to the headless MFA activate endpoint's response data. Maybe we should also include/add `totp_svg`...

pennersr:
Soon, MFA will support webauthn, and for that another dependency is needed: `fido2`. Options:

- `pip install django-allauth[mfa,webauthn]`
- Just `pip install django-allauth[mfa]`, and use `mfa-min` for a version without `webauthn`.

The trouble with micro managing the dependencies is that the test suite would need to cover various combinations. So it is TBD what is the best approach here, just have 2 MFA flavors: minimal, without webauthn and qrcode and regular including all, or ...

pennersr:
One of the issues is that `mfa-min` would not depend on anything.

> What if MFA functionality gets new actually mandatory deps later?

So where do you see `webauthn` in this picture? Is that something part of both `mfa` and `mfa-min`?

pennersr:
I must admit, I am leaning towards not picking this up. Rationale:

- Micromanaging dependencies would almost result in a 1:1 mapping of an `[extra]` matching some `pip install extra`. That does not make sense, because the only benefit would be a version pin, but other than that there is almost no difference with not installing `[mfa]` and cherry picking the dependencies yourself.

- Somehow grouping dependencies into bigger pieces is quite arbitrary. E.g. where does `[webauthn]` reside, likely not in `[min]`, but what if people don't need `[webauthn]`. 

All in all, this way working...

> How about we move that build_totp_svg into the MFA adapter instead? Then, just don't install allauth using the [mfa] extra and you are free to plugin your own rendering engine, without breaking backwards compatibility.

... is starting to look quite attractive.

> As it is, build_totp_svg is only used by the MFA view (and as such is only a dependency for headed installations) 

Currently, that is true, but it is not too far fetched for a mobile client to want to render an SVG as well.

akx:
> Is Segno somehow preferable to qrcode?

Zero dependency, more featureful (IMO) and actively maintained (c.f. https://github.com/lincolnloop/python-qrcode/commits/main/). I'm sure [my PRs to `qrcode` will get looked at any moment now!](https://github.com/lincolnloop/python-qrcode/pulls/akx) 😁 

> So where do you see `webauthn` in this picture? Is that something part of both `mfa` and `mfa-min`?

My favorite (which would imply changing what the `mfa` extra currently does) is

* `mfa`: MFA core – bring your own QR codes, no support for Webauthn
* `mfa` + `qrcode`: MFA with QR code generation support via `qrcode`, no support for Webauthn
* `mfa` + `qrcode` + `webauthn`: MFA with QR code generation support via `qrcode`, Webauthn supported

(and of course `mfa` + `webauthn` if someone needs that combo).

I would appreciate the `fido2` dependency ending up in its own extra in any case.

> The trouble with micro managing the dependencies is that the test suite would need to cover various combinations.

I'm not sure that's 100% necessary, so long as the scope of the optional dependencies is kept atomic, and e.g. late imports have comments near them noting that this is an optional dep, so if anyone was about to hoist them to top-level, they'd, well, not do that.

> Currently, that is true, but it is not too far fetched for a mobile client to want to render an SVG as well.

Speaking of which, a PNG actually sounds more attractive for a mobile client (e.g. React Native doesn't have, pardon the pun, native SVG support as far as I know).

pennersr:
> My favorite (which would imply changing what the mfa extra currently does) is

That would result in a 1:1 mapping from one extra to one package:
- `[mfa]`: MFA core has no dependencies -- so no `pip install`.
- `[qrcode]`: Is identical to `pip install qrcode`.
- `[webauthn]`: Is identical to `pip install fido2`.

So, I really don't see the benefit of doing that, as the end result is the same as not installing `[mfa]` and cherry picking the dependencies your self:

- `pip install django-allauth[mfa,qrcode]` is identical to `pip install django-allauth qrcode`. 
- `pip install django-allauth[mfa,webauthn]` is identical to `pip install django-allauth fido2`.

Sure, somewhere it needs to be documented that you need `fido2` for Passkey support, but that holds for `[webauthn]` as well. 

So there is no real benefit, and at the same time it breaks backwards compatibility. 

> PNG actually sounds more attractive for a mobile client

While RN doesn't have SVG support out of the box, there is a well maintained React Native SVG package out there that I have successfully used in the past.

Summarizing:
- Altering the `[extras]` is not going to fly.
- What is an option is to replace the use of `qrcode` with `segno`.
- And what's also worthwhile pursuing is moving the `build_totp_svg()` logic into the MFA adapter.

pennersr:
Closed via 18c7ca47

akx:
I'm fine with this resolution 😄 

Thanks for the discussion, and as always, thank you for the library and the effort you've put into it!

PR #3884, opened 2024-06-10
refactor(mfa): Add totp_url to headless mfa activate response data

mecampbellsoup:
# Description

Adds `totp_url` to `allauth.headless.mfa.response.TOTPNotFoundResponse`. The rationale is so that we get a ready-to-render TOTP URI for our frontend and I don't have to re-implement any of the `allauth.mfa.views.ActivateTOTPView.get_context_data` logic in my application or elsewhere/some other way.

# Submitting Pull Requests

## General

### DIFF
--- a/allauth/headless/mfa/response.py
+++ b/allauth/headless/mfa/response.py
@@ -38,11 +38,12 @@ class AuthenticatorDeletedResponse(APIResponse):
 
 
 class TOTPNotFoundResponse(APIResponse):
-    def __init__(self, request, secret):
+    def __init__(self, request, secret, totp_url):
         super().__init__(
             request,
             meta={
                 "secret": secret,
+                "totp_url": totp_url,
             },
             status=404,
         )
--- a/allauth/headless/mfa/views.py
+++ b/allauth/headless/mfa/views.py
@@ -10,6 +10,7 @@ from allauth.headless.mfa.inputs import (
     GenerateRecoveryCodesInput,
 )
 from allauth.mfa import totp
+from allauth.mfa.adapter import DefaultMFAAdapter, get_adapter
 from allauth.mfa.internal import flows
 from allauth.mfa.models import Authenticator
 from allauth.mfa.stages import AuthenticateStage
@@ -44,14 +45,17 @@ class AuthenticatorsView(AuthenticatedAPIView):
         return response.AuthenticatorsResponse(request, authenticators)
 
 
+# TODO: Enforce reauthentication
 class ManageTOTPView(AuthenticatedAPIView):
     input_class = {"POST": ActivateTOTPInput}
 
     def get(self, request, *args, **kwargs):
         authenticator = self._get_authenticator()
         if not authenticator:
+            adapter: DefaultMFAAdapter = get_adapter()
             secret = totp.get_totp_secret(regenerate=True)
-            return response.TOTPNotFoundResponse(request, secret)
+            totp_url: str = adapter.build_totp_url(request.user, secret)
+            return response.TOTPNotFoundResponse(request, secret, totp_url)
         return response.TOTPResponse(request, authenticator)
 
     def _get_authenticator(self):
--- a/docs/headless/openapi-specification/openapi.yaml
+++ b/docs/headless/openapi-specification/openapi.yaml
@@ -2458,8 +2458,14 @@ components:
                     description: |
                       A TOTP secret that can be used to setup a new authenticator.
                     example: J4ZKKXTK7NOVU7EPUVY23LCDV4T2QZYM
+                  totp_url:
+                    type: string
+                    description: |
+                      otpauth URI from which a QR code can be generated and scanned by OTP clients.
+                    example: otpauth://totp/Example:alice@fsf.org?secret=JBSWY3DPEHPK3PXP&issuer=Example
                 required:
                   - secret
+                  - totp_url
             required:
               - status
               - meta
--- a/allauth/headless/mfa/tests/test_views.py
+++ b/allauth/headless/mfa/tests/test_views.py
@@ -152,6 +152,7 @@ def test_get_totp_not_active(
     assert resp.status_code == 404
     data = resp.json()
     assert len(data["meta"]["secret"]) == 32
+    assert len(data["meta"]["totp_url"]) == 145
 
 
 def test_get_totp(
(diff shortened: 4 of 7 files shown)


allauth/headless/mfa/views.py
```diff
@@ -50,8 +48,14 @@ class ManageTOTPView(AuthenticatedAPIView):
...
         authenticator = self._get_authenticator()
         if not authenticator:
+            adapter = get_adapter()
             secret = totp.get_totp_secret(regenerate=True)
-            return response.TOTPNotFoundResponse(request, secret)
+            totp_url: str = totp.build_totp_url(
+                adapter.get_totp_label(request.user),
+                adapter.get_totp_issuer(),
+                secret,
+            )
```
mecampbellsoup, line 57: Building this TOTP URI follows the same pattern as in the [headed view](https://github.com/mecampbellsoup/django-allauth/blob/4ee8a1157ea55afd932441aff89ea2dff1821803/allauth/mfa/views.py#L112-L116).

allauth/headless/mfa/views.py
```diff
@@ -44,14 +45,21 @@ def get(self, request, *args, **kwargs):
         return response.AuthenticatorsResponse(request, authenticators)
 
 
+# TODO: Enforce reauthentication
```
mecampbellsoup, line 48: @pennersr BTW, I noticed that in headless mode, enabling MFA does not explicitly check the user has authenticated recently. Is that done intentionally? Is the idea that headless allauth applications should enforce that via a separate API call, i.e. to `headless:browser:mfa:reauthenticate`?

allauth/headless/mfa/response.py
```diff
@@ -38,11 +38,12 @@ class AuthenticatorDeletedResponse(APIResponse):
 
 
 class TOTPNotFoundResponse(APIResponse):
-    def __init__(self, request, secret):
+    def __init__(self, request, secret, totp_url):
         super().__init__(
             request,
             meta={
                 "secret": secret,
+                "totp_url": totp_url,
```
pennersr, line 46: This needs to be documented over here: 

https://docs.allauth.org/en/latest/headless/openapi-specification/#tag/Account:-2FA/paths/~1_allauth~1%7Bclient%7D~1v1~1account~1authenticators~1totp/get

allauth/headless/mfa/views.py
pennersr: How about we add `def build_totp_url(self, secret):` to the adapter method, then we don't have this duplication, and also allow for people to intervene and alter the URL as they see fit?

pennersr (changes requested):

allauth/headless/mfa/views.py
pennersr: When activating TOTP, headless invokes `flows.totp.activate_totp()`, which has `raise_if_reauthentication_required(request)`. Deleting TOTP, viewing and regenerating recovery codes is handled similarly. Doesn't that cover this?

mecampbellsoup:
@pennersr closed via https://github.com/pennersr/django-allauth/commit/18c7ca4701f8891e36808dae47bbba75de92af37 right?

allauth/headless/mfa/views.py
mecampbellsoup: Done in https://github.com/pennersr/django-allauth/commit/18c7ca4701f8891e36808dae47bbba75de92af37?

pennersr: The comment is about `build_topt_url` -- that is not in 18c7ca4.

mecampbellsoup:
@pennersr this one ready to go now I think! 🙏

allauth/headless/mfa/views.py
mecampbellsoup: Hmmm I will play around and confirm!

allauth/headless/mfa/views.py
```diff
@@ -44,14 +45,21 @@ def get(self, request, *args, **kwargs):
...
     def get(self, request, *args, **kwargs):
         authenticator = self._get_authenticator()
         if not authenticator:
+            adapter: DefaultMFAAdapter = get_adapter()
             secret = totp.get_totp_secret(regenerate=True)
-            return response.TOTPNotFoundResponse(request, secret)
+            totp_url: str = adapter.build_totp_url(
+                adapter.get_totp_label(request.user),
+                adapter.get_totp_issuer(),
+                secret,
```
pennersr, line 60: I think it would be good if we change the signature of `build_totp_url()` to just accept `user` and `secret`, because inside that method we can just call `self.get_totp_label()` and so on.

docs/headless/openapi-specification/openapi.yaml
```diff
@@ -2458,8 +2458,14 @@ components:
                     description: |
                       A TOTP secret that can be used to setup a new authenticator.
                     example: J4ZKKXTK7NOVU7EPUVY23LCDV4T2QZYM
+                  totp_url:
+                    type: string
+                    description: |
+                      URL-encoded QR code containing the TOTP secret and other metadata to be scanned by OTP clients.
```
pennersr, line 2464: This description seems off -- it is not a QR code right, nor is it URL encoded. It's just a URL/URI , from which the QR code could be generated.

docs/headless/openapi-specification/openapi.yaml
```diff
@@ -2458,8 +2458,14 @@ components:
...
+                    example: otpauth://totp/Example:alice@google.com?secret=JBSWY3DPEHPK3PXP&issuer=Example
```
pennersr, line 2465:
```suggestion
                    example: otpauth://totp/Example:alice@fsf.org?secret=JBSWY3DPEHPK3PXP&issuer=Example
```

docs/headless/openapi-specification/openapi.yaml
mecampbellsoup: Do you want to link to e.g. https://datatracker.ietf.org/doc/draft-linuxgemini-otpauth-uri/

docs/headless/openapi-specification/openapi.yaml
mecampbellsoup: Ha, good call 😄

mecampbellsoup:
> @pennersr this one ready to go now I think! 🙏

NOW it should be ready 😆

pennersr:
Thanks, merged via 86fbc457

Issue #420, opened 2013-10-17
OAuth2 Refresh feature?


pauricthelodger:
Hey there,

Have you considered adding a feature for refreshing access tokens with a given refresh token? Or do you think it's outside the scope of allauth?

Cheers,
Padraic

phildini:
Hey is there any more thought on this? I've been trying to auth against Google, but refresh tokens aren't saving into the token_secret.

arnaldoanez:
For me it's saving the Google refresh_token in the 'token_secret' field. However, the 'expires_at' seems to work just as a reference, and nothing is done with it. So I am thinking on refreshing the token right after it fails when trying to work with it. For this case I'd need to implement a signal-like method to refresh the token, updating it in SocialToken model and try request again. 

However, would that fresh new token get overwritten on a second login?

What would be the best way to handle this?

wtfrank:
I hacked together a function to acquire a new access token from the refresh token. Its not suitable for production, but I'm going to leave it here as a possible starting point for a full implementation or for someone who just wants to get the job done, hack or no hack!

`def _refresh_authorise(user, sat):
  old_access_token = sat.token
  refresh_token = sat.token_secret

  app = sat.app
  client_id = app.client_id
  api_key = app.secret
  data={
         "grant_type": "refresh_token",
         "refresh_token": refresh_token}

  auth = requests.auth.HTTPBasicAuth(
      client_id,
      api_key)

  url = "%s/token" % "https://login.eveonline.com/oauth"
  params = None

  resp = requests.request(
    'post',
    url,
    params=params,
    data=data,
    headers=None,
    auth=auth)


  if resp.status_code != 200:
    raise Exception("Unexpected code: %d" % res.status_code)


  res = resp.json()
  access_token = res['access_token']
  expires_in = res['expires_in']

  sat.token = access_token
  sat.expires_at = timezone.now() + timedelta(seconds=expires_in)
  sat.save()
  return
`

wli:
I think there's a couple different ways to do this, and they each handle the cases a little differently. There's two main things to determine whether allauth should support: known expired tokens (`expires_at` in `SocialApplicationToken`), and unknown expired tokens (when you attempt to make a query, and the token fails, thus you need to refresh it).

If we only need to support the known expired tokens, then we should just create a `.ensure_valid_token()` function on `OAuth2Adapter` instances. The standard implementation would probably look something like:

```python
def ensure_valid_token(token, force=False):
[...]

PR #3889, opened 2024-06-12
feat(oauth2): add method to handle access token refresh process

Ahuahuachi:
# Description

This PR adds a method to handle the process of refreshing an access token to the OAuth2 Adapter.
The purpose for this is to give developers an option to create custom workflows in their projects, and possibly create future middleware in allauth codebase to handle the most common use cases for this.

This was based on the solution provided by https://github.com/pennersr/django-allauth/issues/420#issuecomment-1684303097

# Submitting Pull Requests

## General

 - [x] ~JavaScript code should adhere to [StandardJS](https://standardjs.com).~
 - [x] ~If your change is substantial, feel free to add yourself to `AUTHORS`.~

### DIFF
--- a/ChangeLog.rst
+++ b/ChangeLog.rst
@@ -12,6 +12,8 @@ Note worthy changes
   include minor documented backwards incompatibilities. Always read the release
   notes before upgrading.
 
+- OAuth2 adapter: Added a method to handle the process of refreshing an access token intended
+  to be used in custom workflows or future attempts to create a common middleware
 
 0.63.3 (2024-05-31)
 *******************
--- a/allauth/socialaccount/providers/oauth2/views.py
+++ b/allauth/socialaccount/providers/oauth2/views.py
@@ -8,6 +8,8 @@ from django.shortcuts import render
 from django.urls import reverse
 from django.utils import timezone
 
+from requests_oauthlib import OAuth2Session
+
 from allauth.account import app_settings as account_settings
 from allauth.core.exceptions import ImmediateHttpResponse
 from allauth.core.internal.httpkit import add_query_params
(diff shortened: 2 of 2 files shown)


pennersr:
This adds a new method to the OAuth2Adapter that is not used by allauth anywhere, nor is it covered by unit tests. So, I assume this is meant to be called by developers integrating allauth, but then the OAuth2Adapter is not the most logical place for that. You also mention:

> possibly create future middleware...most common use cases for this

I think it is important to clear up what use case this is covering, and then it can be decided whether or not it is in scope and if so, what the best API would be to cover that. 

But in its current form, this is not a viable PR.

Ahuahuachi:
Yes, it is meant to be called by developers.
I think the use case is completely covered on #420 discussion where some projects are in need of a refresh feature. As per your comment on 2013:

> That's in scope, though I think it is best to integrate https://github.com/requests/requests-oauthlib to do the actual dirty work...

Please let me know if anything changed since

The common use case for the middleware I was thinking on, its when the site needs to have a valid access token on every page of the site. But I think this deserves a dedicated discussion when allauth has a refresh feature.

If this method being part of the OAuth2Adapter it's not logical, maybe I should try adding it to the OAuth2Client?

pennersr:
That comment from 2013 did not age very well. In all this time, there has been no use case to have up-to-date access tokens for authentication purposes. You can even argue that storing the `SocialToken` unnecessary increases allauth scope. 

> when the site needs to have a valid access token on every page of the site.

Why would you need that? Likely for accessing resources on the side of the provider (e.g. posts, pictures and so on), but that is really all application specific, and not needed for authentication purposes at all.

Ahuahuachi:
The thing that got me into working this was precisely the need to access the resources of the providers.

So with the state of the project right now, a refresh feature of some kind is out of scope then? If so, maybe consider closing the discussion for that feature.

pennersr:
@Ahuahuachi Fair enough, I've added a comment there...

PR #3907, opened 2024-06-20
Miscellaneous mechanical legacy cleanups

akx:

Inspired by #3905, this runs a couple of mechanical upgrades with Ruff's `target-version` set to Python 3.8. (None of these particular upgrades happen to be Python 3.8 specific, but it was inspiration still.)

I elided doing 128 "Use `super()` instead of `super(__class__, self)" changes in this PR, since I didn't feel like double-checking all of them in case there was a sneaky "skipping a superclass to use a super-superclass instead".

There's also a whole bunch of opportunities to use `.format()` instead of `%`, and further to use f-strings instead of `.format()`, but those too are elided from this PR.


pennersr:
> There's also a whole bunch of opportunities to use .format() instead of %, and further to use f-strings instead of .format(), but those too are elided from this PR.

As long as this would not negatively impact existing translations it would be okay...

PR #3909, opened 2024-06-21
added helper script.py, created a requirements.txt file and added myself as author

irvincardoza:
The script.py file makes downloading packages from requirements.txt easier. This script takes pip list packages as input, strips data and compares the packages downloaded to the required packages and tells us what is missing. Have also added a feature that adds a "pip install" before the package name and the version after so that the user can just copy paste the command into terminal and download missing packages.

output:

pip install {missing package} == {version number}

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -77,6 +77,7 @@ Hatem Nassrat
 Hyunwoo Shim
 Ian R-P
 Ignacio Ocampo
+Irvin Cardoza
 Illia Volochii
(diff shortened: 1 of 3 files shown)


pennersr:
Developers are supposed to install django-allauth using `pip install django-allauth[some,extras]`. I don't think we should start offering `requirements.txt` or tooling to somehow support that. Upstream projects may not even be using `requirements.txt`...

PR #3921, opened 2024-06-26
fix(socialaccount): Fix SocialLogin.is_existing always False when sig…

### LINKED ISSUE #3870: is_existing not updating after saving a new user connected to SocialLogin
I'm unsure of the fix, but if you save a new user connected to a sociallogin, call sociallogin.is_existing, it returns False when it should be True. Maybe after line 275 in [socialaccounts/models](https://github.com/pennersr/django-allauth/blob/f6577fa0ff0060f1debb4dfe1103d8484aaea69c/allauth/socialaccount/models.py#L275) set `self.user = user`. Or clear the _is_existing flag. This bug currently breaks a dj_rest_auth flow. 

Alternatively, you could revert [this]( https://github.com/pennersr/django-allauth/commit/2021e0be8b7819916a1585ad3708e9139e227516) commit. The extra complexity might not be worth it.

juchajam:
Fix SocialLogin.is_existing always False when signup.
fix #3870

# Submitting Pull Requests

## General

### DIFF
--- a/allauth/socialaccount/models.py
+++ b/allauth/socialaccount/models.py
@@ -310,8 +310,8 @@ class SocialLogin:
         """When `False`, this social login represents a temporary account, not
         yet backed by a database record.
         """
-        if self._is_existing is not None:
-            ret = self._is_existing
+        if self._is_existing:
+            return self._is_existing
         if self.user.pk is None:
             ret = False
         else:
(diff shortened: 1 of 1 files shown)


pennersr:
See: https://github.com/pennersr/django-allauth/issues/3870#issuecomment-2154221643

The idea behind that `is not None` is to cache the result. This is working perfectly fine, but apparently breaks when projects are completely bypassing the allauth pipelines, which they really should not be doing.

PR #3928, opened 2024-06-28
fix(accounts): Magic Code login should respect EMAIL_UNKNOWN_ACCOUNTS setting, Fixes #3927

### LINKED ISSUE #3927: Magic code login method sends email when ACCOUNT_EMAIL_UNKNOWN_ACCOUNTS is False
The relatively new Magic Code login mechanism appears to ignore the `ACCOUNT_EMAIL_UNKNOWN_ACCOUNTS` setting and always sends an 'unknown account' email to whatever email address you provide.

There also doesn't appear to be a check to see if signups are disabled, so you could use the magic code mechanism to spam people with invites to sign up to a site, even if signups aren't enabled for that site.

It appears that in `allauth/account/internal/flows/login_by_code.py` the block

```python
    if not users:
        send_unknown_account_mail(request, email)
```

should instead match the code in `allauth/account/forms.py`

```python
    if not self.users:
        if app_settings.EMAIL_UNKNOWN_ACCOUNTS:
            flows.signup.send_unknown_account_mail(request, email)
```

It's unclear to me why you would want to send a magic login link to an email that isn't linked to an account that already exists on the site. There might be a reason I'm unaware of. Shouldn't there be a check to see if signups are permitted first, though? Or is checking the `EMAIL_UNKNOWN_ACCOUNTS` setting sufficient?

jpwarren:
Adds a missing check of `settings.EMAIL_UNKNOWN_ACCOUNTS` before attempting to send an email to the address provided for a magic code login.

Fixes: #3927


allauth/account/internal/flows/login_by_code.py
```diff
@@ -28,7 +28,8 @@ def request_login_code(request: HttpRequest, email: str) -> None:
         "failed_attempts": 0,
     }
     if not users:
-        send_unknown_account_mail(request, email)
+        if app_settings.EMAIL_UNKNOWN_ACCOUNTS:
+            send_unknown_account_mail(request, email)
```
pennersr, line 32: Instead of duplicating this `if`, I think it would be better to fold this into the `send_unknown_account_mail` implementation itself.

jpwarren: Sounds good. Additional commit added to do this.

pennersr:
Thanks, merged via 62062968

Issue #3906, opened 2024-06-19
Page listing social connections should use email addresses or usernames, instead of names


Flimm:
Take the page at the URL `/accounts/3rdparty/` (URL name `socialaccount_connections` configured to `allauth.social.views.connections`). If you load that page, it will list all the social connections of the current account, like this:

- ( ) John Smith **Google**
- ( ) John Smith **Facebook**
- ( ) John Smith **Google**

It lists each account by displaying the name of the account, followed by the provider name. Unfortunately, the name of the account is often the first name and last name of the person associated with that account. A user may have multiple accounts with Google under the name "John Smith", and may not be able to discern which one is being referred to here. Even if the user has not connected all their Google accounts that have the name "John Smith", they still cannot tell which one of their Google accounts with the name "John Smith" is being referred to here. To put it succinctly, names are not unique identifiers for Google.

Instead, I suggest using an identifier that is unique to that provider, for example, an email address or a username, like this:

- ( ) **Google** account johnsmith1@gmail.com
- ( ) **Facebook** account johnsmithy
- ( ) **Google** account john.smith.abc@gmail.com

I propose that the method `get_identifier` be added to the `ProviderAccount` class:

```
class ProviderAccount(object):
    def get_identifier(self):
        """
        Returns a string representing an identifier that is uniquely identifies
        this account among the provider's accounts, preferably one that is
        meaningful to human beings. This is usually a username or an email address.
        """
        return None
```

I also propose modify the `to_str` method on that class so that it uses the `get_identifier` method, falling back to the previously implemented return value if `get_identifier` does not return a truthy value.

```
    def to_str(self):
        return self.get_identifier() or self.get_brand()["name"]
```

Then providers can override the `get_identifier` method to return an email address or username, and `to_str` will continue to work even for providers that haven't implemented the `get_identifier` method yet.

pennersr:
Thoughts:
- "representing an identifier that is uniquely identifies this account" -- this would only be usable by humans, as an email address and/or username could be modifiable over at the provider. Hence, this identifier is not to be used for other purposes...
- Given that it is just for display purposes, do we really need a new method, or could we just alter the `to_str()` method?
- You likely put `return None` there so that there is some default, as changing this for all providers is quite a job. But, that leaves if with  new method that sometimes gives a result, sometimes not... so that also points to not adding a new method and just altering `to_str()`.

Flimm:
[...]

PR #3940, opened 2024-07-03
Make to_str method return unique identifiers, not display name

Flimm:
This fixes this issue: https://github.com/pennersr/django-allauth/issues/3906

I added tests, so that the result of `to_str` is always tested.

### DIFF
--- a/allauth/socialaccount/providers/agave/provider.py
+++ b/allauth/socialaccount/providers/agave/provider.py
@@ -12,7 +12,7 @@ class AgaveAccount(ProviderAccount):
 
     def to_str(self):
         dflt = super(AgaveAccount, self).to_str()
-        return self.account.extra_data.get("name", dflt)
+        return self.account.extra_data.get("email", dflt)
 
 
 class AgaveProvider(OAuth2Provider):
--- a/allauth/socialaccount/providers/agave/provider.py
+++ b/allauth/socialaccount/providers/agave/provider.py
@@ -10,10 +10,6 @@ class AgaveAccount(ProviderAccount):
     def get_avatar_url(self):
         return self.account.extra_data.get("avatar_url", "dflt")
 
-    def to_str(self):
-        dflt = super(AgaveAccount, self).to_str()
-        return self.account.extra_data.get("email", dflt)
-
 
 class AgaveProvider(OAuth2Provider):
     id = "agave"
--- a/allauth/socialaccount/providers/amazon/provider.py
+++ b/allauth/socialaccount/providers/amazon/provider.py
@@ -4,8 +4,7 @@ from allauth.socialaccount.providers.oauth2.provider import OAuth2Provider
 
 
 class AmazonAccount(ProviderAccount):
-    def to_str(self):
-        return self.account.extra_data.get("email", super().to_str())
+    pass
 
 
 class AmazonProvider(OAuth2Provider):
--- a/allauth/socialaccount/providers/amazon/provider.py
+++ b/allauth/socialaccount/providers/amazon/provider.py
@@ -5,7 +5,7 @@ from allauth.socialaccount.providers.oauth2.provider import OAuth2Provider
 
 class AmazonAccount(ProviderAccount):
     def to_str(self):
-        return self.account.extra_data.get("name", super(AmazonAccount, self).to_str())
+        return self.account.extra_data.get("email", super().to_str())
 
 
 class AmazonProvider(OAuth2Provider):
--- a/allauth/socialaccount/providers/amazon_cognito/provider.py
+++ b/allauth/socialaccount/providers/amazon_cognito/provider.py
@@ -57,7 +57,10 @@ class AmazonCognitoProvider(OAuth2Provider):
         return {
             "address": data.get("address"),
             "birthdate": data.get("birthdate"),
+            "email": data.get("email"),
+            "family_name": data.get("family_name"),
             "gender": data.get("gender"),
+            "given_name": data.get("given_name"),
             "locale": data.get("locale"),
             "middlename": data.get("middlename"),
             "nickname": data.get("nickname"),
--- a/allauth/socialaccount/providers/amazon_cognito/provider.py
+++ b/allauth/socialaccount/providers/amazon_cognito/provider.py
@@ -12,8 +12,9 @@ from allauth.socialaccount.providers.oauth2.provider import OAuth2Provider
 class AmazonCognitoAccount(ProviderAccount):
     def to_str(self):
         dflt = super(AmazonCognitoAccount, self).to_str()
-
-        return self.account.extra_data.get("username", dflt)
+        username = self.account.extra_data.get("username")
+        email = self.account.extra_data.get("email")
+        return email or username or dflt
 
     def get_avatar_url(self):
         return self.account.extra_data.get("picture")
--- a/allauth/socialaccount/providers/amazon_cognito/provider.py
+++ b/allauth/socialaccount/providers/amazon_cognito/provider.py
@@ -49,26 +49,15 @@ class AmazonCognitoProvider(OAuth2Provider):
         )
 
     def extract_extra_data(self, data):
-        return {
-            "address": data.get("address"),
-            "birthdate": data.get("birthdate"),
-            "email": data.get("email"),
-            "family_name": data.get("family_name"),
-            "gender": data.get("gender"),
-            "given_name": data.get("given_name"),
-            "locale": data.get("locale"),
-            "middlename": data.get("middlename"),
-            "nickname": data.get("nickname"),
-            "phone_number": data.get("phone_number"),
-            "phone_number_verified": convert_to_python_bool_if_value_is_json_string_bool(
-                data.get("phone_number_verified")
-            ),
-            "picture": data.get("picture"),
-            "preferred_username": data.get("preferred_username"),
-            "profile": data.get("profile"),
-            "website": data.get("website"),
-            "zoneinfo": data.get("zoneinfo"),
-        }
+        ret = dict(data)
+        phone_number_verified = data.get("phone_number_verified")
+        if phone_number_verified is not None:
+            ret["phone_number_verified"] = (
+                convert_to_python_bool_if_value_is_json_string_bool(
+                    "phone_number_verified"
+                )
+            )
+        return ret
 
     @classmethod
     def get_slug(cls):
--- a/allauth/socialaccount/providers/amazon_cognito/provider.py
+++ b/allauth/socialaccount/providers/amazon_cognito/provider.py
@@ -10,12 +10,6 @@ from allauth.socialaccount.providers.oauth2.provider import OAuth2Provider
 
 
 class AmazonCognitoAccount(ProviderAccount):
-    def to_str(self):
-        dflt = super(AmazonCognitoAccount, self).to_str()
-        username = self.account.extra_data.get("username")
-        email = self.account.extra_data.get("email")
-        return email or username or dflt
-
     def get_avatar_url(self):
         return self.account.extra_data.get("picture")
 
--- a/allauth/socialaccount/providers/angellist/provider.py
+++ b/allauth/socialaccount/providers/angellist/provider.py
@@ -12,10 +12,6 @@ class AngelListAccount(ProviderAccount):
     def get_avatar_url(self):
         return self.account.extra_data.get("image")
 
-    def to_str(self):
-        dflt = super(AngelListAccount, self).to_str()
-        return self.account.extra_data.get("email", dflt)
-
 
 class AngelListProvider(OAuth2Provider):
     id = "angellist"
--- a/allauth/socialaccount/providers/angellist/provider.py
+++ b/allauth/socialaccount/providers/angellist/provider.py
@@ -14,7 +14,7 @@ class AngelListAccount(ProviderAccount):
 
     def to_str(self):
         dflt = super(AngelListAccount, self).to_str()
-        return self.account.extra_data.get("name", dflt)
+        return self.account.extra_data.get("email", dflt)
 
 
 class AngelListProvider(OAuth2Provider):
--- a/allauth/socialaccount/providers/apple/provider.py
+++ b/allauth/socialaccount/providers/apple/provider.py
@@ -9,10 +9,23 @@ from allauth.socialaccount.providers.oauth2.client import OAuth2Error
 from allauth.socialaccount.providers.oauth2.provider import OAuth2Provider
 
 
+class AppleAccount(ProviderAccount):
+    def to_str(self):
+        dflt = super().to_str()
+        email = self.account.extra_data.get("email")
+        name = self.account.extra_data.get("name", {})
+        if name and (name.get("firstName") or name.get("lastName")):
+            full_name = " ".join([name.get("firstName", ""), name.get("lastName", "")])
+            full_name = full_name.strip()
+            if full_name:
+                return f"{full_name} <{email or dflt}>"
+        return email or dflt
+
+
 class AppleProvider(OAuth2Provider):
     id = "apple"
     name = "Apple"
-    account_class = ProviderAccount
+    account_class = AppleAccount
     oauth2_adapter_class = AppleOAuth2Adapter
     supports_token_authentication = True
 
--- a/allauth/socialaccount/providers/apple/provider.py
+++ b/allauth/socialaccount/providers/apple/provider.py
@@ -11,15 +11,18 @@ from allauth.socialaccount.providers.oauth2.provider import OAuth2Provider
 
 class AppleAccount(ProviderAccount):
     def to_str(self):
-        dflt = super().to_str()
         email = self.account.extra_data.get("email")
-        name = self.account.extra_data.get("name", {})
-        if name and (name.get("firstName") or name.get("lastName")):
-            full_name = " ".join([name.get("firstName", ""), name.get("lastName", "")])
+        if email and not email.lower().endswith("@privaterelay.appleid.com"):
+            return email
+
+        name = self.account.extra_data.get("name") or {}
+        if name.get("firstName") or name.get("lastName"):
+            full_name = f"{name['firstName'] or ''} {name['lastName'] or ''}"
             full_name = full_name.strip()
             if full_name:
-                return f"{full_name} <{email or dflt}>"
-        return email or dflt
+                return full_name
+
+        return super().to_str()
 
 
 class AppleProvider(OAuth2Provider):
--- a/allauth/socialaccount/providers/auth0/provider.py
+++ b/allauth/socialaccount/providers/auth0/provider.py
@@ -7,13 +7,6 @@ class Auth0Account(ProviderAccount):
     def get_avatar_url(self):
         return self.account.extra_data.get("picture")
 
-    def to_str(self):
-        dflt = super(Auth0Account, self).to_str()
-        email = self.account.extra_data.get("email")
-        username = self.account.extra_data.get("username")
-        name = self.account.extra_data.get("name")
-        return email or username or name or dflt
-
 
 class Auth0Provider(OAuth2Provider):
     id = "auth0"
--- a/allauth/socialaccount/providers/auth0/provider.py
+++ b/allauth/socialaccount/providers/auth0/provider.py
@@ -9,7 +9,10 @@ class Auth0Account(ProviderAccount):
 
     def to_str(self):
         dflt = super(Auth0Account, self).to_str()
-        return self.account.extra_data.get("name", dflt)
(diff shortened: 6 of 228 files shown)


allauth/socialaccount/providers/agave/provider.py
```diff
@@ -12,7 +12,7 @@ def get_avatar_url(self):
 
     def to_str(self):
         dflt = super(AgaveAccount, self).to_str()
-        return self.account.extra_data.get("name", dflt)
+        return self.account.extra_data.get("email", dflt)
```
pennersr, line 15: This change assumes `email` is always present, which need not be the case. What if the scope/consent is such that this data is not present?

allauth/socialaccount/providers/pinterest/provider.py
```diff
@@ -20,8 +20,14 @@ def get_avatar_url(self):
...
     def to_str(self):
-        dflt = super(PinterestAccount, self).to_str()
-        return self.account.extra_data.get("username", dflt)
+        username = self.account.extra_data.get("username")
+        if username:
+            return username
+        first_name = self.account.extra_data.get("first_name")
+        last_name = self.account.extra_data.get("last_name")
+        if first_name or last_name:
+            return " ".join(x for x in [first_name, last_name] if x)
```
pennersr, line 29:
```suggestion
            return f"{first_name or ''} {last_name or ''}".strip()
```

allauth/socialaccount/providers/amazon/provider.py
```diff
@@ -5,7 +5,7 @@
 
 class AmazonAccount(ProviderAccount):
     def to_str(self):
-        return self.account.extra_data.get("name", super(AmazonAccount, self).to_str())
+        return self.account.extra_data.get("email", super().to_str())
```
pennersr, line 8: As above.

allauth/socialaccount/providers/angellist/provider.py
```diff
@@ -14,7 +14,7 @@ def get_avatar_url(self):
 
     def to_str(self):
         dflt = super(AngelListAccount, self).to_str()
-        return self.account.extra_data.get("name", dflt)
+        return self.account.extra_data.get("email", dflt)
```
pennersr, line 17: As above.

allauth/socialaccount/providers/apple/provider.py
```diff
@@ -9,10 +9,23 @@
...
+class AppleAccount(ProviderAccount):
+    def to_str(self):
+        dflt = super().to_str()
+        email = self.account.extra_data.get("email")
+        name = self.account.extra_data.get("name", {})
+        if name and (name.get("firstName") or name.get("lastName")):
+            full_name = " ".join([name.get("firstName", ""), name.get("lastName", "")])
+            full_name = full_name.strip()
+            if full_name:
+                return f"{full_name} <{email or dflt}>"
```
pennersr, line 21: I am not sure the `<...>` approach is appropriate here. This is inconsistent compared to the other providers. Also, using `<dflt>` is perhaps unexpected.

allauth/socialaccount/providers/auth0/provider.py
```diff
@@ -9,7 +9,10 @@ def get_avatar_url(self):
 
     def to_str(self):
         dflt = super(Auth0Account, self).to_str()
-        return self.account.extra_data.get("name", dflt)
+        email = self.account.extra_data.get("email")
+        username = self.account.extra_data.get("username")
+        name = self.account.extra_data.get("name")
+        return email or username or name or dflt
```
pennersr, line 15: Thought: all these `to_str()` are almost the same. Given that the world is moving towards a standardized OIDC payload, we could also fold one implementation into a new to be created `OAuth2ProviderAccount`, and inside unify the logic?

pennersr (changes requested):

allauth/socialaccount/providers/agave/provider.py
Flimm: The parent class's `to_str` implementation was changed as well, to retrieve the identifier from `email`, `username` and `name`. So, in this case, if `email` is not found or falsey, then the `dflt` value will probably be taken from the `name` value. I suppose I could have deleted this `to_str` method altogether and just relied on the parent class's `to_str` implementation.

allauth/socialaccount/providers/amazon/provider.py
Flimm: If `email` is falsey, then `super().to_str()` will probably grab the `name` value.

allauth/socialaccount/providers/angellist/provider.py
Flimm: Same here. The `dflt` value will probably be from the `name` value, since the parent class's `to_str` implementation was changed.

allauth/socialaccount/providers/apple/provider.py
Flimm: The email address provided by Apple may be a masked email address ("Hide My Email" feature), and so the user may not recognise the email address, or know which Apple account it belongs to. I'm not sure what the best identifier for an Apple account is.

allauth/socialaccount/providers/auth0/provider.py
Flimm: I changed `to_str` in the base class `ProviderAccount` to look at `email`, `username` and `name`. If you think it is better to create a new class `OAuth2ProviderAccount`, we can do that.

allauth/socialaccount/providers/amazon/provider.py
pennersr: But here as well, `email` is already being returned by the super, so no need to provide a `to_str` anymore.

allauth/socialaccount/providers/apple/provider.py
pennersr: The email is masked if it ends with `@privaterelay.appleid.com`.  So we could check if the email is masked, and if not, return that as is, otherwise, return just the full name (as the email does not convey any information at all). Then, we wouldn't need the `<...>` construct.

allauth/socialaccount/providers/auth0/provider.py
pennersr: Ok, this is fine, but I do think that many of the `to_str()` should be revisitted, as most of them likely became superfluous.

pennersr (changes requested):

allauth/socialaccount/providers/auth0/provider.py
Flimm: We could delete most of the `to_str` implementations, but I don't think we should leave them as is, as many of them return the full name of the user, rather than a unique identifier (like a username or email address).

pennersr: That is what I meant -- delete them, unless they are really doing something specific (as is the case with Apple).

Flimm:
As requested, I deleted the superfluous `to_str` methods in subclasses. I pushed the new commit to this pull request.

Flimm:
I've resolved the other comments with the last two commits that I pushed to this pull request.

allauth/socialaccount/providers/amazon_cognito/provider.py
```diff
@@ -57,7 +52,10 @@ def extract_extra_data(self, data):
         return {
             "address": data.get("address"),
             "birthdate": data.get("birthdate"),
+            "email": data.get("email"),
```
pennersr, line 55: This method was not introduced in this PR, but it does not make any sense. I think we need to replace it with just this:

```
    def extract_extra_data(self, data):
        ret = dict(data)
        phone_number_verified = data.get("phone_number_verified")
        if phone_number_verified is not None:
            ret["phone_number_verified"] = (
                convert_to_python_bool_if_value_is_json_string_bool(
                    "phone_number_verified"
                )
            )
        return ret
```

allauth/socialaccount/providers/base/provider.py
```diff
@@ -261,4 +273,10 @@ def __str__(self):
         So we have this method `to_str` that can be overridden in a conventional
         fashion, without having to worry about it.
         """
-        return self.get_brand()["name"]
+        ed = self.account.extra_data
+        return (
```
pennersr, line 277: Given that this method is the default implementation, and that `extra_data` is completely unstructured, I think we need to be a bit cautious here as to what data we can expect:

- Add a `if isinstance(ed, dict)` before invoking `ed.get()`.
- When doing `ed.get("email")`, check if the result is really a `str`.

allauth/socialaccount/providers/daum/provider.py
```diff
@@ -8,7 +8,7 @@ def get_avatar_url(self):
         return self.account.extra_data.get("bigImagePath")
 
     def to_str(self):
-        return self.account.extra_data.get("nickname", self.account.uid)
+        return self.account.extra_data.get("userid", self.account.uid)
```
pennersr, line 11: Given that `to_str()` is supposed o return human recognizable identifiers, I don't think `userid` nor `uid` is the best choice here. Why not keep `nickname`?

allauth/socialaccount/providers/douban/provider.py
```diff
@@ -12,7 +12,7 @@ def get_avatar_url(self):
 
     def to_str(self):
         dflt = super(DoubanAccount, self).to_str()
-        return self.account.extra_data.get("name", dflt)
+        return self.account.extra_data.get("uid", dflt)
```
pennersr, line 15: As above, why not keep `name` ?

allauth/socialaccount/providers/edmodo/provider.py
```diff
@@ -32,9 +32,13 @@ def extract_common_fields(self, data):
 
     def extract_extra_data(self, data):
         return dict(
-            user_type=data.get("type"),
-            profile_url=data.get("url"),
             avatar_url=data.get("avatars").get("large"),
+            email=data.get("email"),
```
pennersr, line 36: Can we make this? Similar to Cognito, I prefer to store the raw data as is, instead of creating our own abstraction:

``` python
    def extract_extra_data(self, data):
        ret = dict(data)
        # NOTE: For backwards compatibility
        user_type = data.get("type")
        if user_type is not None:
            ret["user_type"] = user_type
        profile_url = data.get("url")
        if profile_url:
            ret["profile_url"] = profile_url
        avatar_url = data.get("avatars", {}).get("large")
        if avatar_url:
            ret["avatar_url"] = avatar_url
        # (end NOTE)
        return ret
```

allauth/socialaccount/providers/eventbrite/provider.py
```diff
@@ -15,6 +15,12 @@ def get_avatar_url(self):
         """Return avatar url."""
         return self.account.extra_data["image_id"]
 
+    def to_str(self):
+        for email in self.account.extra_data.get("emails", []):
+            if email.get("verified") and email.get("email"):
```
pennersr, line 20: For display purposes I think it's fine to fallback to a non-verified email if no verified email exists.

allauth/socialaccount/providers/gitlab/provider.py
```diff
@@ -12,7 +12,7 @@ def get_avatar_url(self):
 
     def to_str(self):
         dflt = super(GitLabAccount, self).to_str()
-        return self.account.extra_data.get("name", dflt)
+        return self.account.extra_data.get("username", dflt)
```
pennersr, line 15: `username` is already part of the base class, so I would expect this to be superfluous?

allauth/socialaccount/providers/kakao/provider.py
```diff
@@ -20,7 +20,11 @@ def get_avatar_url(self):
 
     def to_str(self):
         dflt = super(KakaoAccount, self).to_str()
-        return self.profile.get("nickname", self.properties.get("nickname", dflt))
+        account_data = self.account.extra_data.get("kakao_account", {})
+        if account_data.get("email") and account_data.get("is_email_verified"):
+            return account_data["email"]
```
pennersr, line 25: For display purposes, an unverified email is fine.

allauth/socialaccount/providers/openid_connect/provider.py
```diff
@@ -12,7 +12,9 @@
 class OpenIDConnectProviderAccount(ProviderAccount):
     def to_str(self):
         dflt = super(OpenIDConnectProviderAccount, self).to_str()
-        return self.account.extra_data.get("name", dflt)
+        user_id = self.account.extra_data.get("user_id")
```
pennersr, line 15: Why prefer `user_id` -- I don't expect these to be generally recognizable for OIDC providers?

allauth/socialaccount/providers/openstreetmap/provider.py
```diff
@@ -16,10 +16,11 @@ def get_avatar_url(self):
         return self.account.extra_data.get("avatar")
 
     def get_username(self):
+        # TODO: this actually returns the display name, not the unique username
```
pennersr, line 19: I don't think there is anything that comes closer, given this example payload:

``` json
{
 "version": "0.6",
 "generator": "OpenStreetMap server",
 "user": {"id": 12023, "display_name": "jbpbis", "account_created": "2007-08-16T01:35:56Z", "description": "", "contributor_terms": {"agreed": False}, "roles": [], "changesets": {"count": 1}, "traces": {"count": 0}, "blocks": {"received": {"count": 0, "active": 0}}}
}
```

See: https://wiki.openstreetmap.org/wiki/API_v0.6

allauth/socialaccount/providers/openstreetmap/provider.py
```diff
@@ -16,10 +16,11 @@ def get_avatar_url(self):
...
         return self.account.extra_data["display_name"]
 
     def to_str(self):
-        return self.get_username()
+        return self.account.extra_data.get("display_name") or super().to_str()
```
pennersr, line 23:
```suggestion
        return self.get_username() or super().to_str()
```

allauth/socialaccount/providers/pinterest/provider.py
```diff
@@ -20,8 +20,14 @@ def get_avatar_url(self):
...
-        dflt = super(PinterestAccount, self).to_str()
-        return self.account.extra_data.get("username", dflt)
+        username = self.account.extra_data.get("username")
+        if username:
+            return username
+        first_name = self.account.extra_data.get("first_name")
+        last_name = self.account.extra_data.get("last_name")
+        if first_name or last_name:
+            return f"{first_name or ''} {last_name or ''}".strip()
+        return super().to_str()
```
pennersr, line 30: Can't we rely on the base `to_str()` here? The above logic is not Pinterest specific at all.

allauth/socialaccount/providers/quickbooks/provider.py
```diff
@@ -12,6 +12,10 @@
 class QuickBooksAccount(ProviderAccount):
     def to_str(self):
         dflt = super(QuickBooksAccount, self).to_str()
+        email = self.account.extra_data.get("email")
+        email_verified = self.account.extra_data.get("emailVerified")
```
pennersr, line 16: Checking for `emailVerified` is not needed.

allauth/socialaccount/providers/soundcloud/provider.py
```diff
@@ -16,7 +16,7 @@ def to_str(self):
         dflt = super(SoundCloudAccount, self).to_str()
         full_name = self.account.extra_data.get("full_name")
         username = self.account.extra_data.get("username")
-        return full_name or username or dflt
+        return username or full_name or dflt
```
pennersr, line 19: This could be folded into the base `to_str()`.

allauth/socialaccount/providers/vk/provider.py
```diff
@@ -20,10 +20,11 @@ def get_avatar_url(self):
             return ret
 
     def to_str(self):
+        email = self.account.extra_data.get("email")
         first_name = self.account.extra_data.get("first_name", "")
         last_name = self.account.extra_data.get("last_name", "")
         name = " ".join([first_name, last_name]).strip()
-        return name or super(VKAccount, self).to_str()
+        return email or name or super(VKAccount, self).to_str()
```
pennersr, line 27: I think we can drop this -- it's not vk specific.

allauth/socialaccount/providers/windowslive/provider.py
```diff
@@ -7,6 +7,9 @@
 
 class WindowsLiveAccount(ProviderAccount):
     def to_str(self):
+        email = self.account.extra_data.get("emails", {}).get("preferred")
+        if email:
+            return email
         name = "{0} {1}".format(
             self.account.extra_data.get("first_name", ""),
             self.account.extra_data.get("last_name", ""),
```
pennersr, line 15: This `name = "{0} ..."` is not WindowsLive specific, we could rely on the base class here.

allauth/socialaccount/providers/yandex/provider.py
```diff
@@ -7,6 +7,9 @@
 
 class YandexAccout(ProviderAccount):
     def to_str(self):
+        default_email = self.account.extra_data.get("default_email")
+        if default_email:
+            return default_email
         first_name = self.account.extra_data.get("first_name", "")
         last_name = self.account.extra_data.get("last_name", "")
         name = " ".join([first_name, last_name]).strip()
```
pennersr, line 15: This is not Yandex specific, just rely on the base implementation.

pennersr (changes requested):

allauth/socialaccount/providers/daum/provider.py
Flimm: `userid` is a string with alphanumeric characters, according to the `tests.py` file. Nicknames are not guaranteed to be unique. That's why I changed it to `userid`. I haven't tested this with a real Daum account.

allauth/socialaccount/providers/douban/provider.py
Flimm: It looked like `name` might not be unique, but the alphanumeric `uid` might be unique and recognisable as a username. I haven't tested this with a real Douban account.

allauth/socialaccount/providers/base/provider.py
Flimm: Very good point.

allauth/socialaccount/providers/edmodo/provider.py
Flimm: Thanks. I just committed code similar to this. For backwards compatibility, I always set the `user_type`, `profile_url` and `avatar_url` keys, even if their values are `None`.

allauth/socialaccount/providers/soundcloud/provider.py
Flimm: The base `to_str` does not look at `full_name`, which may be needed here as a fallback.

allauth/socialaccount/providers/windowslive/provider.py
Flimm: The base `to_str` does not look at `first_name` and `last_name`, are you saying that it should?

allauth/socialaccount/providers/yandex/provider.py
Flimm: The base `to_str` does not look at `default_email`, are you saying it should?

pennersr: Yes, why not? Why add `full_name`, `first/last_name` lookups to a few providers requiring a custom provider specific `to_str()`  -- why not do this globally?

allauth/socialaccount/providers/quickbooks/provider.py
Flimm: Done.

allauth/socialaccount/providers/amazon_cognito/provider.py
Flimm: Done.

allauth/socialaccount/providers/yandex/provider.py
pennersr: Correction: not `default_email`, but definitely `first_name` and `last_name`.

pennersr: So to be precise: `if default_email` then return that, else just invoke `to_str()` which should take care of `first/last_name`.

Flimm: Done.

pennersr:
Thanks for preparing this. I've made a few tweaks and merged this as c76e9fbd:

- Prefer username above email when available: as the goal is to display something unique for human identifiable purposes, the `username` makes more sense (especially for sites like Twitter, GitHub, Bitbucket and so on...).
- Don't expose `uid` or `user_id` -- I don't expect these to be human identifiable.
- Move more into the base class.

Flimm:
Brilliant! Thank you so much!

PR #3946, opened 2024-07-05
Render checkboxes and radio buttons before label

Flimm:
The `<label>` element should be placed before its corresponding <input> in most cases. However, in the case of a checkbox input or a radio input, the <label> element should be placed after it.

To illustrate:

Bad:

    I agree to terms and conditions ☑

Good:

    ☑ I agree to terms and conditions

django-allauth ships with the two templates that will be affected positively by this change:

allauth/templates/account/email.html
allauth/templates/socialaccount/connections.html


Flimm:
Thanks for merging this!

PR #3948, opened 2024-07-05
fix: Add first_name and last_name to openid_connect provider

nicodier:
This replaces name with first_name and last_name. Which is used more often.

https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims

### DIFF
--- a/allauth/socialaccount/providers/openid_connect/provider.py
+++ b/allauth/socialaccount/providers/openid_connect/provider.py
@@ -64,9 +64,10 @@ class OpenIDConnectProvider(OAuth2Provider):
         return dict(
             email=data.get("email"),
             username=data.get("preferred_username"),
-            name=data.get("name"),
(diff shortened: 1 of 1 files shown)


allauth/socialaccount/providers/openid_connect/provider.py
```diff
@@ -64,9 +64,10 @@ def extract_common_fields(self, data):
         return dict(
             email=data.get("email"),
             username=data.get("preferred_username"),
-            name=data.get("name"),
```
pennersr, line 67: `name` is actually used in `adapter.populate_user()` as a fallback in case `first_name` / `last_name` are empty. Given that some OpenID Connect providers may only use `name`, and some only `first_name` / `last_name`, I think it's best to keep this line?

pennersr:
Merged via 6007045b

PR #3952, opened 2024-07-08
refactor(saml): Decouple build_saml_config from HttpRequest

mecampbellsoup:
Instead of coupling to `request.build_absolute_uri`, that is injected now by the caller. This allows `build_saml_config` to be used outside of a Django-specific request lifecycle context.

### DIFF
--- a/allauth/socialaccount/providers/saml/tests.py
+++ b/allauth/socialaccount/providers/saml/tests.py
@@ -186,7 +186,8 @@ def test_sls(auth_client, db, saml_settings, user_factory, sls_saml_request):
 )
 def test_build_saml_config_without_metadata_url(rf, provider_config):
     request = rf.get("/")
-    config = build_saml_config(request, provider_config, "org")
+    hostname = request.build_absolute_uri()
+    config = build_saml_config(hostname, provider_config, "org")
     assert config["idp"]["entityId"] == "dummy"
     assert config["idp"]["x509cert"] == "cert"
     assert config["idp"]["singleSignOnService"] == {"url": "https://idp.org/sso/"}
@@ -224,7 +225,8 @@ def test_build_saml_config(rf, provider_config):
                 "x509cert": "cert",
             }
         }
-        config = build_saml_config(request, provider_config, "org")
+        hostname = request.build_absolute_uri()
+        config = build_saml_config(hostname, provider_config, "org")
 
     assert config["idp"]["entityId"] == "dummy"
     assert config["idp"]["x509cert"] == "cert"
--- a/allauth/socialaccount/providers/saml/utils.py
+++ b/allauth/socialaccount/providers/saml/utils.py
@@ -2,7 +2,7 @@ from urllib.parse import urlparse
 
 from django.core.cache import cache
 from django.core.exceptions import ImproperlyConfigured
-from django.http import Http404
+from django.http import Http404, HttpRequest
 from django.urls import reverse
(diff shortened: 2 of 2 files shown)


mecampbellsoup:
@pennersr not sure how to request your review...

pennersr:
Can you elaborate a bit what use cases you have for this:

> ... outside of a Django-specific request lifecycle context.

The problem with using just the hostname is that this completely bypasses important additional logic, such as adding the right protocol http/https, including support for `settings.SECURE_PROXY_SSL_HEADER`. Given that this breaks that logic I would prefer to avoid adding this.

mecampbellsoup:
> Can you elaborate a bit what use cases you have for this:
> 
> > ... outside of a Django-specific request lifecycle context.

Sure! We are basically doing CRUD of SAML configurations. So I was hoping to call `build_saml_config` because it looks like it does some nice validations of the `provider.app.settings` (i.e. the SAML configuration) to sort of validate input SAML config from our users.

Maybe validations isn't the word, but for example it [fetches metadata from a given metadata URL](https://github.com/mecampbellsoup/django-allauth/blob/2e2e6ad6e5ebab13ad1b4c815584ab7df0051a9c/allauth/socialaccount/providers/saml/utils.py#L138).

pennersr:
In that case, you can always add a helper method that takes the host name and returns a dummy Django `HttpRequest` instance based on that. Or, use Django's `RequestFactory`. But do note that even when doing that the remarks about the protocol remain. Given that there is a workaround I am closing this.

mecampbellsoup:
> In that case, you can always add a helper method that takes the host name and returns a dummy Django `HttpRequest` instance based on that. Or, use Django's `RequestFactory`. But do note that even when doing that the remarks about the protocol remain. Given that there is a workaround I am closing this.

Not clear what the workaround is, I'd be reluctant to use `test.RequestFactory` in production code...

I suppose the "right" way to do this would be to extend allauth to have a `CreateSAMLConfiguration` view of some sort, so that our GQL app is not doing the DB CRUD but rather the Allauth view is.

pennersr:
The workaround is:

``` python
def hostname_to_request(hostname):
   from django.http import HttpRequest
   return HttpRequest(...implement_this...)

dummy_request = hostname_to_request("yourhost.com")
build_saml_config(dummy_request, ...)
```

mecampbellsoup:
> I suppose a better way to do this would be to extend allauth to have a `CreateSAMLConfiguration` view of some sort, so that our GQL app is not doing the DB CRUD but rather the Allauth view is.

@pennersr did you have any thoughts/reactions to this idea BTW?

pennersr:
"CRUD of SAML configurations" is a feature that goes beyond the current scope of allauth, I don't think it is wise to add that. But assuming being able to call that method without having a request is what you need, then that `hostname_to_request(hostname)` method really should do the trick.

mecampbellsoup:
> "CRUD of SAML configurations" is a feature that goes beyond the current scope of allauth, I don't think it is wise to add that. But assuming being able to call that method without having a request is what you need, then that `hostname_to_request(hostname)` method really should do the trick.

Yea was mostly wondering if you felt it made sense for allauth to expose a way (not via configuration i.e. `SOCIALACCOUNT_PROVIDERS`) to manage `SocialApp` via the API. I imagine it may not be terribly uncommon for situations like ours where we're using allauth in a multi-tenant application, and each tenant may want to bring their own SAML configuration(s).

PR #3973, opened 2024-07-16
chore(templates): Cleanup unnecessary template content

brianhelba:


allauth/socialaccount/providers/dummy/templates/dummy/authenticate_form.html
```diff
@@ -6,7 +6,6 @@
         Dummy Provider Login
     {% endelement %}
     {% element form form=form method="post" action=action_url %}
-        {% csrf_token %}
```
brianhelba, line 9: This is outside of the named slots, and is a duplicate of the actually-used `csrf_token` below.

PR #3974, opened 2024-07-16
feat(templates): Add socialaccount connections to the base template menu

brianhelba:


allauth/templates/allauth/layouts/base.html
```diff
@@ -37,6 +37,15 @@
                                 <a href="{{ change_password_url }}">{% trans "Change Password" %}</a>
                             </li>
                         {% endif %}
+
+
+                        {% url 'socialaccount_connections' as connections_url %}
+                        {% if connections_url %}
+                            <li>
+                                <a href="{{ connections_url }}">{% trans "Account Connection" %}</a>
```
brianhelba, line 45: This string already exists as a translation.

PR #4014, opened 2024-08-06
docs: update attribute mapping to likely modern SAML2 defaults.

FlipperPA:


docs/socialaccount/providers/saml.rst
```diff
@@ -73,9 +73,9 @@ via the Django admin as well:
                         # If left empty, an attempt will be done to map the attributes using
                         # built-in defaults.
                         "attribute_mapping": {
-                            "uid": "http://schemas.auth0.com/clientID",
-                            "email_verified": "http://schemas.auth0.com/email_verified",
-                            "email": "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress",
+                            "uid": "urn:oid:1.3.6.1.4.1.5923.1.1.1.6",
```
pennersr, line 76: Why this and not e.g. `urn:oid:0.9.2342.19200300.100.1.1` ?

docs/socialaccount/providers/saml.rst
```diff
@@ -73,9 +73,9 @@ via the Django admin as well:
...
+                            "email_verified": "true",
```
pennersr, line 77: `"true"` seems wrong here, that is not an attribute?

pennersr (changes requested):

docs/socialaccount/providers/saml.rst
pennersr: Also wondering, shouldn't (whichever is the right value) be added to `default_attribute_mapping` over at `saml/provider.py`?

FlipperPA: Good points both, I'll make the changes. I'm new to SAML integrations as of this week, so my fresh eyes are both a good and a bad thing!

docs/socialaccount/providers/saml.rst
FlipperPA: This is interesting: everything I've been able to find says this will vary by IdP, and some places have suggested defaulting to using the same field as email. The IdP I'm working with (University of PA) only supports `urn:oid`. I initially had `urn:oid:0.9.2342.19200300.100.1.3` (email) in as the default for this value as well. Do you think that would be a good default? I swtiched to `"true"` after examining the code in the provider checking attribute values.

FlipperPA: After speaking with our identity, team, they recommended using the said `urn:oid` value for `email_verified` as a default, since there's no definition in the schema.

allauth/socialaccount/providers/saml/provider.py
```diff
@@ -16,13 +16,15 @@ class SAMLProvider(Provider):
...
         "uid": [
+            "urn:oid:0.9.2342.19200300.100.1.1",
             "urn:oasis:names:tc:SAML:attribute:subject-id",
         ],
         "email": [
             "urn:oid:0.9.2342.19200300.100.1.3",
             "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress",
         ],
         "email_verified": [
+            "urn:oid:0.9.2342.19200300.100.1.3",
```
pennersr, line 27: `email_verified` is a boolean, indicating whether or not the email address (as passed via another field) is to be treated as already verified. The URN you have here is not a boolean, it is the email itself. That is not correct.

pennersr (changes requested):

docs/socialaccount/providers/saml.rst
pennersr: If there is nothing representing `email_verified` then we shouldn't be listing any `urn` value at all.

allauth/socialaccount/providers/saml/provider.py
FlipperPA: @pennersr I'm at a bit of a loss here on how to proceed, if you have any ideas, that would be great. My proof-of-concept only started working after I provided a value for `email_verified`, and I suspect both the URN and the `"true"` values work because they resolve to Python's `True`.

pennersr: There is not much you can do if there is no attribute from which `email_verified` can be safely derived. So it shouldn't be included here at all. Then, if you are certain your provider does hand over verified email addresses, you can just use the regular `VERIFIED_EMAIL` setting from the provider confg (see https://docs.allauth.org/en/latest/socialaccount/provider_configuration.html) to force those email addresses to be interpreted as verified.

pennersr:
https://codeberg.org/allauth/django-allauth/pulls/4014

PR #4021, opened 2024-08-09
fix(mfa): fix `AuthenticateView` to use form class from setting

eyalch:


eyalch:
In addition to the fix in this PR, it seems that the new WebAuthn–related forms are not customizable. See [allauth/mfa/webauthn/views.py](https://github.com/pennersr/django-allauth/blob/20bedc4f206a65f49a5811eeea165ed4e310d85a/allauth/mfa/webauthn/views.py).

violuke:
We'd like to upgrade to v64 but cant without this fix in place. Thank you :smile:

pennersr (changes requested):
Can you add this to `docs/mfa/forms.rst` as well? Thanks.

eyalch:
Done.

PR #4022, opened 2024-08-10
chore(mfa): add tags to submit button

hoenie-ams:
Same as `mfa/reauthenticate.html`

### DIFF
--- a/allauth/templates/account/reauthenticate.html
+++ b/allauth/templates/account/reauthenticate.html
@@ -14,7 +14,7 @@
             {{ redirect_field }}
         {% endslot %}
         {% slot actions %}
-            {% element button type="submit" %}
+            {% element button type="submit" tags="primary,mfa,login" %}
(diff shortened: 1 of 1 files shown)


hoenie-ams:
I removed the `mfa` tag. Could you please squash these commits?  Apologies for the inconvenience!

allauth/templates/account/reauthenticate.html
```diff
@@ -14,7 +14,7 @@
             {{ redirect_field }}
         {% endslot %}
         {% slot actions %}
-            {% element button type="submit" %}
+            {% element button type="submit" tags="primary,login" %}
```
pennersr, line 17: I am not sure `login` makes sense here.
```suggestion
            {% element button type="submit" tags="primary,reauthenticate" %}
```

pennersr (changes requested):

Issue #4024, opened 2024-08-12
Cannot override headless inputs


mecampbellsoup:
It would be nice, I think, if the headless API endpoints honored any `ACCOUNT_FORMS` overrides set in account configuration.

However, it seems that e.g. the parent classes of `RequestPasswordResetInput` are hard-coded to the base form: https://github.com/pennersr/django-allauth/blob/9440c0a9cef050be11bbad533ea95b11a5f45fee/allauth/headless/account/inputs.py#L107-L108

Could we do something similar to the `BaseSignupForm` pattern in order to honor form overrides? https://github.com/pennersr/django-allauth/blob/9440c0a9cef050be11bbad533ea95b11a5f45fee/allauth/headless/account/inputs.py#L22

Related to https://github.com/pennersr/django-allauth/issues/4011,  I was hoping to put some custom `Form.save()` logic in here.

pennersr:
I would prefer to avoid this:

- The headed and headless behavior would start to behave differently.
- It makes the inputs part of the public API, thus affecting e.g. semver, whereas the implementation should be mostly considered to be internal.
- "It would be nice, I think" -- given the above two points, it would really require a good use case to start doing this.

mecampbellsoup:
I'm not following your bullet points... I'm simply looking for a way to override some part of the headless API endpoint/view.

I would say my proposal actually unifies the headed and headless APIs since any custom form behavior would, under my proposal, behave the same way. Today, if you specify a customization in `ACCOUNT_FORMS['reset_password']`, that customization only applies to the headed API endpoint. Is that really desirable?

Our desired UX is as follows:
* When a user without a usable password requests a password reset, the operation does not result in a password reset email (w/ key) being sent to them. We _could_ focus on the reset password view instead (after they get the email, open the UI, and type in their new password), but we're subjecting them to extra steps that are guaranteed to be futile, so that doesn't seem ideal.

As of the current code base, the only thing I can think of doing (assuming we continue to use the headless API) is to override [`DefaultAccountAdapter.get_reset_password_from_key_url`](https://github.com/mecampbellsoup/django-allauth/blob/88902649f8b59ff10e3a5bca1e26dda940528823/allauth/account/forms.py#L612) to, say, throw an exception, but that seems like a hack.

I have whipped up two POCs for your consideration:

1. [Honoring `ACCOUNT_FORMS['reset_password']` in its headless counterpart view](https://github.com/pennersr/django-allauth/pull/4027); and
[...]

PR #4034, opened 2024-08-15
feat(account): Move RequestPasswordReset.save impl into account adapter

mecampbellsoup:
Related to https://github.com/pennersr/django-allauth/issues/4024.

### DIFF
--- a/allauth/account/adapter.py
+++ b/allauth/account/adapter.py
@@ -563,9 +563,19 @@ class DefaultAccountAdapter(BaseAdapter):
 
         return url_has_allowed_host_and_scheme(url, allowed_hosts=allowed_hosts)
 
+    def send_password_reset_mail(self, user, email, context):
+        """
+        Method intended to be overridden in case you need to customize the logic
+        used to determine whether a user is permitted to request a password reset.
+        For example, if you are enforcing something like "social only" authentication
+        in your app, you may want to intervene here by checking `user.has_usable_password`
+
+        """
+        return self.send_mail("account/email/password_reset_key", email, context)
+
     def get_reset_password_from_key_url(self, key):
         """
-        Method intented to be overriden in case the password reset email
+        Method intended to be overridden in case the password reset email
         needs to be adjusted.
         """
         from allauth.account.internal import flows
--- a/allauth/account/forms.py
+++ b/allauth/account/forms.py
@@ -14,7 +14,7 @@ from allauth.core import context, ratelimit
 from allauth.utils import get_username_max_length, set_form_field_order
 
 from . import app_settings
-from .adapter import get_adapter
+from .adapter import DefaultAccountAdapter, get_adapter
 from .app_settings import AuthenticationMethod
 from .models import EmailAddress, Login
 from .utils import (
@@ -598,18 +598,15 @@ class ResetPasswordForm(forms.Form):
             raise get_adapter().validation_error("unknown_email")
         return self.cleaned_data["email"]
 
-    def save(self, request, **kwargs):
+    def save(self, request, **kwargs) -> str:
         email = self.cleaned_data["email"]
         if not self.users:
             flows.signup.send_unknown_account_mail(request, email)
-        else:
-            self._send_password_reset_mail(request, email, self.users, **kwargs)
-        return email
+            return email
 
-    def _send_password_reset_mail(self, request, email, users, **kwargs):
+        adapter: DefaultAccountAdapter = get_adapter()
         token_generator = kwargs.get("token_generator", default_token_generator)
-
-        for user in users:
+        for user in self.users:
             temp_key = token_generator.make_token(user)
 
             # send the password reset email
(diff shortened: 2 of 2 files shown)


mecampbellsoup:
@pennersr would it be possible to get this one reviewed while you're at it?

allauth/account/adapter.py
```diff
@@ -565,9 +565,26 @@ def is_safe_url(self, url):
 
         return url_has_allowed_host_and_scheme(url, allowed_hosts=allowed_hosts)
 
+    def request_password_reset(self, form, request: HttpRequest, **kwargs) -> str:
```
pennersr, line 568: It would be good to improve the method signature:
- `request` can be dropped -- over time the adapter methods grew inconsisently, some have `request`, some don't. Plan is to move to not requiring it, as the request is always available via `self.request`.
- An untyped/unspecific`**kwargs` is not very nice
- Question: is it wise to make this method require a `form` at all? An alternative would be for it to accept `user` instead.

allauth/account/adapter.py
```diff
@@ -565,9 +565,26 @@ def is_safe_url(self, url):
...
+        in your app, you may want to intervene here.
+
+        """
+        from allauth.account.internal import flows
+
+        email = form.cleaned_data["email"]
+        if not form.users:
+            flows.signup.send_unknown_account_mail(request, email)
+        else:
+            form._send_password_reset_mail(request, email, form.users, **kwargs)
```
pennersr, line 582: The adapter is now calling a private method of the form. The form itself is not using that method, so there is no reason to have it on the form.

pennersr (changes requested):

pennersr:
Must admit, I am still a bit at odds on how to proceed here. #4027 is starting to look more attractive given the impact of the change in this PR. WDYT?

mecampbellsoup:
> Must admit, I am still a bit at odds on how to proceed here. #4027 is starting to look more attractive given the impact of the change in this PR. WDYT?

I think I was able to improve this PR significantly after your review feedback - let me if the changes sway your opinion.

I'm also fine w/ #4027. If we go that route, would you want to also expand/extend #4027 to include the other forms with the ✅ in my comment [here](https://github.com/pennersr/django-allauth/issues/4024#issuecomment-2291128324) while we're at it?

mecampbellsoup:
@pennersr ping 😄

allauth/account/adapter.py
```diff
@@ -563,9 +563,20 @@ def is_safe_url(self, url):
 
         return url_has_allowed_host_and_scheme(url, allowed_hosts=allowed_hosts)
 
+    def request_password_reset(self, users):
```
pennersr, line 566: Thoughts:
- Naming -- it does not request the password reset
- Behavior -- if you filter users here the emails are silently skipped which is not very obvious.

Wouldn't it make be more useful to add this method to the adapter?:

```
def send_password_reset_mail(self, user, email)
```

Then, you are free to implement that as you see fit (send real reset, skip the mail, or send another mail).

pennersr (changes requested):

allauth/account/forms.py
```diff
@@ -629,7 +625,8 @@ def _send_password_reset_mail(self, request, email, users, **kwargs):
 
             if app_settings.AUTHENTICATION_METHOD != AuthenticationMethod.EMAIL:
                 context["username"] = user_username(user)
-            get_adapter().send_mail("account/email/password_reset_key", email, context)
+            adapter.send_password_reset_mail(user, email, context)
```
mecampbellsoup, line 628: I tried to not pass `context` like this but it was pretty cumbersome to try and build that object in the adapter instead of here so I did it this way.

mecampbellsoup: @pennersr LMK if you'd like this reworked to avoid passing `context` around!

allauth/account/forms.py
```diff
@@ -598,18 +598,14 @@ def clean_email(self):
...
+    def save(self, request, **kwargs) -> str:
         email = self.cleaned_data["email"]
         if not self.users:
             flows.signup.send_unknown_account_mail(request, email)
-        else:
-            self._send_password_reset_mail(request, email, self.users, **kwargs)
-        return email
 
-    def _send_password_reset_mail(self, request, email, users, **kwargs):
+        adapter: DefaultAccountAdapter = get_adapter()
```
pennersr, line 606: I think it would be good to either abort the method inside the `if` above by returning `return email`, or, to reintroduce the `else`. As is now, both cases are executed (the if and the else part), and while that works because `users` is empty it did trip me over while reviewing.

allauth/account/forms.py
pennersr: This looks good to me :+1:

pennersr (changes requested):

allauth/account/forms.py
mecampbellsoup: Ah, good catch though!

pennersr (approved):

Issue #3745, opened 2024-04-23
Github verified email does not seem to work


shughes-uk:
I was trying to figure out why Github user emails were not being verified, but it looks like the provider isn't set up for it.

https://github.com/pennersr/django-allauth/blob/1a4ee1d5b3b61f2e97d2e862ccecd47509022a3e/allauth/socialaccount/providers/github/views.py#L37C1-L56C21

The github email api does return the verification status in the call that is already being made. As far as I can tell, just checking that response and adding `verified_email` to the `extra_data` payload used by `complete_login` should be enough to get things working.

I'd be happy to make a PR if it's welcome.

pennersr:
Correct. GitHub also returns a list of email addresses, so we should parse all of those and not just the first one -- which may not even be the primary one. So we need to extract all emails and their properties (`extract_email_addresses`). PR is welcome!

pennersr:
Fixed via bc9fa026

pennersr:
https://codeberg.org/allauth/django-allauth/issues/3745

PR #4037, opened 2024-08-15
fix: Always fetch emails when QUERY_EMAILS is set

mecampbellsoup:
I encountered this situation by having the following context.

First, I have this social account configuration:

```python
SOCIALACCOUNT_LOGIN_ON_GET = True
SOCIALACCOUNT_EMAIL_AUTHENTICATION = True
SOCIALACCOUNT_EMAIL_AUTHENTICATION_AUTO_CONNECT = True

SOCIALACCOUNT_PROVIDERS = {
    "google": {
        "APP": {
            "client_id": getenv("GOOGLE_OAUTH_CLIENT_ID"),
            "secret": getenv("GOOGLE_OAUTH_CLIENT_SECRET"),
        }
    },
    "github": {
        "APP": {
            "client_id": getenv("GITHUB_OAUTH_CLIENT_ID"),
            "secret": getenv("GITHUB_OAUTH_CLIENT_SECRET"),
        }
    },
}
```
Next, I took these steps:

1. Register a new user w/ email and password, `foo@bar.com`.
2. I configured a GH OAuth application and provider.
3. I hit the `github_login` view to initiate the process of linking my GitHub user w/ email address `foo@bar.com` to my local user account.
4. Instead of the existing user having the GH socialaccount data merged/associated w/ it, allauth was trying to register a fresh user with the same email address, which triggered the "account already exists" email flow.

I inspected the data I was getting back from GH, and because the [fetched profile](https://github.com/mecampbellsoup/django-allauth/blob/aa89da5ce199e71b0ee3fe3a2e44007de1d55ea6/allauth/socialaccount/providers/github/views.py#L29) had an `"email"` key in it, the `self.get_emails()` call is never made. Since I don't have a provider setting on the `"github"` provider configuration above, allauth seems to begin new user registration process instead of e.g. querying the DB for the given email address to see if it exists yet or not.

This PR should be completely backwards compatible w/ current behavior, but in some cases it will result in an extra API call to fetch the user's email addresses. I will leave a comment w/ an alternative option in the diff.

Related to https://github.com/pennersr/django-allauth/issues/3745.

### DIFF
--- a/allauth/socialaccount/providers/github/views.py
+++ b/allauth/socialaccount/providers/github/views.py
@@ -1,3 +1,5 @@
+from typing import Optional
+
 from allauth.socialaccount import app_settings
 from allauth.socialaccount.adapter import get_adapter
 from allauth.socialaccount.providers.oauth2.views import (
@@ -30,13 +32,12 @@ class GitHubOAuth2Adapter(OAuth2Adapter):
         )
         resp.raise_for_status()
         extra_data = resp.json()
-        if app_settings.QUERY_EMAIL and not extra_data.get("email"):
-            emails = self.get_emails(headers)
-            if emails:
+        if app_settings.QUERY_EMAIL:
+            if emails := self.get_emails(headers):
                 extra_data["emails"] = emails
         return self.get_provider().sociallogin_from_response(request, extra_data)
 
-    def get_emails(self, headers):
+    def get_emails(self, headers) -> Optional[list]:
         resp = (
             get_adapter().get_requests_session().get(self.emails_url, headers=headers)
         )
--- a/allauth/socialaccount/providers/bitbucket_oauth2/views.py
+++ b/allauth/socialaccount/providers/bitbucket_oauth2/views.py
@@ -21,11 +21,12 @@ class BitbucketOAuth2Adapter(OAuth2Adapter):
             .get(self.profile_url, params={"access_token": token.token})
         )
         extra_data = resp.json()
-        if app_settings.QUERY_EMAIL and not extra_data.get("email"):
-            extra_data["email"] = self.get_email(token)
+        if app_settings.QUERY_EMAIL:
+            if email := self.get_email(token):
+                extra_data["email"] = email
         return self.get_provider().sociallogin_from_response(request, extra_data)
 
-    def get_email(self, token):
+    def get_email(self, token) -> str:
         """Fetches email address from email API endpoint"""
         resp = (
             get_adapter()


allauth/socialaccount/providers/github/views.py
```diff
@@ -30,9 +30,11 @@ def complete_login(self, request, app, token, **kwargs):
         )
         resp.raise_for_status()
         extra_data = resp.json()
-        if app_settings.QUERY_EMAIL and not extra_data.get("email"):
+        if app_settings.QUERY_EMAIL:
             emails = self.get_emails(headers)
             if emails:
+                if email := extra_data.get("email"):
+                    emails = [e for e in emails if e["email"] == email]
```
mecampbellsoup, line 37: IMO we could remove this.

When I was experimenting locally without this filtering, allauth got a list of 4 email addresses back from GH, but seemingly only stored either (1) the one marked as primary, or (2) the one(s) matching the existing local allauth user's email address. TBH I didn't dig deeply enough through the code to know which of those resulted in what I observed.

allauth/socialaccount/providers/github/views.py
```diff
@@ -30,9 +30,11 @@ def complete_login(self, request, app, token, **kwargs):
         )
         resp.raise_for_status()
         extra_data = resp.json()
-        if app_settings.QUERY_EMAIL and not extra_data.get("email"):
```
mecampbellsoup, line 33: There is one other provider, BitBucket, that does this pattern, we may want to also make this change there: https://github.com/mecampbellsoup/django-allauth/blob/aa89da5ce199e71b0ee3fe3a2e44007de1d55ea6/allauth/socialaccount/providers/bitbucket_oauth2/views.py#L24

pennersr: Would be good indeed, can you make that change as part of this PR?

allauth/socialaccount/providers/github/views.py
pennersr: Yes, we should remove this. There is a method `cleanup_email_addresses()` which handles multiple email addresses. If there are issues there, than I would prefer for those to be addressed instead.

pennersr (changes requested):

pennersr:
Note that in your scenario allauth (correctly) starts a new registration because email authentication cannot be used. The provider did provide an email, but as that email is not marked as verified, that email cannot be trusted.

mecampbellsoup:
> Note that in your scenario allauth (correctly) starts a new registration because email authentication cannot be used. The provider did provide an email, but as that email is not marked as verified, that email cannot be trusted.

The code makes it impossible. The [`profiles_url`](https://github.com/mecampbellsoup/django-allauth/blob/aa89da5ce199e71b0ee3fe3a2e44007de1d55ea6/allauth/socialaccount/providers/github/views.py#L29) which makes a GH API call to the [`/user` endpoint](https://docs.github.com/en/rest/users/users?apiVersion=2022-11-28#get-the-authenticated-user) _never_ contains anything except for a key called `"email"` which contains the email address. To get the other properties including `verified` and `primary`, the [`emails_url` call](https://github.com/mecampbellsoup/django-allauth/blob/aa89da5ce199e71b0ee3fe3a2e44007de1d55ea6/allauth/socialaccount/providers/github/views.py#L34) must be made as GH's [`user/emails` endpoint is what has that data](https://docs.github.com/en/rest/users/emails?apiVersion=2022-11-28#list-email-addresses-for-the-authenticated-user).

pennersr (changes requested):
Looks good, though it breaks the build on typing related constructs. Can you look into these? Thanks.

PR #4040, opened 2024-08-16
Email verification by code

pennersr:


KaratasFurkan:
Hi @pennersr , thanks for the feature!

I was testing it today and it works quite well. The only part I couldn't manage is re-sending the verification code email. I want to add a `Resend the verification email` button to the `confirm_email_verification_code.html` in case the user doesn't get the email for some reason. But I got lost in the source code unfortunately.

This is what I've tried so far:
```python
class ResendVerificationCodeEmailView(RedirectView):
    pattern_name = 'account_email_verification_sent'

    def post(self, request, *args, **kwargs):
        # This doesn't send an email:
        # email_verification_by_code.request_email_verification_code(
        #     self.request, user=None, email=request.session['account_email_verification_code']['email']
        # )
        #
        # This raises error regarding anonymous user
        # send_email_confirmation(self.request, request.user, email=request.session['account_email_verification_code']['email'])
        #
        # This raises error regarding anonymous user similar to "send_email_confirmation"
        # account_adapter = get_adapter()
        # if account_adapter.should_send_confirmation_mail(request, email_address, signup):
        #     account_adapter.send_confirmation_mail(request, emailconfirmation, signup)
        return self.get(request, *args, **kwargs)
```

Can you please help?

PR #4049, opened 2024-08-20
feat: Support Django 5.1's LoginRequiredMiddleware

### LINKED ISSUE #4001: Django-allauth views not compatible with Django 5.1 LoginRequiredMiddleware
As of Django 5.1, the `LoginRequiredMiddleware` will be part of the Django core. This will be included by default with new Django projects in the `MIDDLEWARE` setting. All Django views will redirect to `LOGIN_URL` by default if the user is not authenticated.

There is a [decorator](https://docs.djangoproject.com/en/dev/topics/auth/default/#disable-login-required-middleware-for-views) `login_not_required` which should be added to any views you do not wish to apply the redirect.

Currently django-allauth login, signup and other views do not use this decorator, so will cause an infinite loop when trying to access them.

As Django 5.1 is due for production release in a few weeks, is there a branch for handling 5.1 compatibility?

adamchainz:
Fixes #4001.

This is just a first pass to add the decorator to the views that I think should be public. I am not sure if a test is required but I also haven't loaded this into a project to check.

# Submitting Pull Requests

## General


pennersr:
>  I am not sure if a test is required but I also haven't loaded this into a project to check.

Without having tests in place it would be too easy to introduce regression once this feature is merged.  But also now, there is no assurance that this is complete. For example, I would have expected `allauth.mfa` and `allauth.headless` to be impacted as well.

This cannot really be covered by a few tests. You actually need to run the whole test suite with a project that uses the `LoginRequiredMiddleware`. Therefore, we need to introduce a `tests.login_required_middleware` project and run all tests for that project as part of the pipelines.

pennersr:
There is not much point on keeping this PR open as it is far from complete. Feel free to file a PR once the work is sufficiently complete.

pennersr:
I've taken a first stab at what I mentioned above here:

https://codeberg.org/allauth/django-allauth/src/branch/feat/login-required-middleware

PR #4081, opened 2024-08-29
docs: update CERN provider

SebastianLopienski:
Update CERN provider links, add clarifications to the description


SebastianLopienski:
Happy to see this PR accepted and merged.
I wonder when the change will be visible at https://docs.allauth.org/en/latest/socialaccount/providers/cern.html?

pennersr:
It should be there now. Thanks.

PR #4083, opened 2024-09-02
feat(usersessions): Track IP and ua changes and trigger signals

bgervan:
# Submitting Pull Requests

## General

 ## Provider Specifics

### DIFF
--- a/allauth/usersessions/app_settings.py
+++ b/allauth/usersessions/app_settings.py
@@ -21,6 +21,20 @@ class AppSettings:
         """
         return self._setting("TRACK_ACTIVITY", False)
 
+    @property
+    def HISTORY_ENABLED(self):
+        """Whether or not to keep track of session history of user agent and IP. When enabled, a
+        history of the last N sessions will be kept.
+        """
+        return self._setting("HISTORY_ENABLED", False)
+
+    @property
+    def HISTORY_MAX_COUNT(self):
+        """The maximum number of history entries to keep track of. When the number of history
+        entries exceeds this number, the oldest entry will be removed.
+        """
+        return self._setting("HISTORY_MAX_COUNT", 10)
+
 
 _app_settings = AppSettings("USERSESSIONS_")
 
--- a/allauth/usersessions/models.py
+++ b/allauth/usersessions/models.py
@@ -3,15 +3,16 @@ from importlib import import_module
 from django.conf import settings
 from django.contrib.auth import get_user
 from django.core.exceptions import ImproperlyConfigured
-from django.db import models
+from django.db import models, transaction
 from django.http import HttpRequest
 from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
 
 from allauth import app_settings as allauth_settings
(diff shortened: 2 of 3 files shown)


allauth/usersessions/app_settings.py
```diff
@@ -21,6 +21,20 @@ def TRACK_ACTIVITY(self):
         """
         return self._setting("TRACK_ACTIVITY", False)
 
+    @property
+    def HISTORY_ENABLED(self):
```
pennersr, line 25: As per our discussion:

> All in all, if we can keep it limited to only the signals part for allauth, then I think there rest can all be done inside your own project?

At this time I don't want to take more in scope than just the signals. E.g. for history you could that we keep a set, or also include timestamp per IP. Possibly, it should also be exposed to the end user. All in all, I don't want to increase scope, and by just having the signals you are free to do as you see fit in your own project.

allauth/usersessions/signals.py
```diff
@@ -1,8 +1,14 @@
+from django.dispatch import Signal
+
 from allauth.account import app_settings
 
 from .models import UserSession
 
 
+user_agent_changed = Signal()
+ip_changed = Signal()
```
pennersr, line 9: Can you please add these to `docs/usersessions/signals.rst` just like `docs/account/signals.rst` ?

allauth/usersessions/signals.py
```diff
@@ -1,8 +1,14 @@
+from django.dispatch import Signal
+
 from allauth.account import app_settings
 
 from .models import UserSession
 
 
+user_agent_changed = Signal()
```
pennersr, line 8: If the goal is to create an audit trail, it would be good to pass `from_user_agent` (=the old one) and `to_user_agent`) as arguments to this signal.

allauth/usersessions/signals.py
pennersr, line 9: Also `from_ip` and `to_ip` would be welcome here.

allauth/usersessions/models.py
```diff
@@ -36,15 +37,56 @@ def create_from_request(self, request):
...
+                    history = session.data.get("history", [])
+                    session.user = defaults["user"]
+                    session.ip = defaults["ip"]
+                    session.user_agent = defaults["user_agent"]
+                    session.last_seen_at = timezone.now()
+                    changed = False
+
+                    if session.ip != new_history_entry["ip"]:
+                        changed = True
+                        ip_changed.send(sender=UserSession, instance=session, new_ip=new_history_entry["ip"])
```
pennersr, line 72: Can you add test coverage for the newly added functionality?

pennersr (changes requested):

allauth/usersessions/app_settings.py
bgervan: Yes alright, I will revert these parts

pennersr:
Moved to https://codeberg.org/allauth/django-allauth/pulls/4083

PR #4089, opened 2024-09-05
Failing test for get_login_redirect_url when LOGIN_ON_EMAIL_CONFIRMATION

francoisfreitag:
I’m not yet sure how to fix this. It looks like a regression from https://github.com/pennersr/django-allauth/commit/560cb24dce5593342d70cd81db568a2d35471367.

### DIFF
--- a/allauth/account/tests/test_email_verification.py
+++ b/allauth/account/tests/test_email_verification.py
@@ -15,6 +15,7 @@ from pytest_django.asserts import (
 )
 
 from allauth.account import app_settings
+from allauth.account.adapter import DefaultAccountAdapter
 from allauth.account.models import (
     EmailAddress,
     EmailConfirmation,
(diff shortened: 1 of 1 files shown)


pennersr:
That commit is a refactor that removes cleans up the adhoc handling of the email verification link as was the case before. And, in this case it causes a change in behavior. It now calls `get_signup_redirect_url()` instead of `get_login_redirect_url()` -- because, that is actually what the test case is doing. It is in the process of performing a signup. 

So, while indeed a change in behavior, I do think the current behavior is the correct one and we should leave this as is.

pennersr:
Thanks for the test -- that helps in tracking things down fast. I've adjusted it and committed it here:

5c48f117aab48222f2ef3c7389f3a27e8cb2f783

francoisfreitag:
Oops, thanks for your answer!
That’s indeed an issue from my codebase and it should have used `get_signup_redirect_url()` from the start. That’ll be easy to fix :+1:.

PR #4092, opened 2024-09-05
support oidc groups scope

cyberb:
OpenID Connect supports groups scopes at least Authelia does, is it possible to populate that?

### DIFF
--- a/allauth/socialaccount/providers/openid_connect/provider.py
+++ b/allauth/socialaccount/providers/openid_connect/provider.py
@@ -53,7 +53,7 @@ class OpenIDConnectProvider(OAuth2Provider):
         return self.app.settings.get("token_auth_method")
 
     def get_default_scope(self):
-        return ["openid", "profile", "email"]
+        return ["openid", "profile", "email", "groups"]
 
     def extract_uid(self, data):
         return str(data["sub"])
@@ -65,6 +65,7 @@ class OpenIDConnectProvider(OAuth2Provider):
             name=data.get("name"),
             user_id=data.get("user_id"),
             picture=data.get("picture"),
+            groups=data.get("groups"),
         )
 
     def extract_email_addresses(self, data):


pennersr:
Currently, `SCOPE` is picked up from the settings, like so:

```
    SOCIALACCOUNT_PROVIDERS = {
        "gitlab": {
            "SCOPE": ["api"],
            "APPS": [
                {
                    "client_id": "<insert-id>",
```

Though, for OIDC that is not a good fit, because if you have multiple different providiers they all start sharing the same scope. So we should move toward supporting:

```
    SOCIALACCOUNT_PROVIDERS = {
        "openid_connect": {
            "SCOPE": [... works but not a good fit ...],    # <-- CURRENT
            "APPS": [
                {
                    "provider_id": "authelia",
                    "name": "Authelia SSO",
                    "client_id": "<insert-id>",
                    "secret": "<insert-secret>",
                    "settings": {
                        "server_url": "https://auth.example.com/.well-known/openid-configuration",
                        "scope": [...NEW...],  # <-- NEW
                    }
                }
            ]
        }
    }
```

This can be done in: `allauth/socialaccount/providers/oauth2/provider.py::get_scope()`. I think adding support for that makes more sense than adding it as a default.

cyberb:
oh I see, did not know about the current SCOPE part, it is actually ok for me as I am configuring paperless-ngx and I only have one provider.

What about extracting groups from the response, in adapter I do not seem to have access to the original response, only the result of a default extraction in data dictionary?

pennersr:
> What about extracting groups from the response, in adapter I do not seem to have access to the original response, only the result of a default extraction in data dictionary?

Did you check the social account `.extra_data` field?

cyberb:
Oh right this just worked
```
sociallogin.account.extra_data
```
Sorry feel free to close.

aHVzY2g:
@cyberb sorry to ping you, were the groups directly available to you in `sociallogin.account.extra_data`, after adding the scope, or did you have to do anything else?

cyberb:
I think just scope. it was done for paperless+authelia you can check here if you need more details:

https://github.com/paperless-ngx/paperless-ngx/compare/dev...cyberb:paperless-ngx:dev

https://github.com/syncloud/paperless/blob/master/cli/installer/openid_config.go#L42

https://github.com/syncloud/paperless/blob/master/config/paperless.conf#L97

PR #4094, opened 2024-09-06
Default webauthn templates require django.contrib.humanize

julianwachholz:
When using the default MFA templates, the humanize contrib module is required:

https://github.com/pennersr/django-allauth/blob/57154de32994f907718810e72d7063daa09af1af/allauth/templates/mfa/webauthn/authenticator_list.html#L5


pennersr:
Merged via 7dd7c92771973fd4e8c819af411228f20317cfb0

PR #4115, opened 2024-09-14
lithuanian translations

RimJur:
1) Reviewed all fuzzy translations
2) Still left mfa translations (for the next time)

Question: do these all-fuzzy translations at the bottom of the file mean something? Or is it just translations no longer in use?


pennersr:
Please file pull requests over at https://codeberg.org/allauth/django-allauth/pulls -- thanks.

PR #4121, opened 2024-09-22
Provider suffix

Fuzzwah:


pennersr:
Please open pull requests here: https://codeberg.org/allauth/django-allauth/pulls

PR #4122, opened 2024-09-23
better translation to pt-br

paulocoutinhox:
better and concise portuguese brazil translations that make sense


pennersr:
Please open pull requests here: https://codeberg.org/allauth/django-allauth/pulls

PR #4128, opened 2024-10-14
Fix: Convert Unix timestamp to milliseconds in Sessions component

diegomarvid:
# Fix Unix timestamp conversion in Sessions component

## Changes
- Updated the `Sessions.js` file to convert Unix timestamp to milliseconds before creating a Date object.

## Reason
The `created_at` timestamp was being interpreted as milliseconds instead of seconds, resulting in incorrect date display. This fix ensures the correct date is shown for each session.


pennersr:
The project has been moved to https://codeberg.org/allauth/django-allauth.
Please submit your pull request there.

PR #4163, opened 2025-01-10
Fix vulnerabilities and add feature to delete unverified accounts during signup

inspireclicker:
This pull request addresses two key changes:

1. **Fix Vulnerabilities**: Resolved 8 vulnerabilities (2 moderate, 6 high) by adding an override configuration in the package.json file for dependencies related to react-scripts. The following changes were made to ensure that the vulnerabilities are fixed:
`
"overrides": {
  "react-scripts": {
    "@svgr/webpack": "8.1.0",
    "typescript": "4.9.5",
    "postcss": "8.4.38"
  }
}
`
2. **New Feature to Delete Unverified Accounts During Signup**: A new feature was implemented to automatically delete unverified accounts during the signup process. This feature can be enabled by setting DELETE_UNVERIFIED_ACCOUNTS_ON_SIGNUP = True in settings.py. The logic was added to the SignupView class in views.py, where it checks for unverified accounts and deletes them before allowing a new user to register. This ensures that users can continue the registration process without encountering conflicts from old, unverified accounts.

**Files Changed**:

- allauth/headless/account/views.py: Updated the SignupView class to include the logic for deleting unverified accounts.

- examples/react-spa/frontend/package.json: Added override configurations to resolve package vulnerabilities.

### DIFF
--- a/allauth/headless/account/views.py
+++ b/allauth/headless/account/views.py
@@ -105,6 +105,11 @@ class SignupView(APIView):
             return ConflictResponse(request)
         if not get_account_adapter().is_open_for_signup(request):
             return ForbiddenResponse(request)
+        
+        unverified_users = User.objects.filter(email=self.input.cleaned_data["email"], is_verified=False)
+        if settings.DELETE_UNVERIFIED_ACCOUNTS_ON_SIGNUP and unverified_users.exists():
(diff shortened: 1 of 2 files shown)


pennersr:
Please use codeberg to open pull requess.

PR #4185, opened 2025-03-07
Refactored redirect url to be optional in apple oauth client

maxellmilay:
## Description

Made the redirect URL for apple oauth client optional since there are times that it is not needed, like when dealing with mobile apps as the client


pennersr:
As the pull request template clearly indicated, please use Codeberg.

PR #4211, opened 2025-06-21
tests: move all tests into tests/

pennersr:


giohappy:
@pennersr, we are relying on the allauth tests for our own custom providers. This was possible until 65.9.0, but after this commit they were relocated and, unfortunately, removed from the published package.
Do you have suggestions to keep using your test mixins?

